PR #11, opened 2019-04-29
Role to install custom grafana + dashboards

vadimzharov:


etsauer:
hey @vadimzharov, I think that rather than cary a custom ansible role, we would rather use an openshift-applier based approach, as all we're really doing is deploying YAML files. We're using the approach here: https://github.com/redhat-cop/openshift-toolkit/tree/master/custom-dashboards

PR #13, opened 2019-08-26
Replace the hygieia solution with grafana dashboard

etsauer:
@sabre1041 @pcarney8 we're turning over a new leaf! In this PR I am splitting out the dashboard from openshift-toolkit, and reviving this repo so that the MDT work can take on a life of its own again.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,5 +1,5 @@
 # From 'openshift-applier'
 - src: https://github.com/redhat-cop/openshift-applier
   version: v2.1.1
-- src: https://github.com/openshift-toolkit
+- src: https://github.com/redhat-cop/openshift-toolkit
   version: master


tomgeorge:
- I had to change the git URL for openshift-toolkit in `requirements.yml` to the correct git URL.  
- Perhaps as a result of this, I had to change the path to `openshift-cluster-seed.yml` in the `ansible-playbook ... -e include_tags=infrastructure` command to ` galaxy/openshift-toolkit/openshift-toolkit/custom-dashboards/.appler galaxy/openshift-applier/playbooks/openshift-cluster-seed.yml`

After that, and with the `host` file, I was able to verify that it worked in a 4.x cluster with `crc`.  I'm trying to install it on minishift, but I think it's failing because it can't find the `ServiceMonitor` CRD which I think is installed by the prom operator.

I'll spin up a 3.11 cluster tomorrow somewhere and test it out.

PR #19, opened 2019-11-01
federated prometheus instance

ramius345:
I have tested this applying on a 4.1 cluster.  I am going to try it on a 3.11 cluster and will issue more commits if I find any issues with it.  I just wanted to make sure I gave you enough time to review it so here it is.

I wanted to note that the requirements.yml files are still pointing at my repositories for testing.  Once you are satisfied with them, I will do a quick commit to point them at redhat-cop.


ramius345:
I tested this on a 3.11 cluster now.  There was a conflict with the proxy image.  I pulled in the proxy image that 3.11 was using.  This allows it to run on a 3.11 or 4.x cluster.  Everything else appears to be working.  I am seeing metrics in the federated prometheus and data on the grafana dashboard.

PR #20, opened 2019-11-01
Lead Time Exporter

AlyIbrahim:
@etsauer 
I created the required exporter, I tested locally using podman with UBI image as well, the template you created earlier should match except for a needed ~/.kube/config file
The exporter is expecting 3 Enivronment Variables:
GITHUB_USER: mandatory
GITHUB_TOKEN: mandatory 
PROJECTS : list of comma separated project names, This is optional, and if not set, the exporter will fetch all the user projects

The exporter relies on ~/.kube/config file to connect to openshift clusters, the file should be mounted to the pod .

### DIFF
--- a/metrics-exporters/leadtime-exporter/lt-exporter.py
+++ b/metrics-exporters/leadtime-exporter/lt-exporter.py
@@ -87,7 +87,19 @@ def convert_timestamp_to_date_time(timestamp):
 def generate_ld_metrics_list(projects):
 
     urllib3.disable_warnings()
-    k8s_client = config.new_client_from_config()
+    if "OPENSHIFT_BUILD_NAME" in os.environ:
+        config.load_incluster_config()
+        file_namespace = open(
+            "/run/secrets/kubernetes.io/serviceaccount/namespace", "r"
+        )
+        if file_namespace.mode == "r":
+            namespace = file_namespace.read()
+            print("namespace: %s\n" %(namespace))
+    else:
+        config.load_kube_config()
+
+    k8s_config = client.Configuration()
+    k8s_client = client.api_client.ApiClient(configuration=k8s_config)
     dyn_client = DynamicClient(k8s_client)
 
     metrics = []
--- a/metrics-exporters/leadtime-exporter/lt-exporter.py
+++ b/metrics-exporters/leadtime-exporter/lt-exporter.py
@@ -1,0 +1,141 @@
+import yaml
+import json
+import time
+import urllib3
+import requests
+import os
+from datetime import datetime, timezone
+from kubernetes import client, config
+from openshift.dynamic import DynamicClient
+from prometheus_client import start_http_server
+from prometheus_client.core import CounterMetricFamily,InfoMetricFamily,GaugeMetricFamily, REGISTRY
+
+class CommitCollector(object):
+    _prefix = "https://api.github.com/repos/"
+    _suffix = "/commits"
+    def __init__(self, username, token, projects, apps):
+        self._username = username
+        self._token = token
+        self._projects = projects
+        self._apps = apps
+    def collect(self):
+        ld_metric = GaugeMetricFamily('github_commit_timestamp', 'Commit timestamp', labels=['namespace', 'app', 'build', 'commit_hash'])
+        ld_metrics = generate_ld_metrics_list(projects)
+        for my_metric in ld_metrics:
+            print("Namespace: ", my_metric.namespace, ", App: ", my_metric.app_name, ", Build: ", my_metric.build_name)
+            ld_metric.add_metric([my_metric.namespace, my_metric.app_name, my_metric.build_name, my_metric.commit_hash], my_metric.lead_time)
+            yield ld_metric
+class mdt_metric:
+    def __init__(self, app_name):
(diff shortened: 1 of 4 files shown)


AlyIbrahim:
Update python code to control openshift cluster from the deployed pod as in the blogpost: https://developers.redhat.com/blog/2019/07/25/controlling-red-hat-openshift-from-an-openshift-pod/

ramius345:
I think we need to merge the templates in from https://github.com/redhat-cop/mdt-quickstart/pull/18 .  They are required to install this on a cluster.

metrics-exporters/leadtime-exporter/lt-exporter.py
```diff
@@ -0,0 +1,141 @@
...
+        replicationControllers = v1_replicationControllers.get(namespace=project)
+
+        print ('Builds Count = ' , len(builds.items))
+        for build in builds.items:
+            if build['spec']['source']['type'] == 'Git':
+                if build['status']['phase'] =='Complete':
+                    app_name = build.metadata.labels.app
+                    metric = mdt_metric(app_name)
+                    metric.build_name=build.metadata.name
+                    metric.build_config_name = build.metadata.labels.buildconfig
```
ramius345: I'm wondering if we are fetching/storing more metadata than we need to get the commit hash and timestamp here?  Do we need all these fields to do the timestamp calculation?

metrics-exporters/leadtime-exporter/lt-exporter.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(self, username, token, projects, apps):
+        self._username = username
+        self._token = token
+        self._projects = projects
+        self._apps = apps
+    def collect(self):
+        ld_metric = GaugeMetricFamily('github_commit_timestamp', 'Commit timestamp', labels=['namespace', 'app', 'build', 'commit_hash'])
+        ld_metrics = generate_ld_metrics_list(projects)
+        for my_metric in ld_metrics:
+            print("Namespace: ", my_metric.namespace, ", App: ", my_metric.app_name, ", Build: ", my_metric.build_name)
```
ramius345: I would update this log to mention that we are performing a collection here.  This could be useful to track the activity of the collector.  It would also be useful to probably use something like this https://docs.python.org/3/library/logging.html so we get logging levels and timestamps.

metrics-exporters/leadtime-exporter/lt-exporter.py
```diff
@@ -0,0 +1,141 @@
...
+        self.deployment_config_name = None
+        self.deployment_timestamp = None
+        self.lead_time = 0
+    def getCommitTime(self):
+        _prefix = "https://api.github.com/repos/"
+        _suffix = "/commits/"
+        myurl = self.repo_url
+        url_tokens = myurl.split("/")
+        url = _prefix + url_tokens[3] + "/" +url_tokens[4].split(".")[0] +_suffix+self.commit_hash
+        response = requests.get(url, auth=(username, token))
```
ramius345: Some logging around requests would be useful for any errors we are getting around this metric.

metrics-exporters/leadtime-exporter/lt-exporter.py
```diff
@@ -0,0 +1,141 @@
...
+
+if __name__ == "__main__":
+    username = os.environ.get('GITHUB_USER')
+    token = os.environ.get('GITHUB_TOKEN')
+    projects = None
+    if os.environ.get('PROJECTS') is not None:
+        projects = [ proj.strip() for proj in os.environ.get('PROJECTS').split(",") ]
+    apps = None
+    REGISTRY.register(CommitCollector(username, token, projects, apps))
+    start_http_server(9118)
```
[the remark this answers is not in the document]
AlyIbrahim: Yeah we can change the port to default 8080

metrics-exporters/leadtime-exporter/lt-exporter.py
AlyIbrahim: I am creating a data structure to be used in future exporters if needed, we can strip it down later

metrics-exporters/leadtime-exporter/lt-exporter.py
AlyIbrahim: Using Logging library is a good idea, we can add this as an enhanecement, especially when we get have more exporters, Agree !

metrics-exporters/leadtime-exporter/lt-exporter.py
AlyIbrahim: We're not expecting errors in connectivity, and I don't think logging responses will be useful, we may add a try/expect block to log connectivity errors, What do you think?

metrics-exporters/leadtime-exporter/lt-exporter.py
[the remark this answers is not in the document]
AlyIbrahim: Done

ramius345:
Looks good, I'm going to merge it.

etsauer:
[Add commit time to Lead Time metric](https://trello.com/c/Qm4jihMC/100-add-commit-time-to-lead-time-metric)

Issue #25, opened 2019-11-20
Explore deploying mdt stack with helm and operators instead of templates and applier.


etsauer:
It appears the long term gitops/infra as code solution is going to center around Helm as the templating framework of choice, and ArgoCD as the orchestration engine. We should look at what it would take to convert our applier inventory over to helm/argo.

etsauer:
@ramius345 I think we can consider this closed right?

PR #48, opened 2020-03-27
Refactor lead time exporters to just publish commit time for now

### LINKED ISSUE #28: Lead Time exporter should pull git repo information from BuildConfigs
Currently the lead time exporter requires an administrator to pass a comma-separated list of git repos representing application source code via environment variable to the lead time exporter. This doesn't scale well, as we would have to make an administrative change for each application that comes on board. 

What we should do instead is grab the git repository information from the `BuildConfigs` we find in the cluster.

etsauer:
See metrics-exporters/committime/README.md for validation instructions

resolves #28 
related to #25


themoosman:
The dashboard still deploys as it did.  What exactly should I be testing?

themoosman:
Helm works as it should, the dashboard still works, and the new exporter works.

LGTM

Issue #23, opened 2019-11-18
Long term storage solution


etsauer:
By default prometheus only stores 2 weeks worth of data. In order for MDT tooling to be valuable, we need to store at least 6 months to a year's worth of History. We need to do some research into a long term data store for the stack.

etsauer:
Here's one blog post that suggests a solution, but there may be more.

https://blog.openshift.com/thanos-long-term-storage-for-your-prometheus-metrics-on-openshift/

Also, looking at the Prometheus Storage docs, there's nothing that explicitly says that it should only be used for short-term storage, but it does point out that promehteus itself doesn't do much in the way of operationalizing its storage, and references some thrid-party storage integrations.

https://prometheus.io/docs/prometheus/latest/storage/

ramius345:
I will start taking a look at this tomorrow.  Thanks.

garethahealy:
Theres already been some thanos work in the CoP:
- https://github.com/redhat-cop/uncontained.io/pull/190

CC @robbbbh @AndreaFerraresi

ramius345:
I'll have a look, thanks!

ramius345:
I have a working solution on this using nooba for Openshift 4.  However, when I try this on the CI CASL cluster, nooba is having issues installing with PV provisioning.  Using version 2.0.9 of noobaa, I am able to create a backing store, but I am not able to get noobaa to finish its installation, it gets stuck waiting for something.  Looking at the logs, its looking for AWS and Ceph provisioners.

We should be able to use the rest of the solution with any s3 compatible bucket provider.  It might be time to start looking at Ceph or to just suggest straight up AWS s3.

bbeaudoin:
Nooba aggregates S3 API providers, Ceph API and S3 are the default backing types, this cannot be block so if CI CASL doesn't support S3 API compatible storage you'll need something other than Nooba. Thanos is a good fit but the storage needs to be worked out.

etsauer:
@bbeaudoin ok, I misunderstood what Moobaa was then. @ramius345 I think the default option for this project should be to use MinIO instead.

sbailleu:
Functional code ready (iteration 1), code needs to be merged. Documentation to be completed.

etsauer:
@tolarewaju3 now that @ramius345 work has been merged, here's what I think it left TODO:

- Come up with an opinionated installation method for MinIO that we can document, that at a minimum includes a TLS config
[...]

PR #52, opened 2020-04-10
Convert to helm thanos merge master

ramius345:
Thanos support to address Issue #23 long term storage.


README.md
```diff
@@ -63,6 +63,32 @@ Once you are finished adding your extra hosts, you can update your stack by re-r
...
+
+And then:
+
+```
+./runhelm.sh --values file_with_bucket_config.yaml
+```
+
+The above configuration is the default access endpoint when noobaa is installed into the noobaa namespace.  For installaing noobaa, see:
+
+https://github.com/noobaa/noobaa-core#deploy-to-kubernetes
```
etsauer, line 90: I think we want to be a bit more opinionated than just sending the link to the nooba repo. For example, we don't want people installing this in the `default` namespace, which `noobaa install` does, and they don't provide clear directions on how to create buckets.

charts/deploy/templates/thanos-objstore-config-secret.yml
```diff
@@ -0,0 +1,20 @@
+{{- define "pelorus.thanos.objstore.config" }}
+type: s3
+config:
+  bucket: thanos
```
etsauer, line 4: we probably don't want to hardcode this, but allow (and instruct) the user to set a bucket name

runhelm.sh
```diff
@@ -25,6 +27,12 @@ helm template \
     --set openshift_prometheus_basic_auth_pass=$GRAFANA_DATASOURCE_PASSWORD \
     $EXTRA_VALUES \
     ./charts/deploy/ | oc apply -f - -n pelorus
+
+
+#    --set bucket_access_point=$INTERNAL_S3_ENDPOINT \
```
etsauer, line 32: did you intend to leave this here?

README.md
```diff
@@ -63,6 +63,32 @@ Once you are finished adding your extra hosts, you can update your stack by re-r
...
+
```
etsauer, line 91: How does the user verify that this setup is working, and data is persisting?

ramius345: I've fixed a few of these defects in the documentation as you've asked, but I've run out of time this weekend.  This basically is going to require a full writeup on how to use different S3 providers which may be its own backlog item.  I've removed this line in the meantime and am deferring to how each S3 provider documents it themselves.  I will revisit it when I have some more time.

runhelm.sh
ramius345: Removed.

charts/deploy/templates/thanos-objstore-config-secret.yml
ramius345: I've changed this to the default and provided some documentation on the variable I've added.

README.md
ramius345: I didn't have time to write up a full set of documentation on how noobaa works so I have removed this.  I will revisit it when I have more time.  For now the answer is going to be consult your bucket provider of choice' documentation.

README.md
```diff
@@ -27,7 +27,7 @@ Before deploying the tooling, you must have the following prepared
   * The OpenShift Command Line Tool (oc)
   * Helm 3
 
-### Deployment Instructions
+First, we must collect some information from the cluster to feed to our templates.
```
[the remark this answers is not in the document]
ramius345: Removed.

etsauer:
@ramius345 I think since this has been sitting out for a while, we'll just merge and do a followup PR to address these.

PR #58, opened 2020-04-28
Allow pelorus to be installed in any namespace

etsauer:
To Test:

Follow main README to install using the `-n` flag


tolarewaju3:
pelorus-namespace

output of installing with "-n"

PR #65, opened 2020-04-30
Update Helm Script with Mac Conditional

mattheh:
The '-d' flag for base64 is a debugging flag on Mac.  Need to use '-D' instead.  [link](https://electrictoolbox.com/base64-decode-command-line/)


mattheh:
Turns out this works fine, explicitly using `--decode` should be compatible across all OS

Issue #62, opened 2020-04-29
Some cluster-level resources are overwritten when installing in multiple namespaces in the same cluster


etsauer:
I _think_ this is limited to the clusterrolebindings, but I'm noticing that every time I re-run the `./runhelm.sh` script to a new namespace, the clusterrolebindings are getting overwritten for my serviceaccount, which breaks pelorus in all other namespaces. We should reconfigure the helm chart to handle this better.

PR #67, opened 2020-04-30
Cluster level multiple namespace #62

tolarewaju3:
Added namespace prefix to clusterrolebindings to prevent overriding across a cluster


etsauer:
@tolarewaju3 need to update the prometheus datasource service URL as well.

tolarewaju3:
@etsauer done. I can see data in my dashboard now too

PR #75, opened 2020-05-07
modified app.py to prevent json parsing breaking

kenwilli:


etsauer:
@kenwilli I tested the code here. The app.py now no longer crashes, but its also not matching any of the builds in the cluster with the label I'm testing. When I run the app in a debugger, I see that the jsonpath string doesn't look right:

```
jsonpath_str: '"items[*].metadata.labels.app\\.kubernetes\\.io\\/name"'
```

exporters/committime/app.py
```diff
@@ -87,6 +90,10 @@ def convert_timestamp_to_date_time(timestamp):
     date_time = datetime(timestamp)
     return datetime.strftime('%Y-%m-%dT%H:%M:%SZ')
 
+#Function to take care of escaping special characters so json parse does not fail
```
etsauer, line 93: doesn't look like this is being used either.

exporters/committime/app.py
```diff
@@ -111,11 +118,14 @@ def generate_ld_metrics_list(namespaces):
         # only use builds that have the app label
         builds = v1_builds.get(namespace=namespace, label_selector=loader.get_app_label())
 
+
         # use a jsonpath expression to find all values for the app label
-        jsonpath_str = 'items[*].metadata.labels.%s' % (loader.get_app_label())
+        jsonpath_str = "$['items'][*]['metadata']['labels'][" + str(loader.get_app_label()) + "]"
```
etsauer, line 123: i think this line needs to change to:

```
jsonpath_str = "$['items'][*]['metadata']['labels']["' + str(loader.get_app_label()) + "']"
```

kenwilli: This produces errors in my IDE, it doesn't seem like valid python syntax

etsauer: oh, I had one of my quotes misplaced:

```
jsonpath_str = "$['items'][*]['metadata']['labels']['" + str(loader.get_app_label()) + "']"
```

PR #76, opened 2020-05-13
Updates for Feature #55, remove runhelm.sh

### LINKED ISSUE #55: Use Helm chart for install instead of an install script
The installation for Pelorus is really just a helm chart, except that we rely on a couple of values in secrets in the monitoring stack in order to wire up pelorus with the openshift monitoring stack.

Since we've done this, we have added a `lookup` function to helm, which could be used to fetch these values as part of processing the chart. https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function

We should switch over to that so that we can get rid of the install script altogether.

themoosman:
Remove the `runhelm.sh` the installation script and refactor everything to use straight helm.

resolves #55


themoosman:
Blocked by https://github.com/helm/helm/issues/8137

PR #98, opened 2020-06-17
Refactor main committime method to reduce complexity

etsauer:
This should resolve the final linting error as well.

@redhat-cop/mdt


etsauer:
To test:

1. Log in to openshift (with build data present)
2. Start exporter. `python exporters/committime/app.py`
3. Once started, run the following and confirm you get results back: `curl localhost:8080 | grep 'commit_timestamp{.*} \d\.\d+e\+\d\d'`

PR #113, opened 2020-06-27
adding ct lint action

### LINKED ISSUE #110: Add helm linting and client-side testing using GH Actions

deweya:
Closes #110 

Adds a workflow for linting values.yaml and Chart.yaml styles and schema using [ct](https://github.com/helm/chart-testing). Templates are also checked for invalid YAML syntax. The workflow is in a separate yaml file but of course the existing pylama workflow can be combined with this one instead.

By default, ct will check that you have incremented any chart versions that you modified and will fail if the version isn't incremented. Is this ok? The version check is especially useful if you're releasing to a chart repo, but I don't think this is the case here. LMK if you want this off instead.

I had to make a couple changes to charts/exporter to allow a couple linter rules to pass. The good news is that any style rule currently in place can be overridden by lintconf.yaml or chart_schema.yaml if there's something being enforced you'd rather not be.


charts/exporter/Chart.yaml
```diff
@@ -1,2 +1,3 @@
 name: exporter
-version: v0.0.1
+version: v0.0.2
```
etsauer, line 2: I would update the version to `v.1.1.0` to match the current repo version.

deweya: updated

PR #128, opened 2020-07-13
update metrics-dashboard

jonahkh:


charts/deploy/templates/metrics-dashboard.yaml
```diff
@@ -323,7 +323,7 @@ spec:
                 "tableColumn": "",
                 "targets": [
                     {
-                    "expr": "count (deploy_timestamp)",
+                    "expr": "count( count_over_time(deploy_timestamp [2d]) )",
```
etsauer, line 326: replace `2d` with `$__range` so we are in line with grafana's time range.

charts/deploy/templates/metrics-dashboard.yaml
```diff
@@ -891,7 +891,7 @@ spec:
                 "steppedLine": false,
                 "targets": [
                     {
-                    "expr": "count(deploy_timestamp)",
+                    "expr": "count( count_over_time(deploy_timestamp [2d]) )",
```
etsauer, line 894: same here, replace `2d` with `$__range`

etsauer:
replaced by #144

PR #135, opened 2020-07-15
🦩fix secret type indent 🦩

eformat:
tab wrong. this causes issues when syncing from ArgoCD.

these are all fixes to get sync working for pelorus /deploy Chart from argocd.

https://github.com/rht-labs/ubiquitous-journey/pull/179


charts/deploy/templates/grafana-oauth-session-secret.yaml
```diff
@@ -5,4 +5,4 @@ data:
 kind: Secret
```
etsauer, line 5: @eformat I think this template is no longer used anyway. You can just delete it.

charts/deploy/charts/grafana-operator/templates/Grafana.yaml
```diff
@@ -86,10 +86,8 @@ spec:
               properties:
                 ports:
                   type: array
-                  description: Override default ports
```
etsauer, line 89: did you remove these descriptions intentionally?

eformat: argocd-sync-grafana

yeah .. this is annoying - argocd sync fails with these descriptions there (1.6.1 argocd) .. CRD's are fiddly i know as far as sync is concerned will see if there is workaround on argocd side (ignoredifferences etc) ...

charts/deploy/templates/grafana-oauth-session-secret.yaml
eformat: @etsauer - no, looks like its still used here - https://github.com/redhat-cop/pelorus/blob/master/charts/deploy/templates/grafana.yaml#L61

charts/deploy/charts/grafana-operator/templates/Grafana.yaml
eformat: reverted these .. looking for argocd side change instead

etsauer:
@eformat the suggestion for this was to just delete `charts/deploy/templates/grafana-oauth-session-secret.yaml` altogether rather than fix the indentation. This file is no longer used.

eformat:
> @eformat the suggestion for this was to just delete `charts/deploy/templates/grafana-oauth-session-secret.yaml` altogether rather than fix the indentation. This file is no longer used.

yes, it is used ? here: https://github.com/redhat-cop/pelorus/blob/master/charts/deploy/templates/grafana.yaml#L61

etsauer:
@eformat my mistake. I was looking at the wrong file. Can you tick the chart version number please?

PR #142, opened 2020-07-16
Commit time refactor

kkoller:


kkoller:
I'm still debugging this, but I think its to a point where it can be open for review.

exporters/committime/app.py
```diff
@@ -22,233 +18,31 @@
...
-        else:
-            commit = response.json()
-            try:
-                self.commit_time = commit['commit']['committer']['date']
-                self.commit_timestamp = pelorus.convert_date_time_to_timestamp(self.commit_time)
-            except Exception:
-                logging.error("Failed processing commit time for build %s" % self.build_name, exc_info=True)
-                logging.debug(commit)
-                raise
+class GitFactory:
```
etsauer, line 21: we should default to `git_type=github` so that we don't break existing deployments. Right now if I rebuild an existing commit exporter, the pod will fail to start with:

```
  File "committime/app.py", line 45, in <module>
    REGISTRY.register(collector)
  File "/opt/app-root/lib/python3.6/site-packages/prometheus_client/registry.py", line 26, in register
    names = self._get_names(collector)
  File "/opt/app-root/lib/python3.6/site-packages/prometheus_client/registry.py", line 53, in _get_names
    desc_func = collector.collect
AttributeError: 'NoneType' object has no attribute 'collect'
```

exporters/committime/app.py
```diff
@@ -22,233 +18,31 @@
...
+        if git_type == "bitbucket":
+            return BitbucketCommitCollector("", "", "", "")
 
 
 if __name__ == "__main__":
     pelorus.check_required_config(REQUIRED_CONFIG)
     username = os.environ.get('GITHUB_USER')
     token = os.environ.get('GITHUB_TOKEN')
     git_api = os.environ.get('GITHUB_API')
+    git_type = os.environ.get('GIT_TYPE')
```
etsauer, line 37: I think we should change `git_type` to `git_provider`.

themoosman: I think we should change them all to be `GIT_xxx`.  The only caveat to that is what if I want to run two committime exporters against two different Git providers.

kkoller: Agreed with @themoosman that we should rename all these ENV vars.  Making the change now will break existing deployments and we will need a strategy for managing separately running instances.  

Should we make this change now with the refactor, or leave the 3 existing ENV vars and open an issue to fix?

kkoller: Ok I think I fixed it

exporters/committime/app.py
```diff
@@ -1,254 +1,51 @@
...
+from prometheus_client.core import REGISTRY
 
-REQUIRED_CONFIG = ['GITHUB_USER', 'GITHUB_TOKEN']
+REQUIRED_CONFIG = ['GIT_USER', 'GIT_TOKEN']
 
 pelorus.load_kube_config()
 k8s_config = client.Configuration()
 k8s_client = client.api_client.ApiClient(configuration=k8s_config)
 dyn_client = DynamicClient(k8s_client)
+default_git = "github"
```
etsauer, line 19: we currently have all default values declared as constants in pelorus.py. Unless you have some argument, let's continue that pattern.

docs/Configuration.md
```diff
@@ -102,12 +102,17 @@ This exporter supports several configuration options, passed via environment var
```
etsauer, line 102: I think we need to add a section here that explicitly states what "providers" we offer.

PR #146, opened 2020-07-22
Add default datefmt and format to logger

### LINKED ISSUE #115: Enhance logger in the exporters to output a timestamp with every log message
Currently, the exporters don't add a timestamp to each log message.  Upgrade the exporters to include the timestamp.

themoosman:
resolves #115


themoosman:
To test, run any of the exporters locally.

Issue #127, opened 2020-07-09
RFE: Support OpenShift Dedicated


mrjbanksy:
We are attempting to deploy Pelorus to OpenShift Dedicated, where we have `dedicated-admin` access to the cluster, not `cluster-admin`. `dedicated-admin` has a limited subset of the permissions that `cluster-admin` has. I believe the `dedicated-admin` permissions are defined here: https://github.com/openshift/managed-cluster-config/tree/master/deploy/rbac-permissions-operator-config.

Attempting to run the `runhelm.sh` script has the following output. I removed the `exit 1` from the Grafana and Prometheus password checks to get all the error messages in one run:

```
[jeremy@colossus pelorus (master)]$ ./runhelm.sh -n pelorus-1
Error from server (Forbidden): secrets "grafana-datasources" is forbidden: User "Jeremy.Banks@example.com" cannot get resource "secrets" in API group "" in the namespace "openshift-monitoring"
Could not find the Grafana datasource password in the openshift-monitoring namespace!
Error from server (Forbidden): secrets "prometheus-k8s-htpasswd" is forbidden: User "Jeremy.Banks@example.com" cannot get resource "secrets" in API group "" in the namespace "openshift-monitoring"
Could not find the prometheus htpasswd file secret in the openshift-monitoring namespace!
Warning: oc apply should be used on resource created by either oc create --save-config or oc apply
namespace/pelorus-1 configured
serviceaccount/grafana-operator created
serviceaccount/prometheus-operator-pelorus created
serviceaccount/pelorus-prometheus created
secret/grafana-k8s-proxy created
secret/prometheus-pelorus-htpasswd created
secret/additional-scrape-configs created
configmap/pelorus-sdm created
configmap/trusted-certs created
Warning: oc apply should be used on resource created by either oc create --save-config or oc apply
Warning: oc apply should be used on resource created by either oc create --save-config or oc apply
Warning: oc apply should be used on resource created by either oc create --save-config or oc apply
clusterrole.rbac.authorization.k8s.io/grafana-proxy created
clusterrole.rbac.authorization.k8s.io/prometheus-mdt-oauth-clusterrole created
clusterrolebinding.rbac.authorization.k8s.io/pelorus-1-prometheus-mdt-oauth-crb created
service/prometheus-operator-pelorus created
service/prometheus-pelorus created
deployment.apps/grafana-operator created
deployment.apps/prometheus-operator-pelorus created
grafana.integreatly.org/grafana-oauth created
[...]

PR #154, opened 2020-07-29
Create a minimal install assuming that Grafana already exists in the managed cluster.

willowmck:
This change introduces a parallel set of Helm charts to only deploy the GrafanaDashboard assets required for Pelorus assuming that the user is in a managed environment like OpenShift Dedicated and Grafana cannot be installed via this chart due to security restrictions.

## Linked Issues?

related to #127 

## Testing Instructions

I updated the install documentation describing how the minimal install process works.  This is simply the first step towards getting Pelorus working in OpenShift Dedicated.  Next, I will be looking into the exporters so that they will work with Gitlab.

@redhat-cop/mdt


willowmck:
No longer needed as the helm chart refactor fixed my issue on OSD.

Issue #120, opened 2020-07-06
Multi-tenant Pelorus


kkoller:
Filter Pelorus dashboards such that users can see data for the projects in which they have read permissions.

- Continue delegating to cluster authentication
- Don't require cluster read
- Populate dashboards based on current permission levels

prakritiverma:
Enhance (or create separate) dashboard to be able to drill down to metrics for an individual app

PR #190, opened 2020-09-08
Change lead time calculation so that we can actually sort by application

etsauer:

This PR accomplished the following:

* it updates the deploytime and committime exporters to export the same value for `{app=xxx}`. That way we can compare lead times for individual applications
* it updates the sdp:lead_time:by_app metric to fix the app name mismatch, and also ensure that we only count lead time for the _first_ deployment of a given image id.

## Linked Issues?

related to #120 

## Testing Instructions

@redhat-cop/mdt


etsauer:
I've updated the PR to include a prometheus unit test for our recording rules. This unit test actually proves that we still don't have the correct expression for `sdp:lead_time:by_commit`. You can test this yourself by checking out this code, and running:

```
promtool test rules _test/prometheus/test.yaml
```

etsauer:
resolves #191

themoosman (approved):
LGTM

PR #209, opened 2020-10-02
Better project matching logic

### LINKED ISSUE #208: CommitMetric.__parse_repourl() fails to parse correct repo project name when using multiple parents.
I have a GitLab parent group that has a repository that is more than 1 child deep.  This causes the parser to fail to pick up the correct project name and thus it cannot match the commit hash from the repository.

See https://github.com/redhat-cop/pelorus/blob/v1.2.2/exporters/committime/collector_base.py#L245-L251 for the relevant code.

This example from the PyCharm console shows what happens.  You can see here that it is picking 'osd' as the repo name instead of osd-infrastructure.
pelorus_error

willowmck:
There are three fixes here.  The first fixes the parsing of repository projects by selecting the last token for the project name.  
Second, is an addition of type Docker for code builds.
Last, when a search by project name is made to the Gitlab API, this change will attempt to find a project in the search results that matches the Git URL for the repository.

A bit more details on the changes here since Eric and I just had a conversation about this.

First, the original code would call to the Gitlab API with a project name and select the first match in the returned list.  This has been enhanced to take the repo_url from the CommitMetric and attempt to match the project which has an equivalent ssh_url_to_repo or http_url_to_repo.  See the [Gitlab API](https://docs.gitlab.com/ee/api/projects.html#get-single-project) for what is returned in the project object.  The only other possibility on the Project object returned by the API is the web_url property which I could add but have not in this commit.  Note that if we go through all of the projects and there are no equivalency matches, then the project will be set to None.  The Gitlab API doesn't document the Project object and all of its attributes; however, it seems that this equivalency test should be sufficient as forked projects would have different urls from the parent project.  Also, the search method only returns 20 results by default, so I added recursion to walk through the paged results.

Secondly, regarding partial matches.  I'm curious to your thoughts here as to how this would happen?  Would it be that the BuildConfig is created with only part of the url in spec.source.git.uri?  

## Linked Issues?

resolves #208 and #212 and #215 

## Testing Instructions

Use a project that is a grandchild of the group like https://gitlab.com/BhfLab/osd/leia.git and make sure that the correct project name is set for the CommitTime metric.

Make sure that when the Gitlab api search results come back that the correct project is matched from the search results.

@redhat-cop/mdt


exporters/committime/collector_base.py
```diff
@@ -236,16 +235,26 @@ def __parse_repourl(self):
...
+        # In cases where the protocol is found to be 'git', 'http(s)' or 'ssh' it is handled fine
+        # but is not graceful for strange urls like 'kmoos://myhost/mygroup/repo.git'
+        if parsed.protocol == 'ssh' and len(parsed.protocols) > 0:
+            if parsed.protocols[0] == 'https':
+                self.__repo_protocol = parsed.protocols[0]
+            elif parsed.protocols[0] not in ['http', 'https', 'git', 'ssh']:
+                self.__repo_protocol = parsed.protocol
+                self.__repo_fqdn = self.__repo_url.split('/')[2]
+            else:
+                self.__repo_protocol = parsed.protocol
```
etsauer, line 255: Ah, I didn't realize until this comment that you were doing all of this to make the `kmoos://...` test work. I think the actual answer here is that git is more prescriptive than we thought it was, and this is not a valid test. Originally we were really just check of the format of `<something>://<something>/<bla>`, which clearly wasn't sufficient. Let's remove the tests that don't make sense, and remove this code. Sorry for the run around.

willowmck: I'm actually relieved to hear that is not proper grammar.  No problems backing out this code.

exporters/committime/collector_bitbucket.py
```diff
@@ -49,40 +49,41 @@ def get_commit_time(self, metric):
...
-            # set the URL back to the original
-            metric.repo_url = old_url
-        elif api_version == '2.0':
-            # Just set the V2 globals
-            api_root = self.V2_API_ROOT
-            api_pattern = self.V2_API_PATTERN
-
-        # Create the API server from the Git server and API Root
-        api_server = pelorus.url_joiner(git_server, api_root)
+        try:
```
etsauer, line 52: where does this try block end?

exporters/committime/collector_base.py
```diff
@@ -236,16 +238,17 @@ def __parse_repourl(self):
...
-            self.__repo_protocol = self.__repo_protocol[:-1]
-        # token 1 is always a blank
-        self.__repo_fqdn = url_tokens[2]
-        self.__repo_group = url_tokens[3]
-        self.__repo_name = url_tokens[4]
-        if self.__repo_name.endswith('.git'):
-            self.__repo_project = self.__repo_name[:-4]
+        parsed = giturlparse.parse(self.__repo_url)
+        if len(parsed.protocols) > 0 and parsed.protocols[0] not in CommitMetric.supported_protocols:
+            raise Exception("Unsupported protocol %s", parsed.protocols[0])
```
etsauer, line 243: Can we raise a ValueError here, instead of a generic exception, and handle that specifically in the catch?

exporters/committime/collector_bitbucket.py
willowmck: line 189

exporters/committime/collector_base.py
willowmck: sure

willowmck: ValueError is in the latest revision

themoosman:
I tested the most recent changes and everything seems to work just fine.

themoosman (approved):
LGTM

PR #227, opened 2020-10-29
Helm test prometheus

jtudelag:

This PR adds a helm test hook in the form of a k8s job.
This job checks all the Prometheus targets are "UP".

To run the test once Pelorus has been deployed succesfully:
`helm test pelorus`

If this get merged, I would like to reactor testing documentation with this update, as it seems a bit disperse/hidden in the repo.

A couple of things to look into this PR:

- The container image i am using is from docker, it requites curl and jq
- I did bumped chart version manually, as it seems the pre-commit is not working in my laptop.
 

## Linked Issues?

related to #https://github.com/redhat-cop/pelorus/issues/36 

## Testing Instructions

@redhat-cop/mdt

### DIFF
--- a/charts/pelorus/templates/tests/pelorus-prometheus-test-job.yml
+++ b/charts/pelorus/templates/tests/pelorus-prometheus-test-job.yml
@@ -1,0 +1,78 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: pelorus-prometheus-test
+  annotations:
+    "helm.sh/hook": test
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  parallelism: 1
+  completions: 1
+  template:
+    metadata:
+      name: pelorus-prometheus-test
+    spec:
+      containers:
+      - name: prometheus-targets-test
+        image: registry.redhat.io/web-terminal-tech-preview/web-terminal-tooling-rhel8:1.0.1
+        env:
+        - name: PROM_HOSTNAME
+          value: "http://prometheus-pelorus:9090"
+        command:
+        - /bin/bash
+        - -c
+        - |
+          # Install jq. Not friendly with disconnected environments...
+          curl -so jq -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
+          chmod +x jq
+
+          # Check if all the discovered targets health are "UP".
+
+          # If this file exists, then I am inside a Pod.
+          _K8s_SA_TOKEN_FILE="/var/run/secrets/kubernetes.io/serviceaccount/token"
+          # Default Pometheus config settings
+          _PROM_API_TARGETS="/api/v1/targets"
+
+          echo -e "Checking Prometheus Targets Health.\n"
+
+          if [[ -f "${_K8s_SA_TOKEN_FILE}" ]]
+          then
+              echo -e "Executing tests inside a Kubernetes Pod.\n"
+              _POD_SA_TOKEN="$(cat "${_K8s_SA_TOKEN_FILE}")"
+
+          else
+              echo -e "Executing tests outside a Kubernetes Pod.\n"
+              _POD_SA_TOKEN=""
+          fi
+
+          # Composing vars
+          _PROM_URL="${PROM_HOSTNAME}${_PROM_API_TARGETS}"
+
+          # Show vars for debug purposes
+          echo -e "Final URL: $_PROM_URL \n"
+
+          # Get targets array
+          _PROM_TARGETS=$(curl -k -v -s "${_PROM_URL}" | ./jq -r -c '.data.activeTargets[]')
+
+          echo -e "\nChecking if all Prometheus targets are UP...\n"
+
+          IFS=$'\n'
+          for t in ${_PROM_TARGETS}
+          do
+
+              HEALTH=$(echo "$t" | ./jq -r '.health')
+              TARGET_URL=$(echo "$t" | ./jq -r '.scrapeUrl')
+              TARGET_POOL_NAME=$(echo "$t" | ./jq -r '.scrapePool')
+              if [[ "$HEALTH" != "up" ]]
+              then
+                  echo -e "Test failed, $TARGET_POOL_NAME = $TARGET_URL is not UP.\n"
+                  exit -1
+              else
+                  echo -e "Target $TARGET_POOL_NAME = $TARGET_URL is UP.\n"
+              fi
+
+          done
+
+          echo -e "All Prometheus targets are UP. Test OK!!\n"
+          exit 0
+      restartPolicy: Never
--- a/charts/pelorus/Chart.yaml
+++ b/charts/pelorus/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.2.10
+version: 1.2.11
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 1.2.2+13.g8d66bdd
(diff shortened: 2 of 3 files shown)


themoosman:
@jtudelag Can you rebase your branch?

themoosman (changes requested):
Can you rebase the branch.

jtudelag:
@themoosman done.

Still, The container image I am using is from docker hub, it requites curl and jq. Given the recent docker hub registry pull restrictions....Would it make sense to have a pelorus-test image with all the "testing" tools included? helm, bats, conftest, etc ,etc.
What do you think?

jtudelag:
Actually sth similar to this would work: https://github.com/redhat-cop/github-actions/blob/master/confbatstest/Dockerfile

themoosman:
```
[kmoos 🎩︎pelorus-jtudelag]$ helm test pelorus
Error: release: not found
```

themoosman (changes requested):
When running the test command, it never returns.  It just creates a job after job and no return code appears.

jtudelag:
@themoosman So can you paste the logs here? Are your prometheus targets up? Is any of them down? How many prometheus targets you have?

charts/pelorus/templates/tests/pelorus-prometheus-test-job.yml
```diff
@@ -0,0 +1,68 @@
...
+spec:
+  parallelism: 1
+  completions: 1
+  template:
+    metadata:
+      name: pelorus-prometheus-test
+    spec:
+      containers:
+      - name: prometheus-targets-test
+        image: docker.io/dwdraju/alpine-curl-jq # TODO: decide  what image to use...
```
themoosman, line 17: This might not work in some client environments.

themoosman:
> @themoosman So can you paste the logs here? Are your prometheus targets up? Is any of them down? How many prometheus targets you have?

What logs?  The cli command never returns, it just hangs.

jtudelag:
> > @themoosman So can you paste the logs here? Are your prometheus targets up? Is any of them down? How many prometheus targets you have?
> 
> What logs? The cli command never returns, it just hangs.

You mention this in a previous comment:

`When running the test command, it never returns. It just creates a job after job and no return code appears.`

So I was asking for the logs of that jobs.

charts/pelorus/templates/tests/pelorus-prometheus-test-job.yml
jtudelag: Please see my previous comment https://github.com/redhat-cop/pelorus/pull/227#issuecomment-721002502

themoosman: Could you use a pattern similar to this?  https://github.com/etsauer/pelorus/blob/grafana-snapshots/charts/pelorus/templates/dashboard-snapshot-cronjob.yaml#L58

jtudelag: Yes, I will do. On the other hand, that approach won't work on disconnected environments.

themoosman: I've created issue #237 to address the image issue.  At least now we're using the same pattern across all the jobs.

themoosman:
There is still something odd going on.  The jobs keeps failing even when the endpoints are healthy.  

So I added a debug line as in:

```
...
# Test all Prom targets are "UP"
          echo -e "\n Checking if all targets are UP...\n"
          for t in "${_PROM_TARGETS_HEALTH}"
          do
              echo "$t"
              if [[ "$t" != "up" ]]
              then
                  echo -e "Test failed, one target is not UP.\n"
                  exit -1
              fi

          done
...

```

and got the following:
```
Checking Prometheus Targets Health.

Executing tests inside a Kubernetes Pod.

Final URL: http://prometheus-pelorus:9090/api/v1/targets

*   Trying 172.30.241.213...
* TCP_NODELAY set
* Connected to prometheus-pelorus (172.30.241.213) port 9090 (#0)
> GET /api/v1/targets HTTP/1.1
> Host: prometheus-pelorus:9090
> User-Agent: curl/7.61.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Thu, 12 Nov 2020 15:17:36 GMT
< Transfer-Encoding: chunked
<
{ [10720 bytes data]
* Connection #0 to host prometheus-pelorus left intact

 Checking if all targets are UP...

up
up
Test failed, one target is not UP.
```

You can see the two `up` statements, but the job still fails.

jtudelag:
Hey @themoosman thanks for the review. The test was broken because of the quotes in "${_PROM_TARGETS_HEALTH}".

I have refactored it a bit to make it a bit more verbose, so when a target is not up, we know which one.

Hopefully now works...

themoosman:
The test is now working just fine.  I did notice one last thing.  When doing a `helm uninstall` the `pelorus-prometheus-test` job is not removed.

jtudelag:
I would say this is helm expected behaviour, as [helm uninstall](https://helm.sh/docs/helm/helm_uninstall/) applies to the helm release, and the tests are not part of the release.

I don't see every user installing Pelorus running the tests, just maybe if sth goes wrong, and I would say the main use case would be development purposes and CI/CD (integration tests?)?

Other option is use a hook deletion policy, as test is considered anothe rhook in helm: https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies

etsauer:
@themoosman @jtudelag couple of things. 

1. I agree that the cleanup is not super important, as this would mostly be used for CI purposes. However, from the annotations, it looks like we should expect it to be cleaned up like a hook. If we're going to have those annotations, they should work.
2. Tests aren't valuable unless they get run. And right now this PR does not include either automatic execution of this test, or instructions on running it. Without one of those, this code becomes tech debt. Let's make sure any tests we have in this repo also have a process in place for using them.

jtudelag:
@etsauer 

>     1. I agree that the cleanup is not super important, as this would mostly be used for CI purposes. However, from the annotations, it looks like we should expect it to be cleaned up like a hook. If we're going to have those annotations, they should work.

So, the current annotation `before-hook-creation` work as expected. The question is if we want to add any additional [deletion policy](https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies) annotation or not. I would argue that we don't want any deletion policy so we can always see if the test went ok or failed. Same behaviour happens to OCP Pod deployments for DCs. the Pod is always kept ;). 

 
>     2. Tests aren't valuable unless they get run. And right now this PR does not include either automatic execution of this test, or instructions on running it. Without one of those, this code becomes tech debt. Let's make sure any tests we have in this repo also have a process in place for using them.

I would rather document this first, and then add any sort of automation in a different PR later.

etsauer:
@jtudelag that's fine, and I agree with just documenting, and not integrating into tests yet (the reality is we can't test this automatically, because we don't yet have openshift environments to CI against). My point was that this PR should include updates to the developer docs to instruct developers/PR reviewers to actually USE this test.

etsauer:
@jtudelag looking at your README updates, it looks like you've made a lot of changes that have nothing to do with this new test. Could you clean that up please?

jtudelag:
@etsauer it is all MD formatting improvements/fixes, but ok.

etsauer:
@jtudelag this looks good, just need to bump the Chart version again.

etsauer (approved):

jtudelag:
@etsauer @themoosman  PTAL

mpryc:
I am closing this PR as a result of current cleanup of open issues/PRs.

Helm testing is still valid scenario, however the way we would want to implement this have changed, because we are creating Prometheus and Grafana deployments via OLM dependency in the Pelorus operator.

PR #229, opened 2020-10-30
Fallback to BuildConfig when Build does not have git uri

### LINKED ISSUE #222: Support Docker builds
We are currently using Docker based builds.  What we found is that setting the spec.source.git.uri in the BuildConfig is not preserved on the corresponding Build resource.  What would be required would be to walk back from the Build resource reference to the corresponding BuildConfig to find the git uri.

willowmck:
This change allows fallback behavior in the commit time exporter to look for the Git URI in the BuildConfig object if the Build object does not specify this information.

In the case it is not found in either location, then there is no change in behavior.

## Linked Issues?

resolves #222 

## Testing Instructions

Although this is easily reproducible in my customer's environment with a Docker strategy BuildConfig, I'm not sure how it happens that the spec.source.git information is not preserved from the BuildConfig to the Build object.  See the reported issue this resolves for an example.

@redhat-cop/mdt


themoosman:
@willowmck Do you have a sample buildconfig we can use for testing?

willowmck:
@themoosman - take a look at https://github.com/redhat-cop/pelorus/issues/222#issuecomment-719649265.  If you need something less verbose, I'll see if I can track something down.

malacourse:
@willowmck I reviewed and looks good.  Could you make the chart version 1.2.11 then I think we can merge.

willowmck:
> @willowmck I reviewed and looks good. Could you make the chart version 1.2.11 then I think we can merge.

Will do. Thanks @malacourse !

malacourse (approved):
Tested and works as expected.

PR #231, opened 2020-11-03
Commit Time: Support JenkinsPipelineStrategy builds with embedded jenkinsfiles

### LINKED ISSUE #145: How could Pelorus work in a JenkinsPipelineStrategy buildconfig that executes a binary build to build the image?

etsauer:
## Describe the behavior changes introduced in this PR\

Currently the commit time exporter fails if it finds a jenkins buildConfig that doesn't have a Git URL set. This code change will allow scanning of environment vars (pipeline params) as well, and handles failures better.

## Linked Issues?

resolves #145 

## Testing Instructions

@redhat-cop/mdt


malacourse:
If you change the JenkinsPIpeline Env Param to point to a different GIT repo after a build has taken place the old Repo is getting picked up unless you delete the previous builds.  Not sure this is a likely scenario but I did come across it.  Seems the latest build should be used.

Also could you add comments on the regex being used for us regex dummies. What it is filtering on?

shaheinm (dismissed):
Haitham and I tested this with our own demo - it works quite well

shaheinm:
@etsauer I merged the latest master into this PR from the UI here - I kept changes from both branches: 
- master had a new function `_get_repo_from_build_config` added by @willowmck in #229
- your branch had changes to the same file in the `get_repo_from_jenkins` function)

shaheinm (approved):
Approving as it works in my tests, but I would like input from @willowmck at least to make sure everything lines up with what was done in #229

willowmck:
> Approving as it works in my tests, but I would like input from @willowmck at least to make sure everything lines up with what was done in #229

Those changes look good to me @shaheinm .  Thanks!

PR #236, opened 2020-11-11
Multi cluster docs

### LINKED ISSUE #132: Document a multi-cluster architecture for Pelorus, and how to deploy it.
Include details on configuring exporters to watch a restricted set of namespaces.

malacourse:
Redo Docs to reflect new multi cluster architecture.  

## Linked Issues?

resolves #132 

## Testing Instructions
Deploy Pelorus using Thanos with a shared S3 bucket across multiple clusters.  Verify metrics are populated from both clusters in the Grafana dashboard.

@redhat-cop/mdt


etsauer:
Need to include actual deployment instructions for setting up a multi-cluster pelorus, including:

* Helm chart value file configs (one per cluster)
* Deployment of different exporters to different clusters

docs/Install.md
```diff
@@ -77,27 +77,39 @@ If you don't have an object storage provider, we recommend [MinIO](https://min.i
 
 ### Deploying Across Multiple Clusters
 
-By default, this tool will pull in data from the cluster in which it is running. The tool also supports collecting data across mulitple OpenShift clusters. In order to do this, we need to point the Pelorus instance at these other clusters.
+By default, this tool will pull in data from the cluster in which it is running. The tool also supports collecting data across mulitple OpenShift clusters. In order to do this, the thanos sidecar can be configured to read from a shared S3 bucket accross clusters. See [Pelorus Multi-Cluster Architecture](/page/Architecture.md) for details. You define exporters for the desired meterics in each of the clusters which metrics will be evaluated.  The main cluster's Grafana dashboard will display a combined view of the metrics collected in the shared S3 bucket via thanos.
```
etsauer, line 80: I think we're still missing an important piece here, which is implied but never actually stated, which is this: The way that you deploy Pelorus across multiple clusters is by deploying an independent Pelorus stack to each cluster you want to collect data from, and then tie them all together by pointing them to a single Object Storage backend. So to show that, I think we should show the entire `values.yaml` file example for each cluster we are deploying to, to show what is the same and what is different in each cluster.

Another idea is that we could use the deployment labels to designate the different types of deployments. For example:

```
deployment:
  labels:
    app.kubernetes.io/component: development
    app.kubernetes.io/name: pelorus
    app.kubernetes.io/version: v0.33.0
```

vs

```
deployment:
  labels:
    app.kubernetes.io/component: production
    app.kubernetes.io/name: pelorus
    app.kubernetes.io/version: v0.33.0
```

docs/Install.md
```diff
@@ -77,27 +77,39 @@ If you don't have an object storage provider, we recommend [MinIO](https://min.i
...
+exporters:
+  instances:
+  ... 
+```
 
-    extra_prometheus_hosts:
-      - id: "ci-1"
-        hostname: "prometheus-k8s-openshift-monitoring.apps.ci-1.example.com"
-        password: "<redacted>"
+#### Configure exporters in additinal clusters to push data into the shared S3 Bucket.
```
etsauer, line 97: This title is misleading. It's not the exporters themselves that push data into object storage. Its the prometheus stack. The thing we want to show here is the values file for the second cluster, showing that the same bucket is being used, but different set of exporters deployed here vs in Dev.

docs/Install.md
etsauer: I believe we should assume that the development cluster is the target here, so we should deploy the commit and failure exporters here.

docs/Install.md
etsauer: Here we are implying the production cluster, so the deployment exporter should be the only exporter deployed here.

etsauer (dismissed):
@malacourse left you some more comments. In addition, why does this PR show that a code change is being made to `collector_base.py`? This seems unrelated.

exporters/committime/collector_base.py
```diff
@@ -106,17 +105,8 @@ def get_metrics_from_apps(self, apps, namespace):
...
-            if jenkins_builds:
-                # we will default to the repo listed in '.spec.source.git'
-                repo_url = jenkins_builds[0].spec.source.git.uri
-
-                # however, in cases where the Jenkinsfile and source code are separate, we look for params
-                git_repo_regex = re.compile(r"(\w+://)(.+@)*([\w\d\.]+)(:[\d]+){0,1}/*(.*)")
-                for env in jenkins_builds[0].spec.strategy.jenkinsPipelineStrategy.env:
-                    result = git_repo_regex.match(env.value)
-                    if result:
-                        repo_url = env.value
```
themoosman, line 119: Why is this code being removed?  It seems valid:  https://github.com/redhat-cop/pelorus/blob/master/exporters/committime/collector_base.py#L109-L119

docs/Install.md
themoosman: It looks as if Eric's concern is addressed.

themoosman (changes requested):
The doc looks good and I tested as much of it as I could.  Just one question on the collector_base.py file.

themoosman (approved):
LGTM

themoosman (approved):

PR #242, opened 2020-12-14
Image sha fix

### LINKED ISSUE #222: Support Docker builds

willowmck:
This change builds upon PR https://github.com/redhat-cop/pelorus/pull/229 and introduces 2 new changes:

1. Modify the commit time exporter to look for the `build.spec.output.to.kind` and if it equals `DockerImage` then take the commit_sha instead of the deploy_sha to match against.
2.  Modify the deploy time exporter to allow regular expression matching against a commit sha without the prefix `sha256:`.

## Linked Issues?

resolves #222 

## Testing Instructions

Create a BuildConfig where the `spec.output.to.kind` is set to `DockerImage`.  Go through the normal build & deploy process to see the linkage created.  If you need an example to pull from I can supply one generically from my customer.

@redhat-cop/mdt


themoosman (changes requested):
Returning to @willowmck after a conversation for some additional work.

willowmck:
@themoosman - I made this a configuration since we are using the commit hash and not using image streams.  In this change, I modified the documentation to point to the new variable NO_IMAGE_STREAM for the commit time exporter and also made some modifications to the startup so that the variables passed through the factory were a bit cleaner.

mpryc:
I am closing this PR as a result of current cleanup of open issues/PRs.

Collection of the commit SHA may happen now from the Image label/annotation, so different approach to resolve similar use-case:
   https://pelorus--820.org.readthedocs.build/en/820/GettingStarted/configuration/ExporterCommittime/#provider-git-options

Issue #246, opened 2021-02-08
Jenkins on OpenShift and mult-branch


tdbeattie:
Pelorus depends on jenkinsPipelineStrategy and is designed to work with Jenkins builds in OpenShift. We have some engagements that use Jenkins on top of OpenShift and uses multi branch strategy. 

This issue is to discuss and align on a solution that will enable Pelorus to work with Jenkins using multi branch.

mpryc:
As noted in the https://bugzilla.redhat.com/show_bug.cgi?id=1785291 the JenkinsPipelineStrategy is deprecated from OpenShift version 4.3.z.

With that I am closing this issue. Note that we currently have a way of configuring Pelorus with images using image exporter documented at:
https://pelorus.readthedocs.io/en/latest/GettingStarted/configuration/ExporterCommittime/#using-commit-time-with-images

PR #250, opened 2021-02-14
Generic CommitTime Exporter with Webhook - GitHub Support

haithamshahin333:
This PR introduces a webhook endpoint to receive build information that can then be exported to the Pelorus dashboard with the generic exporter. This can be used when builds occur outside of OCP.

## Linked Issues?

related to #246 

## Testing Instructions

deploy to dev cluster using helm charts

@etsauer @shaheinm Further Areas for Improvement:

1. Templates for secrets autogenerate random values - this may be problematic with helm install/upgrade since the secrets would then be overwritten.
2. Would be good to refactor generic exporter such that it is in the committime folder since it leverages same collector classes.
3. Add support for gitlab and bitbucket
4. Requires tests

@redhat-cop/mdt


etsauer:
@haithamshahin333 to be clear on this. This code only works with CommitTime correct? Its not 'generic' to the metric?

haithamshahin333:
correct @etsauer good callout - I think the pattern could be leveraged for any metric but right now only implemented for CommitTime.

haithamshahin333:
@etsauer added a basic example pipeline in the samples directory that demonstrates usage of this new exporter alongside the deployed webhook. The Jenkinsfile gives an example for how to post build information to the webhook as part of a pipeline.

shaheinm:
Linking this issue here because it's the first instance I see where the promtool tests fail because of this bug - https://github.com/prometheus/prometheus/issues/8494

haithamshahin333:
@etsauer @shaheinm Here are instructions for how to test this new feature with the sample application that is part of the PR (https://github.com/haithamshahin333/pelorus/blob/master/samples/README.md). I added it as a README within the samples directory so that others in the future can describe how to deploy different use cases.

haithamshahin333 (commented):
@shaheinm may be an issue with this update, doesn't look like the action was able to start: https://github.com/konveyor/pelorus/actions/runs/573100473

shaheinm:
> @shaheinm may be an issue with this update, doesn't look like the action was able to start: https://github.com/konveyor/pelorus/actions/runs/573100473

Yeah, it looks like there were lots of changes from the version that was being used (a v1 release candidate) and the latest (v2.0.1) for the action.

etsauer:
> > @shaheinm may be an issue with this update, doesn't look like the action was able to start: https://github.com/konveyor/pelorus/actions/runs/573100473
> 
> Yeah, it looks like there were lots of changes from the version that was being used (a v1 release candidate) and the latest (v2.0.1) for the action.

@haithamshahin333 @shaheinm will one of you open an issue on this? We can handle it as a followup.

shaheinm:
> > > @shaheinm may be an issue with this update, doesn't look like the action was able to start: https://github.com/konveyor/pelorus/actions/runs/573100473
> > 
> > 
> > Yeah, it looks like there were lots of changes from the version that was being used (a v1 release candidate) and the latest (v2.0.1) for the action.
> 
> @haithamshahin333 @shaheinm will one of you open an issue on this? We can handle it as a followup.

@etsauer I've already got an update on haitham's master branch in PR that'll make into this one if accepted

.github/workflows/prometheus-rules.yml
```diff
@@ -11,6 +11,6 @@ jobs:
         go-version: '^1.13.1' # The Go version to download (if necessary) and use.
     - run: |
         go version
-        go get github.com/prometheus/prometheus/cmd/promtool
+        GO111MODULE=on go get github.com/prometheus/prometheus/cmd/promtool
```
roidelapluie, line 14:
```suggestion
        GO111MODULE=on go get github.com/prometheus/prometheus/cmd/promtool@release-2.25
```

etsauer:
resolves #246

shaheinm:
I've updated the chart-lint.yml workflow to use the latest version of helm/chart-testing-action - this is necessary as older versions use older version of helm, in which the `lint` command has no knowledge of the `lookup` function in charts (used a bit for this work to ensure k8s secrets are not modified every time a helm update occurs).

This also required that I move the dependency charts (webhooks and exporters) out of the `charts/pelorus/charts` directory and into a `charts/pelorus/subcharts` directory, as helm 3 expects to use the charts directory name for remote dependencies that are downloaded.

shaheinm:
Also updated the conftest action to update dependent charts prior to testing (it runs `helm template` which will fail without the update - should stay as long as the dependency pattern is used)

haithamshahin333:
Replaced by https://github.com/konveyor/pelorus/pull/253

PR #251, opened 2021-02-15
New Pelorus Logo

jameslabocki:
Changing Pelorus logo on readme.md


etsauer:
I think we should re-visit the logo change when its time to refresh on Pelorus swag. We just finished a bunch of branding work with the current logo, and I would like to let it cool and use the swag we have before we change it up.

I think the Pelorus core team would also like to participate in making sure any new logos communicate the same message we worked on with the current one.

PR #265, opened 2021-03-23
Image builds

shaheinm:
Use [s2i Github Actions](https://github.com/redhat-actions/s2i-build) to create images for pelorus exporters, and push them to konveyor Quay.io repository.

To use this properly, we'll need access to the konveyor organization in Quay, and three new repos will need to be set up within:
- quay.io/pelorus/committime-exporter
- quay.io/pelorus/deploytime-exporter
- quay.io/pelorus/failure-exporter

We'll also need to set up credentials in github to access and push to Quay.

### DIFF
--- a/.github/workflows/build-pelorus.yaml
+++ b/.github/workflows/build-pelorus.yaml
@@ -1,0 +1,94 @@
+name: Pelorus Exporter Build
+
+on:
+  push:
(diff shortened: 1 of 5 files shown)


etsauer (approved):
lgtm

PR #266, opened 2021-03-23
updated gh action push-to-registry to v2, modified spec to match

shaheinm:
quick fix

Action documentation is [here](https://github.com/marketplace/actions/push-to-registry)


etsauer (approved):
lgtm. let's try this again!

PR #281, opened 2021-05-13
Add mockserver use instructions

### LINKED ISSUE #270: Define a way to test the exporters using mock responses
What its need it:

- Create a directory structure that can hold the sample for GitLab, Github, Bitbucket, Gitea, Jira and Service Now
- The tool or framework need to be ligthweigth
- The tests and ramp up of the mock server need to be triggered be GitHub Actions

The initial tools proposition are:

- [Castemocks](https://github.com/castlemock/castlemock)
- [Hoverfly](https://github.com/spectolabs/hoverfly)

fmenesesg:
This PR introduces the instructions about how to start the mock server and how to create new scenarios.

## Linked Issues?

resolves #270 

## Testing Instructions

Follow the instruction on the README file included in the mocks folder

@redhat-cop/mdt


exporters/committime/collector_github.py
```diff
@@ -28,7 +28,7 @@ def get_commit_time(self, metric):
             return None
 
         url = self._prefix + metric.repo_group + "/" + metric.repo_project + self._suffix + metric.commit_hash
-        response = requests.get(url, auth=(self._username, self._token))
+        response = requests.get(url, auth=(self._username, self._token), verify=False)
```
etsauer, line 31: We'll need to parameterize this before we can merge it.

fmenesesg:
@etsauer I add the "skip tls validation" parametrization

exporters/committime/collector_base.py
```diff
@@ -14,14 +14,16 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
     This class should be extended for the system which contains the commit information.
     """
 
-    def __init__(self, kube_client, username, token, namespaces, apps, collector_name, timedate_format, git_api=None):
+    def __init__(self, kube_client, username, token, namespaces, apps, collector_name,
+                 timedate_format, git_api=None, tls_verify=None):
```
KevinMGranger, line 18: I think we said we wanted to generalize `git_api` to something like `api_url`, right?

As much as I hate to bikeshed, changing parameter names in python is a breaking change because of keyword arguments. Not that it matters when we're the only consumer 😛 

Also, have we tested what happens with `verify=None`? The [requests](https://docs.python-requests.org/en/master/api/#requests.request) docs say it has to be a boolean or string. I think we should have it be `tls_verify=True` just to be safe.

mocks/README.md
```diff
@@ -0,0 +1,140 @@
...
+export TLS_VERIFY=False
+```
+---
+**NOTE**
+
+The env TLS_VERIFY is needed because the mocks server use a self-signed certificate.
+
+---
+
+Because the commit exporters search on Openshift the information from the build, first you need to execute "oc login" against the mock server
```
KevinMGranger, line 44: I want to make sure I understand-- this uses mockoon's proxy mode? I think we should document that, and note that the server to proxy to is hardcoded in mockoon-- do we have some way of templating that out?

fmenesesg: Is not using mockoon on proxy mode, when you start the mock server that is disable, its only need to be enable when your are creating new scenarios but this is not the case

KevinMGranger: So then this isn't necessary for setup, right? Only for development? Then maybe we should document it elsewhere?

fmenesesg: To start the mock server, they are not necessary, but in this case, the instructions are to test an exporter against the mock server, in that case, you need to set up the env vars and login into the server as if you were using Openshift instance.

KevinMGranger (approved):

Issue #225, opened 2020-10-20
Should exporters be organized by Provider instead of by metric?


etsauer:
We've been discussing how we are going to refactor our exporter directory structure to fix some of the upside-down-ness of our imports, but I'm also starting to wonder if the architecture itself might be flawed. We continue adding new providers, and also look to add more types of metrics outside of just Software Delivery, I think that we are going to end up with a many-to-many relationship between providers and exported metrics, but to any given customer.. they probably will only care about one provider of each of the metrics. 

Let's look at software delivery with the providers we have and some that we might implement in the future

```
committime
- OpenShift (or Kubernetes or Public Cloud)
  - GitHub
  - Gitlab
  - Bitbucket
deploytime
- OpenShift
- (Kubernetes)
- (Public Cloud)
failure
- Jira
- ServiceNow
- (Trello)
- (Version1)
```

And now some future metrics we'll likely be adding, slightly less well defined:
```
Platform Adoption Metrics:
- OpenShift
- Kubernetes
- Public Cloud

Flow Load (breakdown of work types: bugs, feature, risk, debt)
- Jira
- SericeNow
- Version1
- Trello
- GitHub/Gitlab/Bitbucket Issue tracker?
```

It seems like we are going to be moving towards a model where we are gathering multiple metrics from the same tools, and I don't think it will make sense to have an exporter per metric anymore. Instead I propose this:

* We look to refactor so that we have a module/library that contains all of our individual metrics defined in abstract classes.
* We refactor our exporters such that we have one exporter per provider/tool, and it can implement ANY of the metrics that are appropriate.

PR #286, opened 2021-06-09
Reorganize some exporter-related code, introduce an exporter-specific unit test, and start typing

KevinMGranger:
## What's it do?

This PR introduces a few different changes that should improve confidence in our code. They're a bit rolled into each other because they support each other, but I can make separate PRs or rebase the commits to be separate if that would be cleaner.

1. Reorganization
Instead of symlinking to common pelorus code, we move the pelorus module to a fully-fledged package, and make it installable. The UBI python image should install the package for us automatically.
This opens the door to moving common backend and/or metric code into the pelorus package, and lets us keep `$exporter/app.py` as a facade to just wire the two together.
I want to make this as cleanly organized as a python package can be-- however, I don't want to break compatibility for those who already are using pelorus-- that means keeping all python-related stuff confined to `exporters`.

2. Exporter unit test
This adds a unit test for the deploytime exporter. As part of it, some of the exporter is refactored to make it easier to unit test (use a mock for the dynamic client, etc.)
While reading through some of the exporter code, I wanted to refactor some of the more intricate parts-- but I wasn't going to touch it without unit tests in place to catch regressions! This is the start of that. It's far from a full test, so I'll add more once if think this is a good structure to go by.

3. Start typing
Python now supports type annotations, which can be checked by external tools (in this case, mypy). It's not only a form of additional documentation, but allows us to catch errors. We anticipate having devs contribute to this porject who might not have a lot of experience with python, and this will help catch issues that might be confusing. There was even a mistake I made while making the unit test that would have been caught by it!
I didn't include mypy in `requirements-dev.txt` yet because I want developer use of it to be opt-in.
mypy is installed in our github workflow, however, so I need to make sure it's configured correctly to ignore modules we haven't gotten to yet.

4. Add `attrs`
attrs lets us easily write self-documenting (generate `__init__` from type annotations), fully-correct (`__eq__`, `__str__`, etc.) classes without the boilerplate (and with no runtime cost-- it's all done at import time). If we're going to present metric structure to exporter developers that they need to fit to, I think it helps us communicate cleanly and avoid errors. It also makes whipping up classes for mocked data easier.

## Related Issues

related to #225 

## Testing Instructions

The unit testing process remains the same, but you'll need to run `pip install -e ./exporters` first.

@redhat-cop/mdt

## Remaining Work

- [x] ~~I want to make the unit test more robust, checking edge cases.~~ (~~maybe~~ in a separate PR)
- [x] I want to test this myself in the cluster to make sure the structural changes don't affect the source2image workflow.
- [x] I want to see the results for the github actions flow to make sure the structural changes and introduction of typing are correct for it.
  - [x] FOLLOWUP: need to bump to version 3.9, and investigate formatting and/or linting checks.
- [x] Some of the typing code would only work on Python 3.9, but making it work in 3.8 is an easy fix. I just want to ask if we care about keeping 3.8 compatibility, or if we can just make the image 3.9 based.


fmenesesg (commented):
LGTM I have test locally the changes and I haven't found any problems with it

fmenesesg (approved):
LGTM everything work while I was testing this PR.

PR #291, opened 2021-06-23
Switch linter tool

KevinMGranger:
## Describe the behavior changes introduced in this PR
Switch from pylama to pylava, fixing #288.

PR #292, opened 2021-06-24
Cache dependency installs in CI

KevinMGranger:
This should speed up python CI runs.

Depends on / based off of https://github.com/konveyor/pelorus/pull/291.


KevinMGranger:
Unfortunately, it looks like the slowest part of dependency installs is running pip itself, and we need to run pip in order to put bins in the right place.

PR #294, opened 2021-06-30
Format all python code with black and isort, check as part of CI.

KevinMGranger:
Compared to https://github.com/konveyor/pelorus/pull/291, this PR formats all of the python code in one go. This is the approach [recommended by Black](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html).

It will simplify the approach taken in #291 once I rework it.

This also includes a ".git-blame-ignore-revs" file, which will allow git blame to still properly report the original authors of each line:
`git blame important.py --ignore-revs-file .git-blame-ignore-revs`

One can configure git to automatically read it:
`git config blame.ignoreRevsFile .git-blame-ignore-revs`

I've also added a script to make formatting (and checking that formatting) easier-- `./scripts/format [--check]]`. Checking is also performed as part of CI.


fmenesesg (approved):
Everything works great and LGTM

PR #296, opened 2021-07-02
Committime github provider integration test

KevinMGranger:
Introduce a GitHub committime integration test and runs it during CI. This will run mockoon to mock the GitHub and OpenShift APIs.

There is currently a strange bug where running integration tests alongside unit tests will cause the integration tests to fail, but this doesn't happen when they're run separately. They're run separately in CI, so I'm not fully concerned, but I'm looking into it now.

## Testing Instructions

Start mockoon's server, then run: `pytest -m integration`

To see the bug I mentioned, run: `pytest -m`

@redhat-cop/mdt

### DIFF
--- a/exporters/tests/integration/test_committime.py
+++ b/exporters/tests/integration/test_committime.py
@@ -9,13 +9,6 @@ from committime.collector_github import GitHubCommitCollector
 from kubernetes.client import ApiClient, Configuration
 from openshift.dynamic import DynamicClient
 
-# export GIT_USER=gituser
-# export GIT_TOKEN=gittoken
-# export GIT_API=localhost:3000
-# export LOG_LEVEL=DEBUG
-# export NAMESPACES=basic-nginx-build,basic-nginx-dev,basic-nginx-stage,basic-nginx-prod
-# export TLS_VERIFY=False
-
 
 def setup_collector():
     kube_config = Configuration()
--- a/exporters/tests/integration/test_committime.py
+++ b/exporters/tests/integration/test_committime.py
@@ -87,7 +87,7 @@ class CommitMetricEssentials:
 
 
 @pytest.mark.integration
-def test_github_collector():
+def test_github_provider():
     collector = setup_collector()
 
     actual = [
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,28 @@
+name: Integration Tests
+on: [push, pull_request]
+
+jobs:
+  integration-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Python 3.9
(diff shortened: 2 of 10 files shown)


fmenesesg (dismissed):
LGTM great work.

etsauer:
@KevinMGranger so, I can follow the integration/test_committime.py file, and I understand the expected results in that file, but what I don't see is how we are loading the `actual` data into the mock kube api server. Was that in a different PR?

exporters/tests/integration/test_committime.py
```diff
@@ -0,0 +1,100 @@
...
+        args = {
+            key: getattr(cm, key)
+            for key in "name namespace commit_hash commit_timestamp image_hash".split()
+        }
+
+        return CommitMetricEssentials(**args)
+
+
+@pytest.mark.integration
+def test_github_provider():
```
etsauer, line 90: this will result in a single "test" per provider. How will be grow this to account for various scenarios such as:

* Commit data from a buildConfig
* Commit data from a Tekton Pipeline
* Handling of commits that might be missing data

etc. It looks like this structure only allows for us to add more data to `expected_commits()`, but in the test results we end up missing context, as they are all lumped into one single test that passes or fails, and we won't know why. Is now the right time to expand this framework, or is that a followup?

exporters/tests/integration/test_committime.py
```diff
@@ -0,0 +1,100 @@
...
+from typing import Optional
+
+import attr
+import pytest
+from committime import CommitMetric
+from committime.collector_github import GitHubCommitCollector
+from kubernetes.client import ApiClient, Configuration
+from openshift.dynamic import DynamicClient
+
+# export GIT_USER=gituser
```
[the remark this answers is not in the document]
KevinMGranger: Yes, I should delete them. They were just to make sure the test setup data matched the mock's data in its README.

exporters/tests/integration/test_committime.py
KevinMGranger: Since we only have mock data for the GitHub provider, this is all we have to work off of. Generalizing this _should_ happen, but later-- premature generalization is the root of all evil, or something like that.

That's part of the work I'm doing right now, actually-- as part of making a GitLab integration test next, I wanted to see if we could standardize on a set of data / pipeline actions / etc to make developing these tests easier.

etsauer: ok, I get it now. I was thrown by the name of this PR. this isn't integration tests for the entire committime exporter, just for the github provider. With that being the scope of this, then I think i'm fine as-is.

etsauer:
@KevinMGranger looks like you've got a merge conflict to resolve before we can merge.

KevinMGranger:
@etsauer Merge conflicts fixed

etsauer (approved):
LGTM

PR #302, opened 2021-07-07
Make PR checks more selective based on paths

KevinMGranger:
Makes PR checks only trigger if their relevant paths are modified.


etsauer (approved):
lgtm

Issue #304, opened 2021-07-15
failure-exporter limited to only 50 results


rmarting:
Testing the integration with Jira Atlassian I found that exporter is limited to only 50 results.

Checking the log of failure-exporter pod you could see that the Query API endpoint add variables to limit the maximum number of results:

```log
---> Running application from Python script (failure/app.py) ...
Initializing Logger wit LogLevel: DEBUG
07-15-2021 13:31:00 INFO     ===== Starting Failure Collector =====
07-15-2021 13:31:00 INFO     Server: https://****************
07-15-2021 13:31:00 INFO     User: **************************
07-15-2021 13:31:00 DEBUG    Starting new HTTPS connection (1): ****************:443
07-15-2021 13:31:00 DEBUG    https://****************:443 "GET /rest/api/2/serverInfo HTTP/1.1" 200 None
07-15-2021 13:31:01 DEBUG    https://****************:443 "GET /rest/api/2/field HTTP/1.1" 200 None
07-15-2021 13:31:01 DEBUG    Starting new HTTPS connection (1): ****************:443
07-15-2021 13:31:01 DEBUG    https://****************:443 "GET /rest/api/2/serverInfo HTTP/1.1" 200 None
07-15-2021 13:31:01 DEBUG    https://****************:443 "GET /rest/api/2/field HTTP/1.1" 200 None
07-15-2021 13:31:04 DEBUG    https://****************:443 "GET /rest/api/2/search?jql=type%3Dbug+and+priority%3Dhighest&startAt=0&validateQuery=True&maxResults=50 HTTP/1.1" 200 None
```

The number of issues found is limite to only 50 entries, but my Jira instance manages hundred of projects with thousands of issues. In my case, the issues that I want to match in Pelorus do not appear in the results, so the Failure Rate metric is not updated for my use case.

How could I set up that value? It could be great to allow to setup the maximum number of issues, or at least execute that query for each project created in Jira. Other idea could be allow to set up the projects in Jira to extract metrics.

Maybe adding some properties in the deployment such as:

* MAX_RESULTS: Maximun number of results to query
* PROJECT: Literal or expression to define which projects in Jira should be checked to get metrics.

etsauer:
hi @rmarting yes this definitely looks like a bug to me. It appears that at one point we did add an optional PROJECT propery, but we never implemented it. https://github.com/konveyor/pelorus/blob/master/exporters/failure/app.py#L31

rmarting:
[...]

PR #308, opened 2021-07-16
Querying issues by a list of projects in JIRA

rmarting:

This PR uses the `PROJECT` environment variable of the failure-exporter to set the list of project keys in JIRA to extract metrics.

This feature will be valuable when you have a large organization with hundreds of projects but you only want to extract metrics from a subset of them. Many projects in that large organization could be used for other kind of projects not related with IT or  software developed or deployed.

*NOTE*: I am not an python expert developer, so an enhancement of this could be define PROJECT, not only as a comma separated string, also as a regular expression. (Maybe a second iteration of this PR)

## Linked Issues?

related to #304 

@redhat-cop/mdt

### DIFF
--- a/exporters/failure/collector_jira.py
+++ b/exporters/failure/collector_jira.py
@@ -17,7 +17,7 @@ class JiraFailureCollector(AbstractFailureCollector):
     def __init__(self, user, apikey, server):
         if not os.environ.get("PROJECT"):
             logging.info(
-                "Missing Project Field Parameter, querying all projects."
+                "Missing Project Field Parameter, querying all projects.",
             )
             self.project_field = None
         else:
--- a/exporters/failure/collector_jira.py
+++ b/exporters/failure/collector_jira.py
@@ -31,8 +31,8 @@ class JiraFailureCollector(AbstractFailureCollector):
     def search_issues(self):
         options = {"server": self.server}
         # Connect to Jira
-        jira = JIRA(options, basic_auth=(self.user, self.apikey))                
-        # TODO FIXME This may need to be modified to suit needs and have a time period.        
+        jira = JIRA(options, basic_auth=(self.user, self.apikey))
+        # TODO FIXME This may need to be modified to suit needs and have a time period.
         query_string = "type=bug and priority=highest"
         if self.project_field is not None:
             query_string = query_string + " and project in (" + self.project_field + ")"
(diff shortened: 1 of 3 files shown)


exporters/failure/collector_jira.py
```diff
@@ -14,6 +15,17 @@ class JiraFailureCollector(AbstractFailureCollector):
...
+            logging.info(
+                "Missing Project Field Parameter, querying all projects."
+            )
+            self.project_field = None
+        else:
+            logging.info(
+                "Querying issues from only '%s' projects.",
+                os.environ.get("PROJECT"),
+            )
+            self.project_field = os.environ.get("PROJECT")
```
KevinMGranger, line 28: We typically try to read all configuration within the exporter app.py itself, and pass that in to the class via a parameter-- it makes testing easier. But this is more or less the pattern we'd want to follow.

KevinMGranger (changes requested):

exporters/failure/collector_jira.py
rmarting: As I could not find a pattern about that, so I review the Service Now Collector and replicated the same structure from the [collector_servicenow.py](https://github.com/konveyor/pelorus/blob/master/exporters/failure/collector_servicenow.py#L24) file.

In both cases make sense to initialize the class with the specific parameters used, otherwise the constructor could be so complex to manage the different properties (requiered and optional).

WDYT?

etsauer: @rmarting the pattern we would recommend would be to follow something similar to what we did for namespaces in the committime exporter: https://github.com/konveyor/pelorus/blob/master/exporters/committime/app.py#L65-L73

Also, since PROJECT wasn't used anyway, let's go ahead and make it the plural, PROJECTS so that we've got more correct nomenclature.

rmarting: @etsauer @KevinMGranger Refactored following your recommendations. Hope now it is aligned with the patterns wanted.

KevinMGranger: Holding off on approving until we can test more but it looks good overall.

rmarting (commented):
WDYT? Do my latest changes fix the reviews?

KevinMGranger (approved):

Issue #303, opened 2021-07-12
deploytime-exporter build crashing


GuilhermeCamposo:
deploytime-exporter Buildconfig is using registry.access.redhat.com/ubi8/python-38, but source code is requiring Python >= 3.9.0

rmarting:
I got the same issue testing in my local CRC instance.

Builder output log:

```text
# Omitted output
Storing signatures
Generating dockerfile with builder image registry.access.redhat.com/ubi8/python-38
STEP 1: FROM registry.access.redhat.com/ubi8/python-38
# Omitted output
---> Installing application ...
Processing /opt/app-root/src
ERROR: Package 'pelorus' requires a different Python: 3.8.6 not in '>=3.9'
WARNING: You are using pip version 19.3.1; however, version 21.1.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
error: build error: error building at STEP "RUN /usr/libexec/s2i/assemble": error while running runtime: exit status 1
```

rmarting:
It seems that this issue is fixed in `master` branch. I got the issue using the 1.3.0.1 tag version. I could be great to release a minor version with this fix.

etsauer:
@KevinMGranger when you get a chance could you cut a 1.3.0.2?

KevinMGranger:
I'll be making a new release soon, but I want to verify-- are you sure you're using the 1.3.0.1 tag @rmarting ? `setup.py` isn't in that tag, so it shouldn't be able to require `>=3.9`.

It's possible you're using the 1.3.0.1 tag for the helm charts, but the exporters are pointing to master in the `values` file. Can you verify what you're seeing for each `exporters.instances[*].source_ref`?

This does bring up a good point: that we should reconsider having the docs and defaults pointing at master.

rmarting:
@KevinMGranger I downloaded the 1.3.0.1 version from [here](https://github.com/konveyor/pelorus/releases/tag/1.3.0.1). However the issue is caused by the definition of a DeploymentConfig of the exporters.

This [line](https://github.com/konveyor/pelorus/blob/1.3.0.1/charts/pelorus/charts/exporters/templates/_buildconfig.yaml#L27) defined the `python-38` image. So this is the root cause of the issue. That line in `master` is defined as `python-39` and works successfully.

etsauer:
[...]

PR #315, opened 2021-08-20
Set default exporter source_ref to "stable", remove appVersion from chart

KevinMGranger:
This will address the core issue we saw in  #303, where pinning your repo to a specific tag still has the exporters pointed at `master`-- causing things like python code versions and container versions to get out of
sync.

Making `stable` the default will make for a better user experience, especially for updates.

I've created the [stable branch](https://github.com/konveyor/pelorus/tree/stable) which right now points to the same commit as the [v1.3.0](https://github.com/konveyor/pelorus/tree/v1.3.0) tag.

This also removes the appVersion from the chart. I don't think it makes sense to keep track of it exporter-wise, since users specify their own version with `source_ref`.

It might make sense to keep track of it for the grafana configuration? I can leave this change out if we want.

@redhat-cop/mdt

### DIFF
--- a/_test/pre-commit
+++ b/_test/pre-commit
@@ -7,11 +7,6 @@ bump_version() {
     echo "Updated chart ${chart} version to ${version}."
 }
 
-set_app_version() {
-    version=$(vert -g ^1 $(git describe))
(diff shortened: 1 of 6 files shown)


KevinMGranger:
Hang on, recreating from my fork for the sake of force-pushing.

PR #316, opened 2021-08-20
Set default exporter source_ref to "stable", remove appVersion from chart

KevinMGranger:
This will address the core issue we saw in  #303, where pinning your repo to a specific tag still has the exporters pointed at `master`-- causing things like python code versions and container versions to get out of
sync.

Making `stable` the default will make for a better user experience, especially for updates.

I've created the [stable branch](https://github.com/konveyor/pelorus/tree/stable) which right now points to the same commit as the [v1.3.0](https://github.com/konveyor/pelorus/tree/v1.3.0) tag.

This also removes the appVersion from the chart. I don't think it makes sense to keep track of it exporter-wise, since users specify their own version with `source_ref`.

It might make sense to keep track of it for the grafana configuration? I can leave this change out if we want.

@redhat-cop/mdt

### DIFF
--- a/charts/pelorus/charts/exporters/templates/_buildconfig.yaml
+++ b/charts/pelorus/charts/exporters/templates/_buildconfig.yaml
@@ -17,7 +17,7 @@ spec:
   source:
     contextDir: {{ .source_context_dir }}
     git:
-      ref: {{ .source_ref | default "stable" }}
+      ref: {{ .source_ref | default "1.3.0.1" }}
       uri: {{ .source_url }}
     type: Git
   strategy:
--- a/charts/pelorus/charts/exporters/templates/_buildconfig.yaml
+++ b/charts/pelorus/charts/exporters/templates/_buildconfig.yaml
@@ -17,7 +17,7 @@ spec:
   source:
     contextDir: {{ .source_context_dir }}
     git:
-      ref: {{ .source_ref }}
+      ref: {{ .source_ref | default "stable" }}
       uri: {{ .source_url }}
     type: Git
   strategy:
--- a/_test/pre-commit
+++ b/_test/pre-commit
@@ -7,11 +7,6 @@ bump_version() {
     echo "Updated chart ${chart} version to ${version}."
 }
 
-set_app_version() {
-    version=$(vert -g ^1 $(git describe))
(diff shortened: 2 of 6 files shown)


etsauer:
I have one question: how does this change the behavior of future releases? Have you thought about that process as part of this?

charts/pelorus/charts/exporters/templates/_buildconfig.yaml
```diff
@@ -17,7 +17,7 @@ spec:
   source:
     contextDir: {{ .source_context_dir }}
     git:
-      ref: {{ .source_ref }}
+      ref: {{ .source_ref | default "stable" }}
```
etsauer, line 20: I fear that this change, paired with removing source_ref from variables fingers will result in different behavior than a user will be expecting, especially if they have cloned the master branch and could also contribute to a feeling of "black magic". I think defaulting to stable makes more sense once we have an install path that doesn't involve cloning a repo (e.g. using images and a published helm chart), or when installing a release. Thoughts?

KevinMGranger:
> I have one question: how does this change the behavior of future releases? Have you thought about that process as part of this?

The main idea behind this change was that there are essentially two types of user: the one who just wants to install Pelorus as-is and have it work, and those who wish to actively develop Pelorus (and those who are both).

For users who want it to Just Work™, the idea is that we'd change the installation instructions to tell people to point to `stable`, and offer a more stable experience on that branch.

For users who want to develop Pelorus, the idea is that they'd already be aware of these changes.

But you make a really good point with your [comment on the code](https://github.com/konveyor/pelorus/pull/316#discussion_r693250924). Our previous problem was having tags point to master, but now we have master pointing to stable! We've just reversed the problem.

Maybe instead, we should just release a new version so folks who are pointed at `master` for the code and a version tag for the charts won't have their setup break. Then, we can change the installation instructions to default to master-- if we're okay with committing to keeping master stable.

Thoughts on that?

KevinMGranger:
We discussed this again. We now think that optimizing for the end-user makes sense, as devs can read the developer setup instructions, telling them to change the branch to master.

KevinMGranger:
I've updated this to use a branch instead.

This means adjusting it for each release. I'm updating the developer documentation to reflect this now.

etsauer (approved):
LGTM

Issue #327, opened 2021-12-17
PipelineRun not supported by committime exporter


jlozano-rh:
With Tekton pipelines, the most cases the build step is done with **buildah** and a `Dockerfile`. With this approach no `BuildConfig` is available. The pod of the task do the build, and **committime exporter** is not able to recognize the necessary data (repository url, last hash commit, image digest, etc.).

jlozano-rh:
I writed some example code to solve this... to run this code with python3 you need the next dependencies:
* kubernetes
* openshift
* jsonpath-ng

This dependencies can be installed with pip.

This is the code to test another resources type (CRD):
```python
from kubernetes import config
from kubernetes import client
from openshift.dynamic import DynamicClient
from jsonpath_ng import parse
import urllib3

print("----- BEGIN -----")
app_label = "pelorus/app.name"

pipeline_runs = []
pipelines = []
runs_by_app = {}

print("Configuring OpenShift Client...")
urllib3.disable_warnings()
k8s_client = client.api_client.ApiClient(configuration=config.load_kube_config())
dyn_client = DynamicClient(k8s_client)

v1_builds = dyn_client.resources.get(
    api_version="tekton.dev/v1beta1", kind="PipelineRun"
)

pipeline_runs = v1_builds.get(namespace="flow-smarttv-pipes", label_selector="tekton.dev/pipeline") 

jsonpath_str = (
    "$['items'][*]['metadata']['labels']['" + "tekton.dev/pipeline" + "']"
)
jsonpath_expr = parse(jsonpath_str)

runs_found = jsonpath_expr.find(pipeline_runs)

pipelines = [match.value for match in runs_found]

pipelines = list(dict.fromkeys(pipelines))
runs_by_app = {}

print("Pipelines Found:")
print(pipelines)


for pipeline in pipelines:
    runs_by_app[pipeline] = list(
        filter(lambda p: p.metadata.labels["tekton.dev/pipeline"] == pipeline, pipeline_runs.items)
    )

for pipe in runs_by_app:
    runs = runs_by_app[pipe]
    for run in runs:
        print(run.metadata.labels["tekton.dev/pipeline"])
        params = run.status.pipelineSpec.params
        if params:
            for p in params:
                if p.name == "IMAGE_NAME":
                    print(p.default)
        task_runs = run.status.taskRuns
        if task_runs:
            for tr in task_runs:
                if tr:
                    group = tr[1].status.taskResults
                    if group:
                        for g in group:
[...]

Issue #317, opened 2021-10-21
committime exporter doesn't work with binary builds


jlozano-rh:
In Red Hat we have customers that have all it's builds configured with Binary builds, that are similar to next:
```yaml
source:
    type: Binary
strategy:
    type: Docker
    dockerStrategy:
      from:
        kind: ImageStreamTag
        namespace: openshift
        name: 'nodejs:12'
      dockerfilePath: openshift/Dockerfile
```

But the committime exporter is trying to find the next properties in builds triggered `.spec.source.git.uri` and `.spec.revision.git.commit`. These properties do not exist because the build is binary, so the committime exporter pod is recording the following log:

```log
10-21-2021 17:26:14 WARNING  Build example-project/example-build-13 in app example-app is missing required attributes to collect data. Skipping.
10-21-2021 17:26:14 DEBUG    'NoneType' object has no attribute 'git'
Traceback (most recent call last):
  File "/opt/app-root/src/committime/collector_base.py", line 185, in get_metric_from_build
    commit_sha = build.spec.revision.git.commit
AttributeError: 'NoneType' object has no attribute 'git'
10-21-2021 17:26:14 DEBUG    Searching for builds with label: pelorus/app.name in namespace: example-project
10-21-2021 17:26:14 DEBUG    Repo URL for app example-app is currently None
10-21-2021 17:26:14 DEBUG    http://gitlab.company.com/example/app.git
10-21-2021 17:26:14 DEBUG    http://gitlab.company.com/example/app.git
10-21-2021 17:26:14 WARNING  Build example-project/example-app-2 in app example-app is missing required attributes to collect data. Skipping.
10-21-2021 17:26:14 DEBUG    'NoneType' object has no attribute 'imageDigest'
```

It would be a good idea to propose that the committime exporter code implement a way of doing things for binary builds.

eformat:
Hi @KevinMGranger @jlozano-rh .. i dont know if any prior art/thinking has gone on with this one?

given that Binary Builds have none of the required information in OCP for git, i have a prototype / hack working that relies on annotating the Build from a pipeline (i'm using Tekton) but it could be any pipeline script really doing:

```
oc start-build $(params.APPLICATION_NAME) --from-archive=$(params.APPLICATION_NAME)-$(params.VERSION).tar.gz --follow --wait
LATEST_BUILD=$(oc get bc pet-battle-api --template='{{ .metadata.name }}-{{ .status.lastVersion }}')
[...]

PR #333, opened 2022-01-04
Tekton compatibility for committime exporter

### LINKED ISSUE #224: Integration with Tekton
To collect commit metrics

KevinMGranger:
copied from #327

Adds the ability to get git commit information from Tekton runs.

It's a good enough implementation for "technology preview" status. There are open questions about how to extract the right information when there's near infinite flexibility in making a pipeline.

Right now, it tries to follow the (I believe) most common usage patterns:

- Looks up git information from the `git-checkout` `ClusterTask` that comes with Tekton
- Gets the image hash from the first `TaskRun` with a `taskResult` named `IMAGE_DIGEST`
- Gets the image location from the first `TaskRun` with `IMAGE` or `IMAGE_NAME` as a param, following patterns from various s2i and buildah tasks

I'm sure we'll need to figure out a way to add some flexibility. I've started exploring some possible solutions while experimenting with this, but that should come later.

Special thanks to @jlozano-rh for starting the work on this.

## Linked Issues?

resolves #224

## Testing Instructions

1. Fork _my_ fork of this repo and check out the tekton branch.
2. Set up the committime exporter. Make sure it is pointed to your fork. The deploytime exporter must also be there but can still point upstream.
3. Run the tekton demo with `./demo/demo-tekton $URL_FOR_FORK`.
4. Make sure the pipeline was successful and look at the sample page.
5. Use the prompt to make a change, and wait for it to show up.
6. Verify the metrics show up in the dashboard.

@redhat-cop/mdt

## Remaining Work

- [ ] feature-flag the tekton features
- [ ] resolve demo issue: occasional `unable to recognize no matches for kind "Pipeline" in version "tekton.dev/v1beta1"`
- [ ] demo should poll for the page to be ready
- [ ] git-clone task unavailable for the first run?

### DIFF
--- a/charts/pelorus/charts/exporters/templates/rbac.yaml
+++ b/charts/pelorus/charts/exporters/templates/rbac.yaml
@@ -41,6 +41,16 @@ rules:
   verbs:
   - list
   - get
+- apiGroups:
+  - tekton.dev
+  resources:
+  - pipelineruns
+  - pipelines
+  - tasks
+  - taskruns
+  verbs:
+  - list
+  - get
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/demo/demo-tekton
+++ b/demo/demo-tekton
@@ -66,8 +66,8 @@ while true; do
    case $a in
       1* )
          echo "We've modified this file, time to build and deploy a new version. Times modified: $counter" | tee -a "$python_example_txt"
-         git commit -m "modifying python example, number $counter" -- "$python_example_txt"
-         git push origin "$current_branch"
+         git commit --no-verify -m "modifying python example, number $counter" -- "$python_example_txt"
+         git push "$url" "$current_branch"
 
          run_pipeline
 
--- a/exporters/committime/collector_base.py
+++ b/exporters/committime/collector_base.py
@@ -155,7 +155,10 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
                 for pipeline in pipelines
             }
 
-            metrics += self.get_metrics_from_pipelineruns(runs_by_app, namespace)
+            try:
+                metrics += self.get_metrics_from_pipelineruns(runs_by_app, namespace)
+            except Exception as e:
+                logging.error(f"Error while getting metrics from Tekton: {e}")
 
         return metrics
 
--- a/exporters/committime/collector_base.py
+++ b/exporters/committime/collector_base.py
@@ -131,13 +131,15 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
             except openshift.dynamic.exceptions.ResourceNotFoundError:
                 pipeline_runs = pelorus.NoOpResourceInstance()
 
-            # use a jsonpath expression to find all possible values for the app label
-            jsonpath_str = f"$['items'][*]['metadata']['labels']['{app_label_key}']"
-            jsonpath_expr = parse(jsonpath_str)
+            jsonpath_labels_expr = parse("$['items'][*]['metadata']['labels']")
 
-            apps: set[str] = {match.value for match in jsonpath_expr.find(builds)}
+            apps: set[str] = {
+                match.value[app_label_key]
+                for match in jsonpath_labels_expr.find(builds)
+            }
             pipeline_run_app_labels: set[str] = {
-                match.value for match in jsonpath_expr.find(pipeline_runs)
+                match.value[app_label_key]
+                for match in jsonpath_labels_expr.find(pipeline_runs)
             }
 
             builds_by_app: dict[str, list] = {
--- a/exporters/committime/collector_base.py
+++ b/exporters/committime/collector_base.py
@@ -92,20 +92,27 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
         for namespace in self._namespaces:
             # Initialized variables
             builds = []
+            pipeline_runs = []
             apps = []
+            pipelines = []
             builds_by_app = {}
+            runs_by_app = {}
             app_label = pelorus.get_app_label()
             logging.debug(
                 "Searching for builds with label: %s in namespace: %s"
                 % (app_label, namespace)
             )
-
+            # This get all Builds in our cluster
             v1_builds = self._kube_client.resources.get(
                 api_version="build.openshift.io/v1", kind="Build"
             )
+            v1_tekton = self._kube_client.resources.get(
+                api_version="tekton.dev/v1beta1", kind="PipelineRun"
+            )
             # only use builds that have the app label
             builds = v1_builds.get(namespace=namespace, label_selector=app_label)
-
+            # only use PipelineRun that have the app label
+            pipeline_runs = v1_tekton.get(namespace=namespace, label_selector=app_label)
             # use a jsonpath expression to find all values for the app label
             jsonpath_str = (
                 "$['items'][*]['metadata']['labels']['" + str(app_label) + "']"
@@ -113,21 +120,35 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
             jsonpath_expr = parse(jsonpath_str)
 
             found = jsonpath_expr.find(builds)
+            runs_found = jsonpath_expr.find(pipeline_runs)
 
             apps = [match.value for match in found]
+            pipelines = [match.value for match in runs_found]
 
-            if not apps:
+            if not apps and not pipelines:
                 continue
-            # remove duplicates
-            apps = list(dict.fromkeys(apps))
-            builds_by_app = {}
+            elif apps:
+                # remove duplicates
+                apps = list(dict.fromkeys(apps))
+                builds_by_app = {}
 
-            for app in apps:
-                builds_by_app[app] = list(
-                    filter(lambda b: b.metadata.labels[app_label] == app, builds.items)
-                )
+                for app in apps:
+                    builds_by_app[app] = list(
+                        filter(lambda b: b.metadata.labels[app_label] == app, builds.items)
+                    )
+                
+                metrics += self.get_metrics_from_apps(builds_by_app, namespace)
+            elif pipelines:
+                # remove duplicates
+                pipelines = list(dict.fromkeys(pipelines))
+                runs_by_app = {}
 
-            metrics += self.get_metrics_from_apps(builds_by_app, namespace)
+                for pipeline in pipelines:
+                    runs_by_app[pipeline] = list(
+                        filter(lambda b: b.metadata.labels[app_label] == pipeline, pipeline_runs.items)
+                    )
+                
+                metrics += self.get_metric_from_pipelineruns(runs_by_app, namespace)
 
         return metrics
 
@@ -178,11 +199,17 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
             if not repo_url:
                 if build.spec.source.git:
                     repo_url = build.spec.source.git.uri
+                elif build.metadata.labels.pelorusgithost:
+                    git_host = build.metadata.labels.pelorusgithost
+                    git_project = build.metadata.labels.pelorusgitproject
+                    git_app = build.metadata.labels.pelorusgitapp
+                    git_protocol = build.metadata.labels.pelorusgitprotocol
+                    repo_url = git_protocol + "://" + git_host + "/" + git_project + "/" + git_app
                 else:
                     repo_url = self._get_repo_from_build_config(build)
 
             metric.repo_url = repo_url
-            commit_sha = build.spec.revision.git.commit
+            commit_sha = build.metadata.labels.peloruscommitsha
             metric.build_name = build.metadata.name
             metric.build_config_name = build.metadata.labels.buildconfig
             metric.namespace = build.metadata.namespace
@@ -191,7 +218,8 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
 
             metric.commit_hash = commit_sha
             metric.name = app
-            metric.committer = build.spec.revision.git.author.name
+            #metric.committer = build.spec.revision.git.author.name
+            metric.committer = "default"
             metric.image_location = build.status.outputDockerImageReference
             metric.image_hash = build.status.output.to.imageDigest
             # Check the cache for the commit_time, if not call the API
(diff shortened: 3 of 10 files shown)


exporters/committime/collector_base.py
```diff
@@ -251,6 +279,85 @@ def get_repo_from_jenkins(self, jenkins_builds):
...
+                    # By convention need to get image location from PARAMS
+                    params = run.status.pipelineSpec.params
+                    if params:
+                        for param in params:
+                            if param.name == "IMAGE_NAME":
+                                image_location = param.default
+                    # save each taskRun inside the pipeline
+                    task_runs = run.status.taskRuns
+                    if task_runs:
+                        for taskrun in task_runs:
```
KevinMGranger, line 305: @jlozano-rh This will set repo_url and the others to the last task_run in the list. Is that intentional? I'm not really sure how it's structured for your use case.

exporters/committime/collector_base.py
```diff
@@ -92,42 +92,63 @@ def generate_metrics(self) -> Iterable[CommitMetric]:
...
-            apps = list(dict.fromkeys(apps))
-            builds_by_app = {}
-
-            for app in apps:
-                builds_by_app[app] = list(
-                    filter(lambda b: b.metadata.labels[app_label] == app, builds.items)
-                )
-
-            metrics += self.get_metrics_from_apps(builds_by_app, namespace)
+            elif apps:
```
KevinMGranger, line 130: @jlozano-rh this means that only apps _or_ pipelines will be checked, not both. Is that intentional?

exporters/committime/collector_base.py
```diff
@@ -251,6 +279,85 @@ def get_repo_from_jenkins(self, jenkins_builds):
...
+                    if params:
+                        for param in params:
+                            if param.name == "IMAGE_NAME":
+                                image_location = param.default
+                    # save each taskRun inside the pipeline
+                    task_runs = run.status.taskRuns
+                    if task_runs:
+                        for taskrun in task_runs:
+                            if taskrun:
+                                # taskResults come in list inside array format for some reason
```
KevinMGranger, line 307: @jlozano-rh Can you show me an example? [Their code](https://pkg.go.dev/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#PipelineRunStatusFields) implies it ought to be a map.
([they unfortunately don't have api reference docs yet](https://github.com/tektoncd/pipeline/issues/1250))

exporters/committime/collector_base.py
jlozano-rh: @KevinMGranger yes, it's intentional... I don't know any other way to walk through the `run` object to get those YAML properties. I mean, a `PipelineRun` has a list of executed tasks (a.k.a. `task_runs`), and in one of them we can find the **URL** and the **hash** of the last commit of the application. I hope I have answered your questions.

exporters/committime/collector_base.py
jlozano-rh: @KevinMGranger no, my bad... I made a mistake there, the idea is the following: if there are no apps or pipelines to process it does the `continue` statement and avoid processing something empty. And then the idea is to get metrics from what is actually there, I mean, if there are no pipelines to process... there is no need to get metrics from them because there are none. You know what I mean?

KevinMGranger: That makes sense!

Our linter was complaining about cyclomatic complexity. Since both `get_metrics_from_pipelineruns` and `get_metrics_from_apps` loop over what's given to them, passing them something empty won't be an issue. That means we can remove another level of nesting.

exporters/committime/collector_base.py
KevinMGranger: Wouldn't those parameters be on the `PipelineRun`'s parameters instead?

exporters/committime/collector_base.py
jlozano-rh: @KevinMGranger yes, here a snippet of YAML code. But inside the committime exporter the object has a JSON format instead:

```
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: myapp-ci-run-3bp0d5
  namespace: myapp-dev
spec:
  params:
  ...
status:
  ...
  taskRuns:
    myapp-ci-3bp0d5-build-and-push-mx4d2:
      pipelineTaskName: build-and-push
      status:
        ...
        taskResults:
          - name: IMAGE_DIGEST
            value: >-
              sha256:771f405494934dfeb46651b447f64f314fc74bcc495648f42794eb4934c890a7
        ...
    myapp-ci-3bp0d5-fetch-repository-hmnp5:
      pipelineTaskName: fetch-repository
      status:
        ...
        taskResults:
          - name: commit
            value: abdb7daef16279ff72d392be1b8ceccb5f121562
          - name: url
            value: 'https://github.com/myproject/myapp.git'
        ...
    myapp-ci-3bp0d5-get-app-version-hzfr6:
      pipelineTaskName: get-app-version
      status:
        ...
        taskResults:
          - name: value
            value: 0.1.1
        ...
    ...

```

KevinMGranger: I'm not sure what you mean by json format?

exporters/committime/collector_base.py
jlozano-rh: @KevinMGranger yes, that's right. I called `run` to each `PipelineRun` for simplicity.

exporters/committime/collector_base.py
jlozano-rh: @KevinMGranger yes, it would even look better... my code is very primitive given the urgency I had at the time with the customer.

exporters/committime/collector_base.py
jlozano-rh: @KevinMGranger in the python code, if you print() the `run` object is shown as a JSON text. But is the same as YAML that you see in the OpenShift web console

exporters/committime/collector_base.py
```diff
@@ -178,11 +198,25 @@ def get_metric_from_build(self, build, app, namespace, repo_url):
...
+                    git_protocol = build.metadata.labels.pelorusgitprotocol
+                    repo_url = (
+                        git_protocol
+                        + "://"
+                        + git_host
+                        + "/"
+                        + git_project
+                        + "/"
+                        + git_app
+                    )
```
etsauer, line 214: I don't think we should break up git URLs in this way. There are a myriad of [supported git url formats](https://stackoverflow.com/questions/31801271/what-are-the-supported-git-url-formats) that are not covered by this structure, and to try to cover them all would be cumbersome and unnecessary. I would just stick to the two standard git parameters that are used all over the place:

- `git_url` - The full URL string in any format
- `git_ref` - The branch or tag to be examined, if applicable

exporters/committime/collector_base.py
```diff
@@ -191,7 +225,8 @@ def get_metric_from_build(self, build, app, namespace, repo_url):
 
             metric.commit_hash = commit_sha
             metric.name = app
-            metric.committer = build.spec.revision.git.author.name
+            # metric.committer = build.spec.revision.git.author.name
+            metric.committer = "default"
```
etsauer, line 229: why not label for this one?

exporters/committime/collector_base.py
KevinMGranger: If it works, it works! Just wanted to make sure I wasn't about to make the mistake of [Chesterton's Fence](https://fs.blog/chestertons-fence/).

exporters/committime/collector_base.py
jlozano-rh: @etsauer I initially approached the solution using labels. But labels have character restrictions for the value part. The idea is to improve this with an annotation to avoid dirty code like this one.

exporters/committime/collector_base.py
jlozano-rh: @etsauer for some reason, in my `build`  object the property `build.spec.revision.git.author.name` doesn't exist... so that the committime exporter does not fail, I set it to "default" value.

exporters/committime/collector_base.py
KevinMGranger: Using annotations makes sense. I'll add a `BUILD_GIT_REPO_URI_ANNOTATION` env variable (or is that too long...)

We should probably use a namespaced default, right? Something like `pelorus.konveyor.io/git-uri`? Or keep it simple with `gitUri`?

exporters/committime/collector_base.py
KevinMGranger: It might only be there if it was kicked off from a `BuildConfig`. I'll make it try but then fall back.

exporters/committime/collector_base.py
jlozano-rh: @KevinMGranger I was updating these values with a pipeline dynamically. Regarding the env variable, I think it is fine, I prefer it to be long and understandable rather than short and confusing.

Regarding the namespaced default, I think it's the same thing... keeping it simple can lead to confusion, I think having pelorus word in the name would indicate what it is.

exporters/committime/collector_base.py
KevinMGranger: I looked into it-- I think it's the weirdness of `openshift.dynamic.ResourceField`. I'll do some testing to verify that.

jlozano-rh: @KevinMGranger exactly, at the time I investigated and found no other way to do it.

demo/demo-tekton
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
+#Assumes User is logged in to cluster
+set -euo pipefail
+
+all_cmds_found=0
+for cmd in oc tkn; do
+   if ! which -s $cmd; then
```
mpryc, line 7: I am in favor of using 'type' then which. Also -s is not available in all 'which' versions.

demo/demo-tekton
```diff
@@ -0,0 +1,83 @@
...
+      echo "No $cmd executable found in PATH" >&2
```
mpryc, line 8: Maybe print what is the $PATH so we know where the command was searched for?

demo/demo-tekton
```diff
@@ -0,0 +1,83 @@
...
+      all_cmds_found=1
+   fi
+done
+if ! [[ $all_cmds_found ]]; then exit 1; fi
+
+
+tekton_setup_dir="$(dirname $0)/tekton-demo-setup"
```
mpryc, line 15: If we are sure it's bash, then it's better to use:
$(dirname "$BASH_SOURCE")

In other case use:
SCRIPT_PATH=$(readlink "$0")
$(dirname "$SCRIPT_PATH")

demo/python-example/example.py
```diff
@@ -0,0 +1,33 @@
...
+response_file_bytes = b""
+response_file_txt = ""
+
+
+class ResponseHandler(BaseHTTPRequestHandler):
+    def do_GET(self):
+        self.log_message(
+            f"Got request, returning {response_file_txt} (as bytes {response_file_bytes})"
+        )
+        self.send_response(200, "PELORUS IS COOL")
```
mpryc, line 12: Indeed :)

mpryc (commented):
Just few nits around shell to get started and understand what's happening in the project :)

exporters/committime/collector_base.py
```diff
@@ -90,54 +103,73 @@ def generate_metrics(self) -> Iterable[CommitMetric]:
...
+                )
+                # only use PipelineRun that have an app label
+                pipeline_runs = v1_tekton.get(
+                    namespace=namespace, label_selector=app_label
+                )
+            except openshift.dynamic.exceptions.ResourceNotFoundError:
+                pipeline_runs = pelorus.NoOpResourceInstance()
+
+            # use a jsonpath expression to find all possible values for the app label
+            jsonpath_str = f"$['items'][*]['metadata']['labels']['{app_label}']"
```
KevinMGranger, line 130: This will break if app_label has single quotes in it for some reason. We could extract this part with plain python.

exporters/committime/collector_base.py
```diff
@@ -251,6 +306,78 @@ def get_repo_from_jenkins(self, jenkins_builds):
                 )
         # If no repo is found, we will return None, which will be handled later on
 
+    def get_metrics_from_pipelineruns(
```
mpryc, line 309: Should this be a "private" method ? _get_metrics_from_pipelineruns()

mpryc, line 309: Any description of what the method does is welcome :)

exporters/committime/collector_base.py
```diff
@@ -273,3 +400,110 @@ def _get_repo_from_build_config(self, build):
...
+    def _find_git_clone_task_name(
+        self, pipeline_name: PipelineName, namespace: str
+    ) -> Optional[str]:
+        """
+        Get the Pipeline associated with a given PipelineRun,
+        and find the task name for the git-clone ClusterTask, if one exists.
+        Used to find the TaskRun that did the git checkout.
+        """
+        pipeline_resource = self._kube_client.resources.get(
+            api_version="tekton.dev/v1beta1", kind="Pipeline"
```
mpryc, line 413: As in previous comment. I do think such api_versions should be hidden in the function body.

exporters/pelorus/utils.py
```diff
@@ -0,0 +1,181 @@
...
+            if type(self) is type(other):
+                return self.data == other.data
+            else:
+                return False
+        elif type(other) is str:
+            return self.data == other
+        else:
+            return NotImplemented
+
+    def __repr__(self):
```
mpryc, line 176: Should you define `__str__(self) `as well?

mpryc (commented):
1. I think we should allow to install tkn and oc commands using Makefile dev-env to a separate PATH. This can however be different PR.

2. Should the project template be compatible as well with kubectl ? It means to use in the demo/tekton-demo-setup/02-project.yaml api version v1, similarly to:
   https://github.com/redhat-cop/openshift-applier/blob/main/tests/files/templates/projectrequest.yml

exporters/committime/collector_base.py
```diff
@@ -19,6 +23,13 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
...
     """
 
+    _kube_client: openshift.dynamic.DynamicClient
+    _git_uri_annotation: str
+    _commit_dict: dict[str, float]
+    """The name of the annotation to check if no repo_url comes from jenkins and
+    the build spec doesn't have git information.
+    """
+
     def __init__(
```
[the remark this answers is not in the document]
KevinMGranger: I'd like to audit and refactor every class we use, but that's out of scope for this PR.

charts/pelorus/charts/exporters/templates/rbac.yaml
```diff
@@ -41,6 +41,16 @@ rules:
   verbs:
   - list
   - get
+- apiGroups:
+  - tekton.dev
+  resources:
```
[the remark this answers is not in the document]
KevinMGranger: I only see pipelineruns and taskruns in that list.

As for the other resources, we should only ask for the permissions we truly need.

exporters/committime/collector_base.py
```diff
@@ -59,7 +72,7 @@ def collect(self):
                     my_metric.name,
                     my_metric.commit_hash,
                     my_metric.image_hash,
-                    str(float(my_metric.commit_timestamp)),
+                    str(float(my_metric.commit_timestamp)),  # type: ignore
```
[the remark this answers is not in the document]
KevinMGranger: Not my code, just making the type checker behave.

exporters/committime/collector_base.py
```diff
@@ -90,54 +103,73 @@ def generate_metrics(self) -> Iterable[CommitMetric]:
...
+            # only use builds that have an app label
+            builds = v1_builds.get(namespace=namespace, label_selector=app_label_key)
 
-            # use a jsonpath expression to find all values for the app label
-            jsonpath_str = (
-                "$['items'][*]['metadata']['labels']['" + str(app_label) + "']"
-            )
+            try:
+                v1_tekton = self._kube_client.resources.get(
+                    api_version="tekton.dev/v1beta1", kind="PipelineRun"
```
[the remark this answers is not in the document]
KevinMGranger: I'm not really sure what you mean. If the api version changes, then the interface may change, and we'll need to account for that. I don't want to abstract this too early until we know what will change.

exporters/committime/collector_base.py
```diff
@@ -160,38 +192,58 @@ def get_metrics_from_apps(self, apps, namespace):
             for build in code_builds:
                 try:
                     metric = self.get_metric_from_build(build, app, namespace, repo_url)
+
+                    if not metric:
+                        continue
+
+                    logging.debug("Adding metric for app %s" % app)
```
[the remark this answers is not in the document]
KevinMGranger: Not my code, just moving it up from below.

I really do want us to audit each log line and ask ourselves if we're getting the info we want, or if it's necessary.

exporters/committime/collector_base.py
```diff
@@ -160,38 +192,58 @@ def get_metrics_from_apps(self, apps, namespace):
...
-            metric.build_name = build.metadata.name
-            metric.build_config_name = build.metadata.labels.buildconfig
-            metric.namespace = build.metadata.namespace
-            labels = build.metadata.labels
+
+            if spec.revision:
+                commit_sha = spec.revision.git.commit
+                metric.committer = spec.revision.git.author.name
+            else:
+                # TODO: check for existence, allow user customization? Annotation instead?
```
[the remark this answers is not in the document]
KevinMGranger: Good call. I missed the `UNKNOWN_AUTHOR` part as well-- I think we can just leave it out, but I need to double check Grafana/Prometheus.

exporters/committime/collector_base.py
KevinMGranger: I guess I got too excited about changing from WIP and forgot to comment 😆  I'll document all the methods I added.

KevinMGranger: Good call!

exporters/committime/collector_base.py
```diff
@@ -273,3 +400,110 @@ def _get_repo_from_build_config(self, build):
...
+            return None
+
+        clone_task_results = name_value_attrs_to_dict(
+            git_clone_task_run.status.taskResults
+        )
+        return (clone_task_results["url"], clone_task_results["commit"])
+
+    def _find_first_image_param(self, pipeline_run) -> Optional[str]:
+        """
+        Find the first image value from the params of any TaskRun in this PipelineRun.
```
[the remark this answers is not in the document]
KevinMGranger: Right now, I don't have a good way of letting the user specify which one matters. This one is Good Enough:tm:, assuming that most pipelines will have a single image build.

I wonder if we really need the image information at all? We track a lot of data in prometheus but don't use it in grafana.

exporters/pelorus/utils.py
KevinMGranger: I need to double check how these are used within the openshift library. I think I was unsure if `__str__` should transparently use the underlying string, or use `__repr__`.

KevinMGranger (commented):
Addressed your comments, looks like I have a little more work to do.

KevinMGranger:
To summarize the remaining work:

- [ ] Move resource versions / names up into constants
- [ ] double-check how typedstrings would be interpreted within the openshift library
- [ ] handle string interpolation in the jsonpath expression

KevinMGranger:
Because tekton pipelines could be customized to do anything, supporting this by _guessing_ isn't worth the technical cost.

We've decided that the best course of action is to annotate the resulting build with the required info, which was implemented with PR https://github.com/konveyor/pelorus/pull/498. It also solves the problem in issue https://github.com/konveyor/pelorus/issues/317 at the same time!

What would make tekton "support" easier is documenting how you might label the resulting build, and including an example of that. We already have demo code in the repo, so we'll likely use that.

PR #344, opened 2022-01-26
Fix default exporter ref pointing to old release

KevinMGranger:
The v.1.4.0 release didn't have an updated default tag.

I've fixed that, and another step in the release process that was missing.


etsauer (approved):
LGTM

PR #346, opened 2022-01-31
Issue #337 - deploytime exporter ignores PROD_LABEL env config

### LINKED ISSUE #337: deploytime exporter ignores PROD_LABEL env config
# context

DeployTime exporter is doced as having PROD_LABEL config option to filter what counts as a production deployment (https://github.com/konveyor/pelorus/blob/master/docs/Configuration.md#deploy-time-exporter) but it is not fully implemented.

The get_prod_label (https://github.com/konveyor/pelorus/blob/master/exporters/pelorus/__init__.py#L73-L74) function exists but it is not invoked by any code (other then a test).

# resoltuion

Update the DeployTime exporter to use the NAMESPACE and PROD_LABEL env variables (or either or depending on what the intended functionality is if both are given): https://github.com/konveyor/pelorus/blob/master/exporters/deploytime/app.py#L171-L177

itewk:

fixes broken functionality for PROD_LABEL for DeployTime exporter.

## Linked Issues?

resolves #337

## Testing Instructions

i tested this in my own environment. If you want me to show that to someone or some sort of unit testing I should be doing let me know.

@redhat-cop/mdt


itewk:
i fixed the formatting issue. i dont know what to do about the "to complex" pylava issue. you want me to factor all the logic for getting namespaces with the prod label into a function? if so, where?

KevinMGranger:
I'll take a look and see re: the complexity. I can usually make the linter happy.

itewk:
> I'll take a look and see re: the complexity. I can usually make the linter happy.

much appreciated.

KevinMGranger:
@itewk I kept the behavior or defaulting to all namespaces if neither `NAMESPACES` nor `PROD_LABEL` is given. That wasn't how it worked in 1.2.*, but it was introduced at some point, so I don't want to just remove it. That idea is inherited (perhaps accidentally) right from the kube API, where `?labelSelector=` will match all.

Can you give this a shot in your environment and see if it's satisfactory?

itewk:
> @itewk I kept the behavior or defaulting to all namespaces if neither `NAMESPACES` nor `PROD_LABEL` is given. That wasn't how it worked in 1.2.*, but it was introduced at some point, so I don't want to just remove it. That idea is inherited (perhaps accidentally) right from the kube API, where `?labelSelector=` will match all.
> 
> Can you give this a shot in your environment and see if it's satisfactory?

The problem with the logic you just put in is if someone supplies a PROD_LABEL like “foo=doesnotexist” which nonnamespaces are labeled as then you get back all namespaces. If user supplies a PROD_LABEL with no labeled namespaces then the watched list should be no namespaces not all namespaces.

KevinMGranger:
> If user supplies a PROD_LABEL with no labeled namespaces then the watched list should be no namespaces not all namespaces.

I can't recall if that was the original behavior, but that shouldn't happen in the code as-is. A non-existent PROD_LABEL will return no results, which will watch _no_ namespaces.

If the PROD_LABEL is the empty string (the default), then it will watch all of them.

exporters/deploytime/app.py
```diff
@@ -50,6 +55,34 @@ def collect(self) -> Iterable[GaugeMetricFamily]:
...
+            return self._namespaces
+
+        logging.info(
+            "No namespaces specified, watching all namespaces with given PROD_LABEL (%s)",
+            self.prod_label,
+        )
+        all_namespaces = dyn_client.resources.get(api_version="v1", kind="Namespace")
+        namespaces = {
+            namespace.metadata.name
+            for namespace in all_namespaces.get(label_selector=prod_label).items
```
itewk, line 77: @KevinMGranger are you saying that this will return all namespaces if `prod_label` is `null` or empty string? Even if that is the case, it seems like a very sideways way of defaulting to all namespaces rather then being explicite about it with an if statement that if no PROD_LABEL is supplied then use all namespaces. not to mention the confusion with th elogging mesage of `No namespaces specfiied, watching all namespaces with given PROD_LABEL (%s)` where PROD_LABEL is empty.

KevinMGranger: You're right that the log message is confusing, I'll change that. But is this sideways? It explicitly retrieves all of them so that they can be logged.

KevinMGranger: Ah, I see what you mean-- the behavior of querying for an empty string is confusing. I've addressed that in the latest commit, if you think that helps.

exporters/deploytime/app.py
```diff
@@ -50,6 +55,42 @@ def collect(self) -> Iterable[GaugeMetricFamily]:
             )
             yield (metric)
 
+    def get_and_log_namespaces(self) -> set[str]:
+        """
+        Get the set of namespaces to watch, and log what they are.
+        They will be either:
+        1. The namespaces explicitly specified
+        2. The namespaces matched by PROD_LABEL
+        3. If the PROD_LABEL is given, that implicitly matches all namespaces.
```
itewk, line 64: this comment does not make ssnese to me. Shoudln't it be "if enither NAMESPACE or PROD_LABEL given, then match all namespaces". it seems in conflict with bullet 2 right now.

KevinMGranger: Whoops, I meant to say "is not given".

itewk:
@KevinMGranger i tested the latest branch against my environment and it seems to be working.

KevinMGranger:
Awesome! I'll fix that comment in a bit and then we should be good to go. Or if you have a spare second, you can fix it.

itewk:
> Awesome! I'll fix that comment in a bit and then we should be good to go. Or if you have a spare second, you can fix it.

pushed commit to fix comment.

KevinMGranger (approved):

PR #352, opened 2022-02-08
Make no-auth github exporter okay

KevinMGranger:
Public repos do not need authentication.

This makes local testing and demoing the committime exporter much easier.

### DIFF
--- a/exporters/committime/app.py
+++ b/exporters/committime/app.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python3
+import logging
 import os
-import sys
 import time
 from distutils.util import strtobool
 
@@ -15,8 +15,6 @@ from prometheus_client import start_http_server
 from prometheus_client.core import REGISTRY
 
 import pelorus
-
-REQUIRED_CONFIG = ["GIT_USER", "GIT_TOKEN"]
 
 
 class GitFactory:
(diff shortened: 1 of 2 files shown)


exporters/committime/app.py
```diff
@@ -46,17 +44,18 @@ def getCollector(
...
     pelorus.load_kube_config()
     k8s_config = client.Configuration()
     k8s_client = client.api_client.ApiClient(configuration=k8s_config)
     dyn_client = DynamicClient(k8s_client)
 
-    username = os.environ.get("GIT_USER")
-    token = os.environ.get("GIT_TOKEN")
+    username = os.environ.get("GIT_USER", "")
+    token = os.environ.get("GIT_TOKEN", "")
+    if not username and not token:
```
mpryc, line 55: How about:
If no GIT_USER or GIT_TOKEN is provided:
   check the git connection, if valid continue with information (meaning it's public repository), otherwise fail saying it could not authenticate and ask to provide valid GIT_USER and GIT_TOKEN?

KevinMGranger: That would have to be done per-repo. I believe authentication errors will already be logged, but I'll have to double check. We could also make a test for that, but that could be done later.

For now, I think logging about it is a good middle ground. Maybe it should be a `warning` instead?

mpryc: @KevinMGranger is currently only one GIT_USER and GIT_TOKEN associated with the exporter for all the repos? If yes then I think it's an issue as USER/TOKEN should be associated with the repo and not all of them. In that case the USER/TOKEN check would be valid as it's 1-1 mapping.

Ideally if the association is one repo - zero or more tokens (even multiple as one git org may have multiple tokens enabled for various periods/users/use-cases).

KevinMGranger: Resolving as per our discussion earlier.

mpryc (approved):

PR #354, opened 2022-02-09
Add python tekton pipeline demo

### LINKED ISSUE #331: Make a simple tekton-based example demo
[basic-spring-boot-tekton](https://github.com/redhat-cop/container-pipelines/tree/ad57251d7d5faf810076be95411b231d4551904e/basic-spring-boot-tekton#readme)

Much like the [existing demo script](https://github.com/konveyor/pelorus/blob/3e91ef811026e205e64717a85fdb1680cafd8559/demo/demo.sh) uses the [basic-nginx-pipeline](https://github.com/redhat-cop/container-pipelines/tree/ad57251d7d5faf810076be95411b231d4551904e/basic-nginx), this will allow us to perform rudimentary testing.

The part that may cause issues is whatever complexity comes from the spring-boot setup-- perhaps an nginx version would be simpler and/or build faster.

KevinMGranger:
This adds a basic python-based tekton demo in the `demo` dir, and a script to set it up and run it.

This is separate from the work in #333 because we should merge early and merge often. #333 will be a big enough PR on its own, so separating this will help with readability.

This is _not_ documented in the `docs/Demo.md` page because we should have the tekton functionality in pelorus up and running first.

Closes #331.

## Testing Instructions

1. Make sure:
  - `oc` and `tkn` are installed
  - you're logged in to the cluster
  - you have permission to install operators and create projects
2. run `./demo/demo-tekton

You should eventually see a simple page served at the link referenced in the output.

Continuing with the script as normal should eventually show changes to that page.

@redhat-cop/mdt

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -1,0 +1,13 @@
+# demo
+
+Scripts to demonstrate pelorus functionality live here.
+
+## demo-tekton
+
+Although the pelorus functionality isn't merged yet, you can set up a basic python tekton pipeline with this script.
+
+It takes a single argument which is the downstream/fork URL you'll be pushing to.
+
+## demo.sh
+
+See the [demo docs in the official pelorus documentation.](https://pelorus.readthedocs.io/en/latest/Demo/)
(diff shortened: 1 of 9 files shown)


mpryc (approved):
LGTM

PR #357, opened 2022-02-10
Small setuptools improvements

mpryc:
With the https://github.com/pypa/pip/issues/7555 the pip install .
is going to be removed with the pip 21.3. We should use instead:
$ pip install --use-feature=in-tree-build exporters/

This change also adds wheel so the python package versionig for
exporters in the python packaging is also included.

## Linked Issues?

## Testing Instructions

@redhat-cop/mdt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,10 @@ share/python-wheels/
 *.egg
 MANIFEST
 
+# We don't publish pip generated files
+exporters/AUTHORS
+exporters/ChangeLog
+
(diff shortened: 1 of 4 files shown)


mpryc:
This is currently draft as we should get other PRs merged and update README /Makefile to incorporate the in-tree-build option, The simple pip install . will not work with this change.

mpryc:
@KevinMGranger with the PR #370 I think this one isn't valid anymore?

PR #366, opened 2022-02-11
Pre-release version 1.4.6

mpryc:
We bump the version by few minor releases ty sync exporter versions
and pelorus Chart.

## Linked Issues?

## Testing Instructions

@redhat-cop/mdt


KevinMGranger (changes requested):
As I mentioned in chat, chart versions don't need to be (and shouldn't be) in lockstep with the exporter version. This is because chart updates would bump the version, and then immediately be out of sync with the exporter code.

There is a larger question: what does it mean to have a tag in this repo, since we have both exporter code and the charts here?

PR #368, opened 2022-02-14
Optional shellcheck and chart lint

KevinMGranger:
I made a typo in #353, making a bad shebang in ./scripts/setup-pre-commit-hook.

Shellcheck would have caught this.

While trying to figure out how to make shellcheck optionally run, I realized the pattern would work for chart-lint as well, so I added that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,11 @@ ifeq ($(PYTHON_VER), False)
                            and <= $(PYTHON_VER_MAX))
 endif
 
+CHART_TEST=$(shell which ct)
+
+SHELLCHECK=$(shell which shellcheck)
+SHELL_SCRIPTS=./scripts/pre-commit ./scripts/setup-pre-commit-hook
+
 
 .PHONY: default
 default: \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -93,18 +98,42 @@ isort-check: $(PELORUS_VENV)
...
+
 chart-lint: $(PELORUS_VENV)
 	. ${PELORUS_VENV}/bin/activate && \
 	./scripts/chart-lint
 
+ifneq (, $(CHART_TEST))
+chart-lint-optional: chart-lint
+else
+chart-lint-optional:
+	@echo "chart test (ct) not installed, skipping"
```
mpryc, line 123: should this be warn ?

KevinMGranger: Yes it should be!

mpryc (commented):
When I am invoking the $ make lint or $ make shellcheck goes into infinite loop.

Makefile
```diff
@@ -93,18 +98,42 @@ isort-check: $(PELORUS_VENV)
 
 # Linting
 
-.PHONY: lint pylava chart-lint
-lint: pylava # TODO: not using chart-lint until we can automate installing it or at least conditionally run it
+.PHONY: lint pylava chart-lint chart-lint-optional shellcheck shellcheck-optional
+lint: pylava shellcheck chart-lint-optional shellcheck-optional
```
mpryc, line 102: The issue with infinite loop is that "make lint" calls shellcheck which calls ./scripts/pre-commit that calls "make -k lint" back.

mpryc (approved):

Issue #358, opened 2022-02-10
Create image per pelorus exporter


mpryc:
Currently pelorus exporter images are the same.

The path_context is exactly the same for 3 different exporters resulting in the images being exactly the same.

[https://github.com/konveyor/pelorus/blob/master/.github/workflows/build-pelorus.yaml#L32](https://github.com/konveyor/pelorus/blob/master/.github/workflows/build-pelorus.yaml#L32
)

There should be separation on how exporters are being built/tested/provided and not being all in one.

Currently the exporters are available at:
[https://quay.io/repository/pelorus/failure-exporter](https://quay.io/repository/pelorus/failure-exporter
)
[https://quay.io/repository/pelorus/committime-exporter](https://quay.io/repository/pelorus/committime-exporter
)
[https://quay.io/repository/pelorus/deploytime-exporter](https://quay.io/repository/pelorus/deploytime-exporter)

If we inspect each of the container it's exactly the same, e.g.

```
# podman run -ti --name  deploytime-exporter --hostname deploytime --network host d9509bae4bc7 /bin/bash
(app-root) ls
committime  deploytime	failure  pelorus  README.md  requirements-dev.txt  requirements.txt  setup.py  tests
(app-root) pwd
/opt/app-root/src

```

KevinMGranger:
Related: #276

mpryc:
This is closed, however comment is desired.

We do create quay.io images and tag them correctly during release process.

This is done via github actions:

Build action which builds and tags the image with it's sha and "latest" tag on every PR merged:
    https://github.com/konveyor/pelorus/blob/master/.github/workflows/build-pelorus.yaml

Release action to tag the images with "stable" tag on release event, from the latest image sha:
    https://github.com/konveyor/pelorus/blob/master/.github/workflows/release.yaml

Pre-release action which tags the images on release and pre-release with it's version from the latest image sha:
    https://github.com/konveyor/pelorus/blob/master/.github/workflows/pre-release.yaml

The resulting images:
    https://quay.io/repository/pelorus/committime-exporter?tab=tags
    https://quay.io/repository/pelorus/failure-exporter?tab=tags
    https://quay.io/repository/pelorus/deploytime-exporter?tab=tags


There is still room for improvement as s2i have same code stored in each image, the difference is app.py being invoked, so we could get rid of the code from other exporters then the one for which image is built.

PR #370, opened 2022-02-17
Build image per pelorus exporter

mpryc:
Fix for issue #358.

Currently pelorus builds exporters as one, meaning each exporter is exactly the same,
just the name of the image differs.

This improves how we are building and publishing images to image registry,
ensuring each exporter contains only requires set of python files.

Futher work will involve ensuring we are installing only those
requirements that are needed per exporter.

Symbolic links in each exporter are created to ensure backwards compability
with the current helm charts and it's data from values.yaml.

Github action to build image was updated to use v2 version of s2i image and
some variables were introduced to easy testing in a separate quay.io namespace.

## Linked Issues?

## Testing Instructions

@redhat-cop/mdt


etsauer:
@mpryc I wouldn't worry about introducing non-backwards compatible changes. We don't allow/support/recommend/document intermixing older/newer versions of the exporters together so compatibility-wise it won't be an issue.

KevinMGranger:
A couple thoughts:

- Is this still a WIP? If so, we should mark it as a draft.
- We said we wanted to use the makefiles as the standard way of setting up the environment-- do we want the github actions to use `make`? Then we're not updating things in multiple spots each time.
- Is there a reason to have the additional layer of `pelorus_common`? Can we put `requirements.txt` and `setup.*` in `exporters/pelorus`?
    - Or maybe symlink to `pelorus` instead of `pelorus_common`?
- Will a symlink out of the context dir work in the first place? If we want to make the image only include what's necessary, then we still need to keep the other directories when we do this. Maybe we should delete the other directories at build time?

mpryc:
@KevinMGranger 

> * Is this still a WIP? If so, we should mark it as a draft.
I had this as draft when was testing locally, but now it's not. We need to make decision if this is the right approach or drop it.

> * We said we wanted to use the makefiles as the standard way of setting up the environment-- do we want the github actions to use `make`? Then we're not updating things in multiple spots each time.
Definitely, but not as part of this review. Filled issue #374 for that. 

> * Is there a reason to have the additional layer of `pelorus_common`? Can we put `requirements.txt` and `setup.*` in `exporters/pelorus`?
I remember why I did that. In short isort was unhappy with the order in which the imports were pulled and there was difference on the local venv system versus github actions. I fixed it in the last commit by adding .isort.cfg and explicitly stating pelorus is known_first_party package. Note that this caused isort to reorder a bit the imports in various files. But now both venv and github is giving same results.

>   * Or maybe symlink to `pelorus` instead of `pelorus_common`?
Yes I like symlink better. 
In the latest revision of this PR I have removed requirements.txt from pelorus_common all together. Each exporter is using pelorus_common not as real python library but more of a folder with just one common python script and then it should list it's dependencies in the requirements.txt  

> * Will a symlink out of the context dir work in the first place? If we want to make the image only include what's necessary, then we still need to keep the other directories when we do this. Maybe we should delete the other directories at build time?

Not really. Source 2 image makes it easier as symlink is being created on the resulting image as a standard folder with content in it (content from what was in the symlink).

I have updated latest revision as there was also file conflict from your last merge.

mpryc:
I am closing this PR. Currently we are building images, not fully optimized, however if needed this will be implemented different way. Current codebase have lot's of newer changes and many of the conflicting files, so it's easier to create fresh change, but it's not priority atm.

Issue #321, opened 2021-12-10
Unable to install pelorus on openshift version 4.9.9


metaversed:
Steps followed

    git clone --depth 1 --branch v1.3.0 https://github.com/konveyor/pelorus
    cd pelorus
    oc create namespace pelorus
    helm install operators charts/operators --namespace pelorus

After above step is performed , received below error on terminal
```
elorus git:(cfff5f2) helm install operators charts/operators/ --namespace pelorus --debug                                     
install.go:178: [debug] Original chart version: ""
install.go:199: [debug] CHART PATH: /Users/raghurambanda/repos/pelorus/charts/operators

client.go:128: [debug] creating 6 resource(s)
wait.go:48: [debug] beginning wait for 6 resources with timeout of 5m0s
client.go:128: [debug] creating 1 resource(s)
client.go:528: [debug] Watching for changes to Job installplan-approver with timeout of 5m0s
client.go:556: [debug] Add/Modify event for installplan-approver: ADDED
client.go:595: [debug] installplan-approver: Jobs active: 0, jobs failed: 0, jobs succeeded: 0
client.go:556: [debug] Add/Modify event for installplan-approver: MODIFIED
client.go:595: [debug] installplan-approver: Jobs active: 1, jobs failed: 0, jobs succeeded: 0
Error: INSTALLATION FAILED: failed post-install: timed out waiting for the condition
helm.go:88: [debug] failed post-install: timed out waiting for the condition
INSTALLATION FAILED
main.newInstallCmd.func2
	helm.sh/helm/v3/cmd/helm/install.go:127
github.com/spf13/cobra.(*Command).execute
	github.com/spf13/cobra@v1.2.1/command.go:856
github.com/spf13/cobra.(*Command).ExecuteC
	github.com/spf13/cobra@v1.2.1/command.go:974
github.com/spf13/cobra.(*Command).Execute
	github.com/spf13/cobra@v1.2.1/command.go:902
main.main
	helm.sh/helm/v3/cmd/helm/helm.go:87
runtime.main
	runtime/proc.go:255
runtime.goexit
```

    # Verify the operators are completely installed before installing the pelorus helm chart
`    helm install pelorus charts/pelorus --namespace pelorus`

```
➜  pelorus git:(cfff5f2) oc get sub -n pelorus
NAME               PACKAGE            SOURCE 
               CHANNEL
grafana-operator   grafana-operator   community-operators   alpha
prometheus         prometheus         community-operators   beta
```

[...]

PR #372, opened 2022-02-22
Adapt Pelorus charts to deploy in Openshift 4.9+ clusters

flacatus:
The current PR tries to deploy pelorus on Openshift 4.9 with the following changes:
- Change subscriptions of grafana to v4 channel and update the grafana versions
- Increase Prometheus version to not use v1alpha CRDS(not compatible with 4.8 or less openshift versions)
- Remove the spec.name value from grafana CR specs, due it is not compatible with latest grafana operators
- 
## Linked Issues?

related to #321  

## Testing Instructions

@redhat-cop/mdt


weshayutin:
nice.. will give this a go :)

mpryc (approved):
Tested on the following versions:
OCP 4.9.9
OCP 4.8.2

Helm install/uninstall works great.

As in the offline discussion with the author of this PR it will cause pelorus incompatible with EOL version of OCP (4.6). The following change is incompatible:
```
apiVersion: batch/v1
kind: CronJob
```

mpryc:
@flacatus one more version bump in the file:
 charts/operators/Chart.yaml
 
 I think you can sync it with pelorus as we will have it in the next release tag.
@KevinMGranger do you agree?

weshayutin:
on 4.7 grafana is hitting an issue.

grafana-operator-controller-manager:
(combined from similar events): Error: container create failed: time="2022-02-28T20:32:00Z" level=error msg="container_linux.go:366: starting container process caused: chdir to cwd (\"/home/nonroot\") set in config.json failed: permission denied"

weshayutin:
> 

FYI... my cluster was at 4.7.0, I'm updating the cluster to 4.7.43 and will retry :)

mpryc:
> > 
> 
> FYI... my cluster was at 4.7.0, I'm updating the cluster to 4.7.43 and will retry :)

I can confirm this issue on 4.7.0.

Sounds like similar issue to https://bugzilla.redhat.com/show_bug.cgi?id=1934177

mpryc:
@weshayutin @flacatus The grafana oerator is failing on 4.7.0, but I believe it's a mentioned problem with runc. Here is really nice post about the overall issue: https://danielmangum.com/posts/runc-chdir-to-cwd/

Anyway I have tested this on 4.7.14 and 4.8.12 as well on the 4.9.9 and everything works ok. So I am pretty confident this is the right approach and for the releases 4.7.0 the path is to update OCP as it's not the issue with pelorus itself.

mpryc (approved):

Issue #375, opened 2022-02-25
Enhance dev-env to include needed tools such as oc or tkn


mpryc:

KevinMGranger:
Handled by #376

PR #376, opened 2022-02-25
Enhance dev-env to include needed tools such as oc or tkn #375

mpryc:
New script which helps to setup dev-env and include required
non python CLI tools in the virtual environment.

Simply make dev-env && source .venv/bin/activate

Script can be also invoked outside of Makefile pointing to
different VENV.

Currently Linux and Mac OS X are working.

## Linked Issues?

## Testing Instructions

@redhat-cop/mdt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,12 @@ git-blame:
 .git/hooks/pre-commit: scripts/pre-commit
 	./scripts/setup-pre-commit-hook
 
-dev-env: $(PELORUS_VENV) exporters git-blame .git/hooks/pre-commit
+.PHONY: cli_dev_tools
+cli_dev_tools:
+	./scripts/install_dev_tools -v $(PELORUS_VENV)
+
+dev-env: $(PELORUS_VENV) cli_dev_tools exporters git-blame \
+         .git/hooks/pre-commit
 	$(info **** To run VENV: $$source ${PELORUS_VENV}/bin/activate)
 	$(info **** To later deactivate VENV: $$deactivate)
 
--- a/scripts/install_dev_tools
+++ b/scripts/install_dev_tools
@@ -1,0 +1,214 @@
+#!/usr/bin/env bash
+#
+# Copyright Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+#
+
+# Get the full absolute path to the script. Needed while calling the script
+# via various partial paths or sourcing the file from shell
+# BASH_SOURCE[0] is safer then $0 when sourcing.
+# We enter the dirname of the invoked script and then get the current
+# path of the script using pwd
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+# Match the .venv created by the Makefile
+DEFAULT_VENV="${SCRIPT_DIR}/../.venv"
+
+TMP_DIR_PREFIX="pelorus_tmp_"
+
+TKN_CLIENT_API_URL="https://api.github.com/repos/tektoncd/cli/releases/latest"
+CT_CLIENT_API_URL="https://api.github.com/repos/helm/chart-testing/releases/latest"
+
+HELM_INSTALL_SCRIPT="https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3"
+
+# Currently we are interested in development on Linux and Mac OS platforms
+# Because OCP_CLIENT URL links do not follow standard binary name format
+# e.g. Linux/Darwin we need to map those 
+PLATFORM="linux"
+[[ "$(uname -s)" =~ "Darwin" ]] && PLATFORM="mac"
+OCP_CLIENT_URL="https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/openshift-client-${PLATFORM}.tar.gz"
+
+
+function download_file_from_url() {
+    local url=$1
+    local dest_folder=$2
+
+    pushd "${dest_folder}" || exit
+      echo "Downloading file: ${url}"
+      echo "To: ${dest_folder}"
+      curl -LO "${url}"
+    popd || exit
+}
+
+function extract_file_to_dir() {
+    local file_path=$1
+    local dest_dir=$2
+    local file_names=$3 # Extract only specific files
+
+    if [ ! -f "${file_path}" ]; then
+        echo "extract_file_to_dir(): File does not exists: ${file_path}" >&2
+        return 2
+    fi
+
+    if [ ! -d "${dest_dir}" ]; then
+        echo "extract_file_to_dir(): Dest dir does not exists: ${dest_dir}" >&2
+        return 2
+    fi
+
+    # shellcheck disable=SC2086 # We need to expand $file_names
+    tar xvzf "${file_path}" -C "${dest_dir}" ${file_names}
+}
+
+function get_download_url_from_github_api() {
+    local api_url=$1
+    if [ -z "${api_url}" ]; then
+        echo "get_download_url_from_github(): API URL not provided" >&2
+        return 2
+    fi
+
+    local arch
+    local kernel_name
+    arch="$(uname -m)"
+    kernel_name="$(uname -s)" # e.g. Linux / Darwin
+
+    # Little hack for grep as some projects are shipping release tarballs
+    # as x86_64 and some as amd64
+    if [[ "${arch}" == "x86_64" ]]; then
+        arch="-e ${arch} -e amd64"
+    elif [[ "${arch}" == "amd64" ]]; then
+        arch="-e ${arch} -e x86_64"
+    fi
+
+    # Call the API to find the latest binary matching kernel and architecture
+    local url_cmd
+    local download_url
+    url_cmd="curl -s ${api_url} | \
+          grep -o -E 'https://(.*).tar.gz' | \
+          grep -i '${kernel_name}' | \
+          grep ${arch}"
+    download_url=$(eval "${url_cmd}")
+
+    if [ -z "${download_url}" ]; then
+        echo "get_download_url_from_github(): download_url not available." >&2
+        echo "command used to get URL: $ ${url_cmd}" >&2
+        return 2
+    fi
+
+    # Return URL
+    echo "${download_url}"
+}
+
+# Installs helm into .venv path
+function install_helm() {
+    local helm_install_script=$1
+    local dest_dir=$2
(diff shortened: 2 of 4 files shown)


mpryc:
Ready to be reviewed.

scripts/install_dev_tools.sh
```diff
@@ -0,0 +1,211 @@
...
+    VENV="${venv_dir}"
+fi
+
+# Create download directory inside virtual env dir
+DWN_DIR=$(mktemp -p "${VENV}" -d -t "${TMP_DIR_PREFIX}XXXXX") || exit 2
+
+# Helm install
+if ! [ -x "$(command -v "${VENV}/bin/helm")" ]; then
+    echo "Installing helm CLI to: ${VENV}/bin/helm"
+    download_file_from_url "${HELM_INSTALL_SCRIPT}" "${DWN_DIR}" || cleanup_and_exit 2
```
KevinMGranger, line 178: Instead of appending every line with the cleanup_and_exit call, we can set a [signal trap](https://mywiki.wooledge.org/SignalTrap) to do the cleanup. This'll do the cleanup even if there's an exit condition we didn't catch, and even handle the happy path.

Makefile
```diff
@@ -67,7 +67,12 @@ git-blame:
 .git/hooks/pre-commit: scripts/pre-commit
 	./scripts/setup-pre-commit-hook
 
-dev-env: $(PELORUS_VENV) exporters git-blame .git/hooks/pre-commit
+.PHONY: cli_dev_tools
+cli_dev_tools:
+	./scripts/install_dev_tools.sh -v $(PELORUS_VENV)
+
+dev-env: $(PELORUS_VENV) cli_dev_tools exporters git-blame \
+         .git/hooks/pre-commit
```
KevinMGranger, line 75: Can we make this use file targets instead of a phony target? That will mean the script won't need to run every time. I know it will skip installing if it's present, but why not skip it in the first place?

scripts/install_dev_tools.sh
```diff
@@ -0,0 +1,211 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+#
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+# Match the .venv created by the Makefile
+DEFAULT_VENV="${SCRIPT_DIR}/../.venv"
```
KevinMGranger, line 21: The `SCRIPT_DIR` thing gets the real path in case it's a symbolic link, right?

- Could we pass in the venv from the makefile, so we're not duplicating the logic for it?
- If not, does it matter if it's a symlink or not?
- If it does matter, can we comment this line out with what it does? I only figured it out by trying it.

scripts/install_dev_tools.sh
```diff
@@ -0,0 +1,211 @@
...
+
+function download_file_from_url() {
+    local url=$1
+    local dest_folder=$2
+
+    pushd "${dest_folder}" || exit
+      echo "Downloading file: ${url}"
+      echo "To: ${dest_folder}"
+      curl -LO "${url}"
+    popd || exit
```
KevinMGranger, line 46: These exits should be an error, right?

scripts/install_dev_tools.sh
```diff
@@ -0,0 +1,211 @@
...
+        return 2
+    fi
+
+    if [ ! -d "${dest_dir}" ]; then
+        echo "extract_file_to_dir(): Dest dir does not exists: ${dest_dir}" >&2
+        return 2
+    fi
+
+    # shellcheck disable=SC2116,SC2046 # We need to expand $file_names
+    tar xvzf "${file_path}" -C "${dest_dir}" $(echo "${file_names}")
```
KevinMGranger, line 65: What does this echo do?

scripts/install_dev_tools.sh
```diff
@@ -0,0 +1,211 @@
...
+    local exit_val=$1
+    if [ -z "${DWN_DIR}" ]; then
+        echo "cleanup_and_exit(): Temp download dir not provided !" >&2
+    else
+      # Ensure dir exists and starts with prefix
+      if [ -d "${DWN_DIR}" ]; then
+          PELORUS_TMP_DIR=$(basename "${DWN_DIR}")
+          if [[ "${PELORUS_TMP_DIR}" =~ "${TMP_DIR_PREFIX}"* ]]; then
+              echo "Cleaning up temporary files"
+              eval rm -f "${DWN_DIR}/*"
```
KevinMGranger, line 137: Do we need this eval?

scripts/install_dev_tools.sh
```diff
@@ -0,0 +1,211 @@
+#!/bin/bash
```
KevinMGranger, line 1: Should this be `#!/usr/bin/env bash`?

scripts/install_dev_tools.sh
```diff
@@ -0,0 +1,211 @@
...
+}
+
+function print_help() {
+    printf "\nUsage: %s [OPTION]... -v [DIR]\n\n" % "$0"
+    printf "\tStartup:\n"
+    printf "\t  -h\tprint this help\n"
+    printf "\n\tOptions:\n"
+    printf "\t  -v\tpath to virtualenv DIR\n"
+
+    exit 2
```
KevinMGranger, line 154: Printing help isn't an error condition, right?

scripts/install_dev_tools.sh
```diff
@@ -0,0 +1,211 @@
...
+}
+
+### Options
+OPTIND=1
+while getopts "h?v:" option; do
+    case "$option" in
+    h|\?) print_help;;
+    v)    venv_dir=$OPTARG;;
+    esac
```
KevinMGranger, line 163: Should we handle the case of an unknown flag?

KevinMGranger (changes requested):
I like what I'm seeing, but I have one general thought: instead of each step having manual error handling, if we set up the signal trap, can we just do `set -e`?

KevinMGranger:
Also, one small nit-- it should be `scripts/install_dev_tools` if it's meant to be executed directly. You don't call `cat.elf`, right? :)

Makefile
mpryc: Added as well, but I think having this in a file is also nice to have in case I want to re-download only one tool e.g. newest and latest oc client then:
```
rm .venv/bin/oc
scripts/install_dev_tools
```

scripts/install_dev_tools.sh
mpryc: Of course.

scripts/install_dev_tools.sh
mpryc: - It gets the PATH from which it's being invoked. If we want to have exact symlink location then $(pwd) would be used, but I don't see any use case for that.

```
mkdir /tmp/test
cd ~
ln -s test /tmp/test
./test/install_dev_tools
# SCRIPT_DIR --->    /home/user/test/install_dev_tools
/tmp/test/install_dev_tools
# SCRIPT_DIR --->    /tmp/test/
```

Added comment inside the script.


As for the second question:
venv is passed from the Makefile, this one is only to have it included in the script in case someone wants to run it outside of Makefile.

```
cli_dev_tools:
	./scripts/install_dev_tools.sh -v $(PELORUS_VENV)
```

scripts/install_dev_tools.sh
mpryc: Already is:
```
$ TEST=$(pushd /tmp/nonexiting || exit)
$ echo $TEST
1
```

scripts/install_dev_tools.sh
mpryc: It expands file_names, so if the file_names is a string "oc kubectl" then it's being passed as two strings to the tar command. If we don't do it then tar will not find "oc kubectl" file in the archive (it's trying to search for the file with space in the filename).

scripts/install_dev_tools.sh
mpryc: Of course.

scripts/install_dev_tools.sh
mpryc: getopts does it for us:
```

$ ./scripts/install_dev_tools -n
./scripts/install_dev_tools: illegal option -- n

Usage: % [OPTION]... -v [DIR]


Usage: ./scripts/install_dev_tools [OPTION]... -v [DIR]

	Startup:
	  -h	print this help

	Options:

```

scripts/install_dev_tools.sh
mpryc: definatelly !

mpryc (commented):
@KevinMGranger left comments and uploaded new version that should address all your questions / improvements.

Makefile
KevinMGranger: Oh, my mistake! I thought the other targets depended upon `dev-env`, not `$(PELORUS_VENV)`. The way you had it would be fine.

The way I was thinking would be a target per tool, and the script could take an argument specifying which tool to download.

scripts/install_dev_tools.sh
KevinMGranger: Oh right, I didn't think about how it was trying to go upwards. Good call.

scripts/install_dev_tools.sh
KevinMGranger: Ah right, I forgot it uses the last exit code.

scripts/install_dev_tools
```diff
@@ -0,0 +1,215 @@
...
+done
+
+if [ -z "${venv_dir}" ]; then
+    VENV="${DEFAULT_VENV}"
+else
+    VENV="${venv_dir}"
+fi
+
+# Create download directory inside virtual env dir
+DWN_DIR=$(mktemp -p "${VENV}" -d -t "${TMP_DIR_PREFIX}XXXXX") || exit 2
```
KevinMGranger, line 178: The `-p` flag doesn't work on macOS. But it looks like both versions accept `TMPDIR` env var.

scripts/install_dev_tools.sh
KevinMGranger: Weird! That doesn't match my understanding of `help getopts` but that is indeed what it does.

scripts/install_dev_tools.sh
KevinMGranger: Right, but can't we just use regular word splitting with `$file_names`?

mpryc: Yes we can :) will update PR shortly.

scripts/install_dev_tools
mpryc: changed.

Makefile
```diff
@@ -67,7 +67,12 @@ git-blame:
 .git/hooks/pre-commit: scripts/pre-commit
 	./scripts/setup-pre-commit-hook
 
-dev-env: $(PELORUS_VENV) exporters git-blame .git/hooks/pre-commit
+$(PELORUS_VENV)/binary_tools_installed:
+	./scripts/install_dev_tools -v $(PELORUS_VENV)
+	touch $(PELORUS_VENV)/binary_tools_installed
```
KevinMGranger, line 72: In my last comment, I realized the old way you were doing it was fine. This way has the downside you pointed out where deleting one won't trigger an update.

mpryc: @KevinMGranger ok changed back. Thanks.

KevinMGranger (approved):

PR #381, opened 2022-03-01
🦜 Support for binary builds 🦜

eformat:
Initial support for OpenShift binary builds by annotating the build config each time you build (e.g. from a pipeline).

## Linked Issues?
resolves https://github.com/konveyor/pelorus/issues/317

## Testing Instructions
deploy the committime exporter and test using an openshift binary build. you pipeline must annotate the build config on each build run.

@redhat-cop/mdt


exporters/committime/README.md
```diff
@@ -27,3 +27,24 @@ Then we get commit data from the following systems through their respective APIs
...
+|---------------------------------|:----------------------------------|------------------------------:|
+| buildSpecRevisionGitCommit      | cae392a                           | Short or Long Git Commit Hash |
+| buildSpecSourceGitUri           | https://github.com/org/myapp.git  | URL of the Source Repository  |
+| buildSpecRevisionGitAuthorName  | joe.bloggs                        | Name of the committer         |
+
+Example command to put in your build pipeline each time you start an OpenShift `Build`:
+
+```bash
+oc annotate bc/${BUILD_CONFIG_NAME} --overwrite \
+  buildSpecSourceGitUri=${GIT_URL} \
```
mpryc, line 47: @eformat do you think it's worth matching the annotations with similar labels that are result of build process? 
Those labels are described [here](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.3/html/builds/managing-build-output#builds-output-image-labels_managing-build-output)

Here is example:

```
"io.openshift.build.source-location"
"io.openshift.build.commit.id" 
"io.openshift.build.commit.author"
"io.openshift.build.commit.ref"
```

exporters/committime/README.md
```diff
@@ -27,3 +27,24 @@ Then we get commit data from the following systems through their respective APIs
...
+  buildSpecRevisionGitCommit=${GIT_COMMIT} \
+  buildSpecRevisionGitAuthorName="${GIT_AUTHOR}"
+```
```
mpryc, line 50: Also note that we also missing branch / ref in your PR.

exporters/committime/collector_base.py
```diff
@@ -182,7 +182,11 @@ def get_metric_from_build(self, build, app, namespace, repo_url):
                     repo_url = self._get_repo_from_build_config(build)
 
             metric.repo_url = repo_url
-            commit_sha = build.spec.revision.git.commit
+
+            if build.spec.revision is None:
+                commit_sha = self._get_revision_from_build(build)
+            else:
```
mpryc, line 188: Should we add logic to get commit_sha if one was not resolved from the self._get_revision_from_build() func and not always when the build.spec.revision is not None?

exporters/committime/collector_base.py
```diff
@@ -191,7 +195,12 @@ def get_metric_from_build(self, build, app, namespace, repo_url):
 
             metric.commit_hash = commit_sha
             metric.name = app
-            metric.committer = build.spec.revision.git.author.name
+
+            if build.spec.revision is None:
```
mpryc, line 199: Same as above, get it when metric.committer is not valid from the:
`
metric.committer = self._get_author_from_build(build)
`

I don't know if such use case is valid when information is missing from build.spec.revision and have generic option to fill that data in the build?

exporters/committime/collector_base.py
```diff
@@ -264,6 +273,9 @@ def _get_repo_from_build_config(self, build):
         build_config = v1_build_configs.get(
             namespace=build.status.config.namespace, name=build.status.config.name
         )
+        if build.metadata.annotations.buildSpecSourceGitUri:
+            return build.metadata.annotations.buildSpecSourceGitUri
```
mpryc, line 277: Is there needed validation of URI at some point? e.g. valid git URI regex?

exporters/committime/collector_base.py
```diff
@@ -273,3 +285,25 @@ def _get_repo_from_build_config(self, build):
...
 
         return None
+
+    def _get_revision_from_build(self, build):
+        """
+        Determines the git revision from the parent BuildConfig
+        :param build: the Build resource
+        :return: revisions as a str or None if not found
+        """
+        if build.metadata.annotations.buildSpecSourceGitUri:
```
mpryc, line 295: Is this mistake? Should it be:

`if build.metadata.annotations.buildSpecRevisionGitCommit:`

I also don't understand the logic:

```
if something:
   return something
return None
```

This is equivalent to just:
`return something`

Because:

```
something = None
if something:
   return something
return None
```

Unless you wanted to return None if string is empty, then check the len(something) as well.

But as on lines 276-277, I would add some validaitons if the revision is valid regex.

exporters/committime/collector_base.py
```diff
@@ -273,3 +285,25 @@ def _get_repo_from_build_config(self, build):
...
+
+        return None
+
+    def _get_author_from_build(self, build):
+        """
+        Determines the git author from the parent BuildConfig
+        :param build: the Build resource
+        :return: author as a str or None if not found
+        """
+        if build.metadata.annotations.buildSpecRevisionGitAuthorName:
```
mpryc, line 306: Very similar comment as above with the unnecessary checking if value exists.

mpryc (changes requested):
@eformat reviewed your PR. Many thanks looks great with some minor questions/suggestions. Let me know if you will continue working on this PR or don't have bandwidth in which case we can look at reworking and merging.

Secondly some unit test would be nice to have.

Also this PR requires rebase as few things changed since it was created, but I will focus on getting it merged soon.

Thanks again.

exporters/committime/collector_base.py
KevinMGranger: That is handled in the `CommitMetric` class and/or the various collector implementations themselves.

mpryc:
I took the idea and implemented this as another PR #498. Co-authored by @eformat.
Please let me know if we can close this one?

mpryc:
I am closing this PR as it's being already addressed via PR #498

PR #382, opened 2022-03-01
Release 1.4.6

mpryc:

We will release Pelorus 1.4.6, so two files requires version bump.

charts/pelorus/charts/exporters/templates/_buildconfig.yaml:
  The default version will match TAG that will be created once this PR is merged.

charts/pelorus/Chart.yaml:
  This requires incremental change as we did change the _buildconfig.yaml which belongs to the pelorus chart.

Next steps are to crate TAG and release that are described in the:
  https://pelorus.readthedocs.io/en/latest/Development/#release-management-process

This is desired as we are going to merge PR #372, which will cause incompatibility with OCP <= 4.6, however that OCP version is no longer supported: [OpenShift Life Cycle Dates](https://access.redhat.com/support/policy/updates/openshift#dates) 

@redhat-cop/mdt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,12 @@ git-blame:
 .git/hooks/pre-commit: scripts/pre-commit
 	./scripts/setup-pre-commit-hook
 
-dev-env: $(PELORUS_VENV) exporters git-blame .git/hooks/pre-commit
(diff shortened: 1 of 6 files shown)


mpryc:
I have made a mistake and rebased this on top of enhance dev-env rather then master, will create new proper PR.

PR #391, opened 2022-03-11
CronJob should use different api starting from 1.21

### LINKED ISSUE #386: Deployment is failing on 4.7 cluster due to apiVersion in CronJob
PR #372 introduced incompatibility with OCP 4.7.

Was:
  https://docs.openshift.com/container-platform/4.7/rest_api/workloads_apis/cronjob-batch-v1beta1.html

OCP >=4.8:
  https://docs.openshift.com/container-platform/4.9/rest_api/workloads_apis/job-batch-v1.html

Related change:
  https://github.com/konveyor/pelorus/pull/372/files#diff-86099516e722665905ababa1882a6321cc5a8f9ad4281870dc48f3ddba42bc90

mpryc:
Fixes #386 where on ocp4.7 where CronJob used different API then
on newer OCP. We compare the KubeVersion as it's where the API
change happened with the following info:

batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob

## Linked Issues?

## Testing Instructions

@redhat-cop/mdt


charts/pelorus/templates/dashboard-snapshot-cronjob.yaml
```diff
@@ -38,7 +38,14 @@ subjects:
...
 ---
+{{- /*
+Issue #386 - we need to use different APIs
+*/}}
+{{ if semverCompare ">=1.21" .Capabilities.KubeVersion.Version -}}
 apiVersion: batch/v1
+{{- else }}
+apiVersion: batch/v1beta1
+{{- end }}
 kind: CronJob
```
weshayutin, line 49: nice! didn't realize jinja was an option :)  that's good to know

weshayutin (approved):

charts/pelorus/templates/dashboard-snapshot-cronjob.yaml
mpryc: @weshayutin yeah, but there are quite a lot of limitations... first was playing with:

{{- $useCronJobV1 := .Capabilities.APIVersions.Has "batch/v1" -}}

Even if it should work... Helm has some wired problems while testing and decided to check for the k8s version in which API was changed.

KevinMGranger: That should be "batch/v1/CronJob", no? Maybe that's the issue?

mpryc: @KevinMGranger yes! That worked perfectly. Updated PR with your suggestion.

KevinMGranger (approved):

PR #395, opened 2022-03-15
Extract image SHAs from status.containerStatuses as well as spec.containers

### LINKED ISSUE #330: Support for other image references in deploytime exporter
Deploying an app with tekton pipeline and another registry... the resulting pod has not the correct SHA in its place. I mean, the correct way is:
```yaml
spec:
  containers:
    - name: appname
       image: >-
         image-registry.openshift-image-registry.svc:5000/project/appname@sha256:afg1zk03979456a9023eff2587652d30
```

But, In my case, when use quay as registry for my docker images, the same snippet is slightly different:
```yaml
spec:
  containers:
    - name: appname
       image: >-
         quay-registry.project.cluster-url.com/organization/appname:latest
```

It does not have a SHA in the image URL. So my proposal is to take the image SHA from somewhere else. For example:
```yaml
status:
  containerStatuses:
    - name: retail-queloque
       imageID: >-
         quay-registry.project.cluster-url.com/organization/appname@sha256:afg1zk03979456a9023eff932587652d30
```

For the deploytime exporter works well, I had to change the line number 132 for this one:
```python
images = (sha for s in pod.status.containerStatuses if (sha := image_sha(s.imageID)))
```

This worked for me.

KevinMGranger:
## Behavior Changes

The deploytime exporter will now check a pod's `status.containerStatuses` for image IDs as well.

## Linked Issues?

resolves #330

## Testing Instructions

I'm planning on adding a simple mock for our deploytime exporter as well, which will have some images that don't have the ID in their spec.

@redhat-cop/mdt

## Remaining Work

- [x] I need to see if `pod.status.containerStatuses` might list sidecars from MutatingAdmissionControllers (e.g. from Istio / Service Mesh), which we wouldn't want to count. It's also possible that those are already included in `spec.containers`, which would be a separate issue.
- [x] add a unit test / address new unit test failures
- ~~[ ] I need to add the mock as mentioned above.~~

### DIFF
--- a/exporters/deploytime/app.py
+++ b/exporters/deploytime/app.py
@@ -92,13 +92,25 @@ class DeployTimeCollector:
         return namespaces
 
 
-@attr.define(kw_only=True)
+@attr.frozen(kw_only=True)
 class DeployTimeMetric:
     name: str
     namespace: str
+    # WARNING: do not mutate the dict after hashing or things may break.
     labels: dict[str, str]
     deploy_time: object
     image_sha: str
+
+    def __hash__(self):
+        return hash(
+            (
+                self.name,
+                self.namespace,
+                hash(tuple(self.labels.items())),
+                self.deploy_time,
+                self.image_sha,
+            )
+        )
 
 
 def image_sha(image_url_or_id: str) -> Optional[str]:
--- a/exporters/deploytime/app.py
+++ b/exporters/deploytime/app.py
@@ -101,12 +101,20 @@ class DeployTimeMetric:
     image_sha: str
 
 
-def image_sha(img_url: str) -> Optional[str]:
+def image_sha(image_url_or_id: str) -> Optional[str]:
+    """
+    Gets the hash of the image, extracted from the image URL or image ID.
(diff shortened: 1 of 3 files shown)


etsauer:
Bump. Any updates on this?

KevinMGranger:
@etsauer the update is that I had a nice vacation and am back to working on this today. All that's left is to add / check the mocked API for testing this, and investigate the unit test failure that CI turned up.

etsauer:
@KevinMGranger 🤣 welcome back!

KevinMGranger:
I forgot that we only have mocks for committime for now.

Mocking this can come later-- the unit test should be sufficient.

KevinMGranger:
Hmm, looks like our python version pinning is complaining about the new version. @mpryc, is there a reason we restricted the patch version of the python version?

mpryc:
@KevinMGranger there is not, I think it's safe to include the removal of patch in this PR.

etsauer (approved):
LGTM

PR #402, opened 2022-04-04
Disable fail-fast on all python github actions

KevinMGranger:
Also add a version matrix to unit tests and integration tests.

This means that actions for certain python versions will not be cancelled if there's a failure with another version.

This makes it easier to diagnose python-version-specific issues.

This is a draft until the python version restriction issue is resolved.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -18,13 +18,17 @@ on:
 jobs:
   integration-test:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.9', '3.10']
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Set up Python 3.9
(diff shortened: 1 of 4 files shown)


mpryc:
@KevinMGranger ah we can't just remove patch because the comparison for:

X.Y (without Z) treats Z as 0, so:

```
PYTHON_VER_MAX=3.10
  $(error $(PYTHON_BINARY) needs to be at >= $(PYTHON_VER_MIN)\
                           and <= $(PYTHON_VER_MAX))
```

Will fail for any version with patch number such as 3.10.1, because 3.10.1 < 3.10.0

We could have here:
PYTHON_VER_MAX=3.11
and then:
```
  $(error $(PYTHON_BINARY) needs to be at >= $(PYTHON_VER_MIN)\
                           and < $(PYTHON_VER_MAX))
```

Or ?:
PYTHON_VER_MAX=3.11.999

KevinMGranger:
Continuing convo re: version checking in #403

PR #404, opened 2022-04-05
Remove python patch version check

### LINKED ISSUE #403: Python patch version should not be checked
We have reasonable confidence that patch versions will not affect the correctness of the code.

Right now, the max version we marked as supported is 3.10.2. However, 3.10.4 is out, resulting in breakage.

[As per the conversation in the fail-fast PR](https://github.com/konveyor/pelorus/pull/402#issuecomment-1088657868), we can either use the max version an exclusive / open, or hard-code its patch to something like .999.

As an alternative, we could also pull the information out into tuples/lists, e.g. `(version_info[0], version_info[1]) <= MAX`.

KevinMGranger:
Also move version checking to separate script for cleanliness.

Python patch versions (3.10.PATCH) will no longer be checked.

This also moves the version checking into a separate script for cleanliness-- we no longer need to print and re-parse versions, and can do the simplest approach for major and minor version checking.

## Linked Issues?

resolves #403

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,6 @@
 
 # Minimal python version supported by exporters
 PYTHON_BINARY=python3
-PYTHON_VER_MIN=3.9
-PYTHON_VER_MAX=3.10.2
 
 ifndef PELORUS_VENV
   PELORUS_VENV=.venv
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -14,17 +12,12 @@ ifeq (, $(shell which $(PYTHON_BINARY) ))
...
 $(info Found system python version: $(SYS_PYTHON_VER));
-PYTHON_VER=$(shell $(PYTHON_BINARY) -c 'from pkg_resources import packaging; \
-  print("%s" % (packaging.version.parse("$(PYTHON_VER_MAX)") >= \
-  packaging.version.parse("$(SYS_PYTHON_VER)") >= \
-  packaging.version.parse("$(PYTHON_VER_MIN)")))')
-
-ifeq ($(PYTHON_VER), False)
-  $(error $(PYTHON_BINARY) needs to be at >= $(PYTHON_VER_MIN)\
-                           and <= $(PYTHON_VER_MAX))
+PYTHON_VER_CHECK=$(shell $(PYTHON_BINARY) scripts/python-version-check.py $(PYTHON_BINARY))
```
mpryc, line 17: Why PYTHON_BINARY is passed as argument? If it's being invoked by interpreter already it will get proper version_info.

If you wish to get path to interpreter then python script  can get it:
```
import sys
print(sys.executable)
```

KevinMGranger: I was disappointed to find the executable name wasn't in `sys.argv`, I guess I should have kept looking!

It wasn't to get `version_info`, it was because we were already using it in the error message. That looks like it'll be even cleaner.

scripts/python-version-check.py
```diff
@@ -0,0 +1,14 @@
+from sys import argv, exit, version_info
```
mpryc, line 1: Ah three minor nits:
1. rename to python-version-check ?
2. add interpreter / just so default python3 will be executed if not directly invoked by the /path/to/python python-version-check ?
  #!/usr/bin/env python3
3. License header ?

KevinMGranger: I normally advocate for dropping the `.py`, but we intend to use this with the python version determined in the makefile (which could be overridden).

Do we have a guide on how to do the license header correctly?

mpryc: Ok, then we can leave .py. Yes after we discussed this offline I've added one here:

https://github.com/konveyor/pelorus/blob/master/scripts/install_dev_tools

PR #405, opened 2022-04-05
Remove duplicate charts values.yaml file

weshayutin:
The default values.yaml was missing exporters:
on the top line, making a little harder to just add
new configured exporters for users.

https://github.com/konveyor/pelorus/blob/master/docs/Configuration.md


KevinMGranger:
That values file isn't used, `charts/pelorus/values.yaml` is. We should consider removing this one.

weshayutin:
Thanks for the review @KevinMGranger, welcome back!

PR #406, opened 2022-04-06
Cleanup and documentation of the python requirements

mpryc:
Fixes #359

Scanned the code to check every usage of the particular
import, then used mixture of pipreqs and pipdeptree to find
appropriate packages.

Note that pipreqs can not be used for this project as some
dependencies found by this tool are obsolete, examples are
azure_storage, gitlab.

@redhat-cop/mdt

PR #408, opened 2022-04-06
Make chart-lint github action trigger on changes to requirements-dev.txt

KevinMGranger:
PR #406 showed that this dependency is important.

The chart-lint github action will now run when requirements-dev.txt is changed,
because that could affect the yamllint and yamale versions.


mpryc:
@KevinMGranger actually could you add as well Makefile? The logic of ct is in that file, so worth triggering this workflow for it.

mpryc (changes requested):
As in the comment, please add Makefile as well.

KevinMGranger:
What do you mean by adding the Makefile? The github action already uses the makefile.

PR #411, opened 2022-04-07
Better missing attribute errors

KevinMGranger:

This adds `get_nested` and uses it to give us better error messages for the committime collector.

We often have to deal with heavily nested data, that may or may not be present or properly formed.  When something goes wrong, we often deal with unhelpful error messages like `AttributeError: None does not contain attribute foo`. This issue even motivated the python folks to give [better error messages for similar errors in 3.11](https://peps.python.org/pep-0657/).

Another problem is with how most validation / conversion processes work: if it fails at the first issue, then you fix that... and it just fails at the next issue... it'll take forever to get through. That's not necessary if later steps are independent of the earlier ones! The python folks also recognize this issue and are introducing [exception groups](https://peps.python.org/pep-0654/) because of it.

Our committime collector deals with _tons_ of nested data, and its error messages have been limited to telling you which build had an issue-- but nothing else.

This PR addresses all of the above:

- make it easy to find where in a nested access things went wrong
- collect all data issues at once instead of failing early
- improve error messages for the committime collector

## Testing Instructions

Run the committime integration test and observe that the results are the same.

Remove some nested field from the mocked build response, such as the git author, and observe the error message (you'll need to run pytest with `-s` to capture output.)

### DIFF
--- a/exporters/committime/__init__.py
+++ b/exporters/committime/__init__.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Optional
+from typing import Optional
 
 import attr
 import giturlparse
@@ -14,7 +14,7 @@ SUPPORTED_PROTOCOLS = {"http", "https", "ssh", "git"}
 @attr.define
 class CommitMetric:
     name: str = attr.field()
-    labels: Any = attr.field(default=None, kw_only=True)
+    labels: dict = attr.field(default=None, kw_only=True)
     namespace: Optional[str] = attr.field(default=None, kw_only=True)
 
     __repo_url = attr.field(default=None, init=False)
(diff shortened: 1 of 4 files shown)


exporters/committime/__init__.py
```diff
@@ -91,3 +95,31 @@ def __parse_repourl(self):
...
+
+_COMMIT_METRIC_MAPPING = dict(
+    build_name="metadata.name",
+    build_config_name="metadata.labels.buildconfig",
+    namespace="metadata.namespace",
+    commit_hash="spec.revision.git.commit",
+    committer="spec.revision.git.author.name",
+    image_location="status.outputDockerImageReference",
+    image_hash="status.output.to.imageDigest",
+)
```
mpryc, line 125: I am missing some metrics removed from https://github.com/konveyor/pelorus/pull/411/files#diff-2d6d20dd9681db562d438793a6b1730a740965526f49c4b7b34fbdb8d0fa354bL184-L196

Those are ones which I found missing: repo_url, labels, name, commit_timestamp

Could you explain why they are out ?

Could this be moved to the top of the file? Easier to read such mappings when they are declared in one place (below SUPPORTED_PROTOCOLS).

mpryc (commented):
Just one question, other then that lgtm.

exporters/committime/__init__.py
KevinMGranger: Alternatively, for labels, I could make a custom setter. But that seems like more work than necessary.

mpryc: Yep, let's keep it simple.

Class attribute sounds fine.

For the rest a little doc would be sufficient, so we know later why it was done this way.

PR #412, opened 2022-04-08
Github actions should use Makefile

### LINKED ISSUE #374: Github actions should use Makefile to setup test & build environment
We want to have one common way of setting up and tearing down test and build environment for pelorus project, so it's easier to fix&maintain that part.

Currently github actions are using separate pip invoke to run install of deps. This issue is to fix it and ensure all actions and other parts of the pelorus project is using setting up dev env via Makefile.

mpryc:
This partialy fixes #374, integration and unit tests have now
makefile targets and those are used in the github workflows.

Some github workflow file formatting was done to make it easier
to read.

Also some of the workflow files were using old setup-python
version, which is now corrected and all of them uses matrix
rather then single python version.

## Linked Issues?

## Testing Instructions

@redhat-cop/mdt


mpryc:
Remaining tests are bats (conftest), however those will be presented in separate PR as the approach is slightly different to get this framework working.

Makefile
```diff
@@ -82,6 +82,26 @@ minor-release:
...
+# Integration tests
+
+.PHONY: integration-tests
+integration-tests: $(PELORUS_VENV)
+	. ${PELORUS_VENV}/bin/activate && \
+	./scripts/run-integration-tests
+
+# Unit tests
+.PHONY: unit-tests
+unit-tests: $(PELORUS_VENV)
```
KevinMGranger, line 94: Can we add the following comment to explain the CLI flags?

```
# -r: show extra test summaRy: (a)ll except passed, (p)assed
# because using (A)ll includes stdout
# -m filters out integration tests
```

exporters/requirements-dev.txt
```diff
@@ -1,6 +1,9 @@
 # Needed by exporters/tests
 pytest
 
+# Integration Tests
+httpie
+
```
KevinMGranger, line 6: We actually don't need this. I was initially using it because httpie had nicer timeout/retry flags than curl, but then forgot to use it myself 😅 

Looks like you've handled that in the script anyway.

scripts/run-integration-tests
```diff
@@ -0,0 +1,112 @@
...
+    RUNNING_ID="$(${CONTAINER_ENGINE} ps -a -q -f name=${TEST_CONTAINER_NAME})"
+    if [ "${RUNNING_ID}" != "" ]; then
+        echo "Removing test container: ${TEST_CONTAINER_NAME} with ID: ${RUNNING_ID}"
+        ${CONTAINER_ENGINE} rm --force "${RUNNING_ID}"
+    fi
+}
+
+function cleanup_and_exit() {
+    local exit_val=$1
+    echo "Exitting..."
```
KevinMGranger, line 64: typo: exiting

scripts/run-integration-tests
```diff
@@ -0,0 +1,112 @@
...
+
+trap 'cleanup_and_exit' INT TERM EXIT
+
+# Check if test container already exists, if yes remove it
+
+cleanup_mockoon_pod
+
+# We need to check if port 3000 is used by other process
+
+if lsof -Pi :"${MOCKOON_PORT}" -sTCP:LISTEN -t >/dev/null ; then
```
KevinMGranger, line 79: Do we need to do this? Won't the `docker/podman run` call fail if it's in use?

I'll test this myself when I test the script as a whole.

KevinMGranger (commented):
Some small nits, but I'll have real feedback once I run the script to test.

exporters/requirements-dev.txt
mpryc: ok removed.

Makefile
mpryc: done

scripts/run-integration-tests
KevinMGranger: That makes a lot of sense. Will test it shortly.

mpryc: Ok please resolve this conversation if it's tested by you.

KevinMGranger: This mostly works. macOS users will need to set up a podman machine themselves, mounting the mocks dir in advance themselves.

I think that's an acceptable manual step that we can document in the dev docs.

I've added two small changes that I hope you won't mind:

1. Make `MOCK_JSON` overrideable for the above case
2. Add the test script to the list of shellcheck targets.

PR #413, opened 2022-04-08
WIP: add a 'make help', 'make list' action

weshayutin:
Assist new developers in better understanding the options
in the make file.

This is not required, but may help developers on board to Pelorus.

## Testing Instructions
```
make help
```
can also be aliased to make list or $other

# WDYT?
Is this helpful, is the additional cruft in the makefile worth it?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,11 +25,21 @@ CHART_TEST=$(shell which ct)
 SHELLCHECK=$(shell which shellcheck)
 SHELL_SCRIPTS=./scripts/pre-commit ./scripts/setup-pre-commit-hook ./demo/demo-tekton
 
+.PHONY: help
+# tabs below are required for lines starting in @
+help:
+	@printf "%-20s %s\n" "Target" "Description"
+	@printf "%-20s %s\n" "------" "-----------"
+	@make -pqR : 2>/dev/null \
+        | awk -v RS= -F: '/^# File/,/^# Finished Make data base/ {if ($$1 !~ "^[#.]") {print $$1}}' \
+        | sort \
+        | egrep -v -e '^[^[:alnum:]]' -e '^$@$$' \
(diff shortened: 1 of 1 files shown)


KevinMGranger:
I like the idea of having information about each target.

However, for the current implementation:

- I think it's hard to understand
- It took 4.9 seconds to run on my machine
- I'm seeing descriptions listed on the wrong targets

As an alternative, [the implementation described in this article](https://swcarpentry.github.io/make-novice/08-self-doc/index.html) is much simpler.

KevinMGranger:
Here's a quick example I've tested on macOS. I think the flags should work the same.

```
## help	print each make target with a description
.PHONY: help
help:
	@sed -n 's/^## //p' Makefile | column -t -s $$'\t'
```

One more thing: if we put `help` at the top of the file, it becomes the new default target. Is that what we want? I think so, but let's be sure about it and make sure `default` reflects that.

KevinMGranger:
I've got my own changes for this here: https://github.com/KevinMGranger/pelorus/commit/985dda4495606c31dc364cab70c91de5d5bb1049

I think it makes the table formatted nicely, and corrects some inaccurate help descriptions. Let me know what you think.

weshayutin:
> I've got my own changes for this here: [KevinMGranger@985dda4](https://github.com/KevinMGranger/pelorus/commit/985dda4495606c31dc364cab70c91de5d5bb1049)
> 
> I think it makes the table formatted nicely, and corrects some inaccurate help descriptions. Let me know what you think.

Yup, I like your changes.  I'll pull that into my change and finish up and hold for Migi's makefile additions

weshayutin:
@mpryc @KevinMGranger ok.. think this is ready now :)

mpryc (approved):

PR #415, opened 2022-04-12
Github action to use Makefile for conf-tests

mpryc:
Final fix for #374, conf-tests have new make target and it's
being used in the github action.

@redhat-cop/mdt


scripts/install_dev_tools
```diff
@@ -261,3 +265,27 @@ if should_cli_be_installed "promtool" "${cli_tools_arr[@]}" && \
...
+# Install bats via repository
+if should_cli_be_installed "bats" "${cli_tools_arr[@]}" && \
+    ! [ -x "$(command -v "${VENV}/bin/bats")" ]; then
+    if ! [ -x "$(command -v git)" ]; then
+        echo "git command not found"
+        exit 2
+    fi
+    git clone "${BATS_REPO_URL}" "${DWN_DIR}/bats"
+    pushd "${DWN_DIR}/bats"
+    ./install.sh "../../"
```
KevinMGranger, line 288: This will be relative to the `DWN_DIR` from `mktemp`, which is system-dependent. This didn't work on macOS.

Should we pass the venv as the target?

The other thing is-- if this is a directory under `/tmp`, wouldn't this install it to the root fs? Shouldn't that fail unless this script is run as root?

KevinMGranger, line 288: This will be relative to the `DWN_DIR` from `mktemp`, which is system-dependent. This didn't work on macOS.

Should we pass the venv as the target?

The other thing is-- if this is a directory under `/tmp`, wouldn't this install it to the root fs? Shouldn't that fail unless this script is run as root?

mpryc: Where was the directory created ?

The idea is to create directory inside .venv/ and not the /tmp/ target, that is why we have before calling mkdir:

TMPDIR="${VENV}"

https://github.com/konveyor/pelorus/blob/master/scripts/install_dev_tools#L206

This should work on Mac OS X as well. Unless you have other setting specified such as _CS_DARWIN_USER_TEMP_DIR or _CS_DARWIN_USER_TEMP_DIR

But I will change it to pass venv, will be more robust

mpryc: Secondly this script should not be ran as root, we really don't want that.

mpryc: I am closing this comment as it is the same as previous one where the conversation happens.

mpryc: @KevinMGranger ok in the newest revision I have changed two things:

To ensure download dir is created always in the .venv regardless of the system env I am passing this directly to the mktemp(1) command rather then relying on the env variable:

```
> diff
-DWN_DIR=$(TMPDIR="${VENV}" mktemp -d -t "${TMP_DIR_PREFIX}XXXXX") || exit 2
+DWN_DIR=$(mktemp --tmpdir="${VENV}" -d -t "${TMP_DIR_PREFIX}XXXXX") || exit 2

```

Secondly to calculate absolute path of the place where bats is installed I am ensuring script enters this directory first and then takes this value from simple pwd(1). This should be platform independent/works anywhere.

scripts/install_dev_tools
```diff
@@ -203,7 +207,7 @@ cli_tools_arr=("${cli_tools//' '/}")
 cli_tools_arr=(${cli_tools_arr//','/ })
 
 # Create download directory inside virtual env dir
-DWN_DIR=$(TMPDIR="${VENV}" mktemp -d -t "${TMP_DIR_PREFIX}XXXXX") || exit 2
+DWN_DIR=$(mktemp --tmpdir="${VENV}" -d -t "${TMP_DIR_PREFIX}XXXXX") || exit 2
```
KevinMGranger, line 210: This is incompatible with macOS. It worked as-is, right?

Makefile
```diff
@@ -105,6 +105,12 @@ test-prometheusrules: $(PELORUS_VENV)
 	. ${PELORUS_VENV}/bin/activate && \
 	./_test/test_prometheusrules
 
+# Conf tests
+.PHONY: conf-tests
+conf-tests: $(PELORUS_VENV)
```
KevinMGranger, line 110: This needs to depend on `dev-env`, otherwise `bats` won't be installed.

KevinMGranger (changes requested):
Just need those two things changed, but otherwise it looks good.

That said, conftest.sh fails on macOS because of how grep is used, but that's a problem for another day :)

KevinMGranger (approved):

PR #418, opened 2022-04-12
More commit debug logs, exit Build inspection early if not yet ready

### LINKED ISSUE #419: in-progress builds should be gracefully skipped if they're missing data
Builds with `status.phase: Running` may have missing output data. We currently treat this as a warnable issue, even though this is expected. We should fail gracefully, giving this information in an `INFO` log, instead of a `WARNING`.

We can still optimistically try to gather the data, in case it's still running after outputting an image.

KevinMGranger:
Adds more committime debug logs and adds some verbosity to debug logging in general.

Also skips Builds that aren't ready yet, without further inspection. This avoids cluttering logs for builds that have failed or haven't completed yet.

Closes #419.

### DIFF
--- a/exporters/committime/collector_base.py
+++ b/exporters/committime/collector_base.py
@@ -174,6 +174,10 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
         errors = []
         try:
             metric = commit_metric_from_build(app, build, errors)
+
+            if not self._is_metric_ready(namespace, metric, build):
+                return None
+
             if repo_url:
                 metric.repo_url = repo_url
             elif get_nested(build, "spec.source.git", default=None):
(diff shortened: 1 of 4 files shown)


exporters/committime/collector_base.py
```diff
@@ -174,6 +174,10 @@ def get_metric_from_build(self, build, app, namespace, repo_url):
         errors = []
         try:
             metric = commit_metric_from_build(app, build, errors)
+
+            if not self._is_metric_ready(namespace, metric, build):
+                return None
```
mpryc, line 179: Should there be at least warning/debug that the metric isn't yet available?

exporters/committime/collector_base.py
```diff
@@ -224,6 +209,72 @@ def get_metric_from_build(self, build, app, namespace, repo_url):
             )
             logging.debug(e, exc_info=True)
             return None
+        except Exception as e:
```
mpryc, line 212: I am not always keen on catching Exception, but in this case I think it's ok'ish...

exporters/committime/collector_base.py
```diff
@@ -224,6 +209,72 @@ def get_metric_from_build(self, build, app, namespace, repo_url):
...
+                    "Build %s/%s has status %s and doesn't have an image_hash yet, skipping",
+                    namespace,
+                    build.metadata.name,
+                    build_status,
+                )
+                return False
+            else:
+                return True
+        else:
+            return True
```
mpryc, line 248: So here the logic is as follows:

if something1 goes wrong, return False
if something2 else goes wrong, return False
return True

So if there is something3 that goes wrong you are assuming the metric is ready. I think it would be better to get positive filter rather then negative one, but maybe it's impossible??

mpryc (commented):
Just 3 small comments, that may not require rework if you disagree/answer.

exporters/committime/collector_base.py
KevinMGranger: The logging happens in `_is_metric_ready` because it can offer greater perspective as to why it was skipped.

exporters/committime/collector_base.py
KevinMGranger: That's how it already was. I just wanted to differentiate an AttributeError from any other Exception, because we already treated them the same way.

exporters/committime/collector_base.py
KevinMGranger: According to the docs, the only other value it could have is `Complete`. In this case, I'd like to follow the Robustness Principle, and try to process any status we don't recognize. It'll fail if it's missing info anyway.

exporters/committime/collector_base.py
mpryc: Ok great.

exporters/committime/collector_base.py
mpryc: Ok, no change in the logic then.

exporters/committime/collector_base.py
mpryc: I am fine with this approach. Thanks for clarification

PR #424, opened 2022-04-19
Test to cover JIRA token and password authentication.

### LINKED ISSUE #423: JIRA failure exporter should not use basic auth with username/password
JIRA exporter should use mail/token for the basic auth as the username/password was deprecated.

Documentation also requires changing to ensure it's clearly stated.

```
Basic authentication with passwords is deprecated.  For more information, see: https://developer.atlassian.com/cloud/confluence/deprecation-notice-basic-auth/
```

mpryc:
Added test which checks for the deprecated password authentication.

JIRA uses same method to authenticate in the cloud jira instance for both deprecated password authentication and token based auth.

This may lead to the situation where wrong API key is actually considered password and then failure exporter is presenting deprecation error.

This change also moves connection func to separate private one for easy testing and readibility + may in the future include more sophisticated auth methods.

Added proper headers to the files which were modified.

resolves #423

@redhat-cop/mdt

### DIFF
--- a/exporters/failure/collector_jira.py
+++ b/exporters/failure/collector_jira.py
@@ -54,7 +54,7 @@ class JiraFailureCollector(AbstractFailureCollector):
             logging.error(
                 "Status: %s, Error Response: %s", error.status_code, error.text
             )
-            raise error
+            raise
 
         return jira_client
 
--- a/exporters/failure/collector_jira.py
+++ b/exporters/failure/collector_jira.py
@@ -1,9 +1,27 @@
+#!/usr/bin/env python3
+#
+# Copyright Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+#
+
 import logging
 from datetime import datetime
 
 import pytz
 from collector_base import AbstractFailureCollector, TrackerIssue
 from jira import JIRA
+from jira.exceptions import JIRAError
 
 import pelorus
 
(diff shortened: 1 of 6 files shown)


mpryc:
@KevinMGranger I was wondering if we should actually catch the JIRAError and then tell that this may be as well wrong api key instead of just passing original JIRAError with:

```
JIRAError('Basic authentication with passwords is deprecated.  For more information, see: https://developer.atlassian.com/cloud/confluence/deprecation-notice-basic-auth/
```

Note that I have created test pelorus atlassian.net that is used in the connection tests, they require valid e-mail to create  api key to communicate with JIRA instance and I am using it in test.

exporters/failure/collector_jira.py
```diff
@@ -17,15 +35,36 @@ def __init__(self, user, apikey, server, projects):
...
+        jira_client = JIRA(options={"server": server}, basic_auth=(email, api_key))
+
+        # Ensure connection was performed
+        try:
+            jira_client.session()
+        except JIRAError as error:
+            logging.error(
+                "Status: %s, Error Response: %s", error.status_code, error.text
+            )
+            raise error
```
KevinMGranger, line 57: Using `raise` without the `error` will re-throw the exception while keeping the original traceback correct

pyproject.toml
```diff
@@ -4,6 +4,12 @@ src_paths = ["exporters" ,"scripts"]
 known_first_party = ["pelorus"]
 
 [tool.pytest.ini_options]
+pythonpath = [
+    "exporters/comitttime",
+    "exporters/deploytime",
+    "exporters/failure",
+    "exporters/pelorus"
+]
```
KevinMGranger, line 12: I'd rather we make `failure` a proper module so we can just change the imports to `from failure.$submodule ...`. Seems more proper that way, right? If you agree, I can push the small change to the PR.

exporters/tests/test_failure_exporter.py
```diff
@@ -0,0 +1,101 @@
...
+            "fake@user.com",
+            "WIEds4uZHiCGnrtmgQPn9E7D",
+        )
+    ],
+)
+def test_jira_prometheus_register(server, username, apikey):
+    def mock_search_issues(self):
+        return []
+
+    monkeypatch = MonkeyPatch()
```
KevinMGranger, line 93: The preferred way to use monkeypatch is as a fixture-- we can just make a parameter named `monkeypatch` and it will work.

exporters/tests/test_failure_exporter.py
```diff
@@ -0,0 +1,101 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+#
+
+import pytest
+from _pytest.monkeypatch import MonkeyPatch
```
KevinMGranger, line 19: `MonkeyPatch` is exposed within `pytest`, no need to look for it in a private module.

KevinMGranger (changes requested):
I've got some small notes, which I've modified locally and can push if you wish.

Overall:

- What does each test do? Can we document them?
- #423 says the deprecation should be documented. So we either need documentation in this PR or to not mark it as resolved with this PR.
- I'm confused by a deprecation _raising an exception_ -- shouldn't it let you keep going but warn you?
- If we're switching to insisting on an email address, should we change the envionment variable as well?

exporters/tests/test_failure_exporter.py
mpryc: @KevinMGranger normally yes by fixture, but for some reason pytest was failing that the module was not found.

pyproject.toml
mpryc: @KevinMGranger  Ok, please do.

exporters/tests/test_failure_exporter.py
mpryc: Yeah, I had troubles with fixture here, but will play with it.

KevinMGranger:
The small changes I made are here: https://github.com/kevinmgranger/pelorus/tree/issue_423

KevinMGranger (approved):

Issue #278, opened 2021-05-12
Failure-exporter pod will crash if doesn't find a Jira project with scrum template on it.


varodrig:
When there is no a project with a scrum template created in Jira, the failure-exporter pod will crash.
Once a project is created with that template the deployment works.

![image](https://user-images.githubusercontent.com/64439402/118033235-938c8d80-b336-11eb-83da-d044bc5fa9b6.png)
logs

![image](https://user-images.githubusercontent.com/64439402/118033322-ab641180-b336-11eb-8b7d-e64eb16e0ba1.png)
  File "/opt/app-root/lib64/python3.8/site-packages/jira/resilientsession.py", line 56, in raise_on_error
    raise JIRAError(
jira.exceptions.JIRAError: JiraError HTTP 400 url: https://testset.atlassian.net/rest/api/2/search?jql=type%3Dbug+and+priority%3Dhighest&startAt=0&validateQuery=True&maxResults=50
	text: The value 'bug' does not exist for the field 'type'.

	response headers = {'Server': 'AtlassianProxy/1.19.3.1', 'cache-control': 'no-cache, no-store, no-transform', 'Content-Type': 'application/json;charset=UTF-8', 'Strict-Transport-Security': 'max-age=315360000; includeSubDomains; preload', 'Date': 'Wed, 12 May 2021 19:28:01 GMT', 'ATL-TraceId': '1d08e6fe039e3e0f', 'x-arequestid': '74f459b5-70dd-4339-80d3-2e343a890cfb', 'x-aaccountid': '557058%3A45c754ef-9829-4895-81ef-863c688bdb60', 'X-XSS-Protection': '1; mode=block', 'Transfer-Encoding': 'chunked', 'timing-allow-origin': '*', 'x-envoy-upstream-service-time': '95', 'X-Content-Type-Options': 'nosniff', 'Connection': 'close', 'Expect-CT': 'report-uri="https://web-security-reports.services.atlassian.com/expect-ct-report/global-proxy", enforce, max-age=86400'}
	response text = {"errorMessages":["The value 'bug' does not exist for the field 'type'."],"warningMessages":[]}

etsauer:
@malacourse is this a bug?

mpryc:
This issue is resolved with #425

PR #425, opened 2022-04-20
Fix issue #278 - failure pod crashing

mpryc:
Fix issue #278 - failure pod crashing
    
We should log error rather then crash pod if JIRA is missing type required by failure exporter. JIRA type may be added after pelorus deployment.

Depends on #424 

@redhat-cop/mdt

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,13 +4,8 @@ src_paths = ["exporters" ,"scripts"]
 known_first_party = ["pelorus"]
 
 [tool.pytest.ini_options]
-pythonpath = [
-    "exporters/comitttime",
-    "exporters/deploytime",
-    "exporters/failure",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -4,7 +4,14 @@ src_paths = ["exporters" ,"scripts"]
 known_first_party = ["pelorus"]
 
 [tool.pytest.ini_options]
+pythonpath = [
+    "exporters/comitttime",
+    "exporters/deploytime",
+    "exporters/failure",
+    "exporters/pelorus"
+]
```
KevinMGranger, line 12: I thought we didn't need these. Are you still seeing issues without these?

mpryc: you are right, removed from latest rev.

PR #427, opened 2022-04-22
Fix for #304 to allow JIRA obtain more then 50 results

### LINKED ISSUE #304: failure-exporter limited to only 50 results

mpryc:
JIRA api allows to pull all issues at once, but in our case we want to get subsequent results for two reasons:
     - some users were reporing problems gathering more then 100 results in one query.
     - during collection we present feedback to the user that operation is ongoing.
    
As part of this change also the requested data was limited to the data required by pelorus failure exporters, which is summary,labels,created,resolutiondate

Fixes #304 

@redhat-cop/mdt


KevinMGranger:
I think this should have an accompanying unit and/or integration test, right?

PR #429, opened 2022-04-26
Fix for issue #428 and some minor Makefile nits

### LINKED ISSUE #428: Split integration tests from macoon ones

mpryc:
Split of mockoon tests from integration tests, which allows to run e.g. failure integration tests easilly from makefile.
This requires creation of new GitHub workflow to run the Mockoon tests separatelly.

Updated README.md to reflect current set of scripts in the scripts/ directory.

Added missing shell scripts in the Makefile, that required small nit change in the install_dev_tools script.

resolves #428

@redhat-cop/mdt

### DIFF
--- a/.github/workflows/mockoon-tests.yml
+++ b/.github/workflows/mockoon-tests.yml
@@ -1,17 +1,19 @@
-name: Integration Tests
+name: Mockoon Tests
 on:
   push:
     paths:
       - 'exporters/**'
-      - 'pyproject.toml'
-      - '.github/workflows/integration-tests.yml'
+      - 'mocks/**'
+      - '!mocks/README.md'
+      - '.github/workflows/mockoon-tests.yml'
       - 'Makefile'
 
   pull_request:
     paths:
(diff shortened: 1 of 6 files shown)


scripts/run-mackoon-tests
```diff
@@ -106,7 +106,7 @@ done
 # Remove exit trap, so we can propagate exit value from pytest
 trap - EXIT
 
-# Run integration tests
-pytest -m integration
+# Run mackoon tests
+pytest -m mackoon
```
KevinMGranger, line 110: We should probably use `-rap` here too, for consistent output.

KevinMGranger (changes requested):
Typo: it's mockoon, not mackoon. Named after the mocking it's doing (and a raccoon).

scripts/run-mackoon-tests
mpryc: Fixed. Also changed all mackoon to mockoon filenames and strings in source files.

PR #431, opened 2022-04-26
Fix issue #430 to allow running Mockoon container without SELinux

### LINKED ISSUE #430: Mockoon container is failing to run with SELinux enabled
There are couple of options here to fix it, either add to the podman run command:

`--security-opt label=disable
`
or similar the **:Z** while mouting the volume:

Yet another option is to run, however we don't really want to change security context here:

`chcon -Rt svirt_sandbox_file_t /var/data
`

mpryc:
Fix issue #430 to allow running Mockoon container without SELinux

SELinux causes EACCES: permission denied error to access /data folder in a container when it's not disabled. For running tests it's ok to run it with selinux disabled.

resolves #430

@redhat-cop/mdt

### DIFF
--- a/scripts/run-mockoon-tests
+++ b/scripts/run-mockoon-tests
@@ -83,7 +83,7 @@ fi
 
 # Run Mockoon container
 echo "Running mockoon container"
-RUNNING_ID=$("${CONTAINER_ENGINE}" run --name="${TEST_CONTAINER_NAME}" -d -v "${MOCK_JSON}":/data -p "${MOCKOON_PORT}":"${MOCKOON_PORT}" "${MOCKOON_CONTAINER}" -d /data)
+RUNNING_ID=$("${CONTAINER_ENGINE}" run --name="${TEST_CONTAINER_NAME}" -d -v "${MOCK_JSON}":/data:Z -p "${MOCKOON_PORT}":"${MOCKOON_PORT}" "${MOCKOON_CONTAINER}" -d /data)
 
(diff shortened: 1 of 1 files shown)


scripts/run-mackoon-tests
```diff
@@ -83,7 +83,7 @@ fi
 
 # Run Mockoon container
 echo "Running mockoon container"
-RUNNING_ID=$("${CONTAINER_ENGINE}" run --name="${TEST_CONTAINER_NAME}" -d -v "${MOCK_JSON}":/data -p "${MOCKOON_PORT}":"${MOCKOON_PORT}" "${MOCKOON_CONTAINER}" -d /data)
+RUNNING_ID=$("${CONTAINER_ENGINE}" run --name="${TEST_CONTAINER_NAME}" -d -v "${MOCK_JSON}":/data --security-opt label=disable -p "${MOCKOON_PORT}":"${MOCKOON_PORT}" "${MOCKOON_CONTAINER}" -d /data)
```
KevinMGranger, line 86: Can we use your `:Z` idea instead? It means we get to keep label enforcement. I just tried it on my fedora machine and it worked.

KevinMGranger (changes requested):

scripts/run-mackoon-tests
mpryc: yep there were 3 methods to fix it as described in the #430, changed to :Z

KevinMGranger (approved):

PR #441, opened 2022-04-27
Make exporters log version information on startup

### LINKED ISSUE #379: Each exporter should log its version upon startup
This will make debugging others' deployments easier.

KevinMGranger:
Logs version information at startup, obtained from environment variables set by a source to image build.

## Linked Issues?

resolves #379

## Testing Instructions

1. Deploy an exporter from `source_url: https://github.com/kevinmgranger/pelorus.git`, `source_ref: log_version_at_startup`
2. Look at the top of the logs for a pod from that deployment
3. See the logged repo, branch / ref, and commit information

### DIFF
--- a/exporters/committime/app.py
+++ b/exporters/committime/app.py
@@ -43,6 +43,7 @@ class GitFactory:
 
 
 if __name__ == "__main__":
+    pelorus.print_version("commit time")
     pelorus.upgrade_legacy_vars()
 
     pelorus.load_kube_config()
--- a/exporters/committime/app.py
+++ b/exporters/committime/app.py
@@ -43,7 +43,6 @@ class GitFactory:
 
 
 if __name__ == "__main__":
-    pelorus.print_version("commit time")
     pelorus.upgrade_legacy_vars()
 
     pelorus.load_kube_config()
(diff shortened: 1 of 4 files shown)


mpryc:
I wonder if we can do something that does not require giving explicit name of the exporter, e.g. in the pelorus __init__.py have print_version() without arguments, this will give us path to the exporter being run.

```
def print_version():
    import __main__
    collector_type = str(__main__.__file__)

```

This could be included in the __setup_logging(), so we don't need to call this function from every individual exporter.


Secondly it's print(), why not use log ?

mpryc:
I wonder if we can do something that does not require giving explicit name of the exporter, e.g. in the pelorus __init__.py have print_version() without arguments, this will give us path to the exporter being run.

```
def print_version():
    import __main__
    collector_type = str(__main__.__file__)

```

This could be included in the __setup_logging(), so we don't need to call this function from every individual exporter.


Secondly it's print(), why not use log ?

KevinMGranger:
I didn't know about `import __main__`! That is much better. I also agree with putting it in the logging setup.

As for print, that's because this should be printed regardless of how logging is configured.

mpryc (approved):

weshayutin (approved):

Issue #436, opened 2022-04-26
Exporters should log the missing label metadata key 


weshayutin:
To give users a better idea of what they are missing in their deployment configs to work well with Pelorus we should log the missing key per app.

For example:
warnings.warn app: foo found, but missing the following label: app.kubernetes.io/name

PR #444, opened 2022-04-29
Log resources that are missing app labels

KevinMGranger:

Logs warnings for resources that are missing the required app label.

Currently only handles deploytime resources (pods and replicators).

Remaining work / open questions:

- Do we envision gradual adoption of Pelorus, and is this presenting the information too much like a firehose?
  - Was the current design chosen in order to make pelorus opt-in?
  - If so, perhaps we could have a separate tool to check for issues?
- [ ] We already only ask for replicators with the label. The check is currently pointless.
  - Should we be logging when an ownerReference doesn't have the required label, then?
  - Wouldn't the value always be the same as the pod anyway?
  - Why don't we look up the replicators lazily? Why all at the same time?
- Is this too expensive to do?
  - For deploytime, logging per replicator is relatively low-cost.
  - For deploytime, logging per _pod_ that doesn't have the label should be fine since it only tracks running pods.
  - For committime, we need to look at _every_ build. That might be a bit much. Right now this PR only has it done for deploytime.
- [ ] This will generate a lot of logs for well-used clusters / namespaces.
We should use a child logger for this so these warnings can be silenced.
- Do we want this for the committime exporter as well? See above.

## Linked Issues?

related to #436 (if we sufficiently handle these cases in every collector, it will close it)

## Testing Instructions

For deploytime:
1. Create pods that are missing the configured APP_LABEL (or, just set the APP_LABEL to something else)
2. Validate that pods without the label are logged.

### DIFF
--- a/exporters/deploytime/app.py
+++ b/exporters/deploytime/app.py
@@ -28,6 +28,7 @@ class DeployTimeCollector:
         self.prod_label = prod_label
         self._warned_missing_replicas = set()
         self.app_label = pelorus.get_app_label()
+        self._warned_missing_pods = set()
 
     def collect(self) -> Iterable[GaugeMetricFamily]:
         logging.info("collect: start")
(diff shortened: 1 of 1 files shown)


KevinMGranger:
We have decided against this approach. The rationale will be documented in the related issue.

KevinMGranger:
Relevant: https://github.com/konveyor/pelorus/discussions/511

Issue #448, opened 2022-05-11
Handle shell variables in consistent way


mpryc:
Variable can be set in SHELL as empty string, e.g. export MY_VAR=""

Then our exporters are using two different methods:

-   When the variable is empty string, use default from python
-   When the variable is empty string use empty string (in which some code is failing)

We should make it more consistent.

mpryc:
Fixed in #449

PR #449, opened 2022-05-11
Fix issue #448 to make handling of shell env vars consistently

mpryc:
Fix issue #448 to make handling of shell env vars consistently
    
Proposed change makes handling of shell env variables in a consstent way and moves the logic to one function, which may be easily adopted in the future development.

@redhat-cop/mdt

### DIFF
--- a/exporters/pelorus/__init__.py
+++ b/exporters/pelorus/__init__.py
@@ -33,7 +33,7 @@ def _print_version():
     exporter_name = pathlib.PurePath(__main__.__file__).parent.name
 
     repo, ref, commit = (
-        os.environ.get(f"OPENSHIFT_BUILD_{var.upper()}")
+        utils.get_env_var(f"OPENSHIFT_BUILD_{var.upper()}")
         for var in "source reference commit".split()
     )
     if repo and ref and commit:
@@ -47,7 +47,7 @@ def _print_version():
 # region: logging setup
 def _setup_logging():
     _print_version()
-    loglevel = os.getenv("LOG_LEVEL", DEFAULT_LOG_LEVEL).upper()
+    loglevel = utils.get_env_var("LOG_LEVEL", DEFAULT_LOG_LEVEL).upper()
     numeric_level = getattr(logging, loglevel, None)
     if not isinstance(numeric_level, int):
         raise ValueError("Invalid log level: %s", loglevel)
@@ -73,7 +73,7 @@ NamespaceSpec = Optional[Sequence[str]]
 
 
 def load_kube_config():
-    if "OPENSHIFT_BUILD_NAME" in os.environ:
+    if utils.get_env_var("OPENSHIFT_BUILD_NAME") is not None:
         config.load_incluster_config()
         file_namespace = open(
             "/run/secrets/kubernetes.io/serviceaccount/namespace", "r"
@@ -108,17 +108,17 @@ def convert_timestamp_to_date_time_str(timestamp, format_string="%Y-%m-%dT%H:%M:
 
 
 def get_app_label():
-    return os.getenv("APP_LABEL", DEFAULT_APP_LABEL)
+    return utils.get_env_var("APP_LABEL", DEFAULT_APP_LABEL)
 
 
 def get_prod_label():
-    return os.getenv("PROD_LABEL", DEFAULT_PROD_LABEL)
+    return utils.get_env_var("PROD_LABEL", DEFAULT_PROD_LABEL)
 
 
 def missing_configs(vars):
     missing_configs = False
     for var in vars:
-        if var not in os.environ:
+        if utils.get_env_var(var) is None:
             logging.error("Missing required environment variable '%s'." % var)
             missing_configs = True
 
@@ -126,14 +126,14 @@ def missing_configs(vars):
 
 
 def upgrade_legacy_vars():
-    username = os.environ.get("GITHUB_USER")
-    token = os.environ.get("GITHUB_TOKEN")
-    api = os.environ.get("GITHUB_API")
-    if username and not os.getenv("GIT_USER"):
+    username = utils.get_env_var("GITHUB_USER")
+    token = utils.get_env_var("GITHUB_TOKEN")
+    api = utils.get_env_var("GITHUB_API")
+    if username and not utils.get_env_var("GIT_USER"):
         os.environ["GIT_USER"] = username
-    if token and not os.getenv("GIT_TOKEN"):
+    if token and not utils.get_env_var("GIT_TOKEN"):
         os.environ["GIT_TOKEN"] = token
-    if api and not os.getenv("GIT_API"):
+    if api and not utils.get_env_var("GIT_API"):
         os.environ["GIT_API"] = api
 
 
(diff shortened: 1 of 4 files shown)


exporters/pelorus/__init__.py
```diff
@@ -33,7 +33,7 @@ def _print_version():
     exporter_name = pathlib.PurePath(__main__.__file__).parent.name
 
     repo, ref, commit = (
-        os.environ.get(f"OPENSHIFT_BUILD_{var.upper()}")
+        utils.get_env_var(f"OPENSHIFT_BUILD_{var.upper()}")
```
weshayutin, line 36: honest question... what is the advantage of utils.get_env_var over os.environ.get?

exporters/pelorus/__init__.py
```diff
@@ -73,7 +75,7 @@ def _setup_logging():
 
 
 def load_kube_config():
-    if "OPENSHIFT_BUILD_NAME" in os.environ:
+    if utils.get_env_var("OPENSHIFT_BUILD_NAME") is not None:
```
weshayutin, line 78: ++ improvement

exporters/pelorus/utils.py
```diff
@@ -142,3 +145,20 @@ def format(self, record):
             return logging.Formatter.format(self, record)
         finally:
             self._style._fmt = prior_format
+
+
+def get_env_var(var_name: str, default_value: Optional[str] = None) -> str:
```
weshayutin, line 150: OH.. I see :)

weshayutin (approved):

exporters/pelorus/__init__.py
mpryc: It's explained in the utils.get_env_var function, there are two reasons:

1. Have consistent fallback option when we will start using config maps, so we don't have case where SOME_VAR="" is considered non empty value and instead of default one empty string is used.
2. Have one function where we will change the behavior if needed across all usages in pelorus.

KevinMGranger (commented):
I just tested some different things with ConfigMaps and `env` in OpenShift, and found some interesting behavior:

The variable will not be set as long as the variable is not set in ALL of the following places:
1. a ConfigMap brought in with `envFrom`
2. an env var in the container spec: `env: [{name: VAR_NAME, value: VAR_VALUE}]`
3. the image itself

YAML's null value can be given in any of the following ways:
1. literal `null`
2. literal `~`
3. having a key in an object with no value

With that in mind...

If an environment variable is in any of the following conditions:
1. In a config map (brought in with `envFrom`), the name is specified but null
2. In the `Container`'s `env` array as any of the following:
  1. `{name: VAR_NAME, value: null}`
  2. `{name: VAR_NAME, value:}`
  3. `{name: VAR_NAME}`

Then it _will still be set_, but to an empty string.

In other words, if we leave the keys present but blank in the configmap approach, it will _always_ be an empty string unless specified. We have no way of telling if a literal empty string was deliberately given.

I propose that we treat absence/empty string to both mean `default`. This will let us have "default" as a meaningful value (maybe someone has a project named default? Who knows!)

It also simplifies the `get_env_var` implementation:
```python
env_var = os.getenv(var_name)
if not env_var and default_value is None:
    raise ValueError(...)
return env_var or default_value
```

exporters/pelorus/__init__.py
```diff
@@ -47,7 +47,9 @@ def _print_version():
 # region: logging setup
 def _setup_logging():
     _print_version()
-    loglevel = os.getenv("LOG_LEVEL", DEFAULT_LOG_LEVEL).upper()
+    loglevel = utils.get_env_var("LOG_LEVEL", DEFAULT_LOG_LEVEL)
+    if isinstance(loglevel, str):
+        loglevel = loglevel.upper()
```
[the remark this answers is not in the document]
mpryc: In case where LOG_LEVEL is missing from the env, then the return value is None, which isn't a string, so it fails:

`AttributeError: 'NoneType' object has no attribute 'upper'
`

exporters/pelorus/utils.py
```diff
@@ -142,3 +145,20 @@ def format(self, record):
...
+    """
+    `get_env_var` modifies standard os.getenv behavior to allow using default python variable values
+    when DEFAULT_VAR_KEYWORD keyword is present in the SHELL env variable. This is required for the
+    config map to define fallback vars in a consistent way.
+    """
+    env_var = os.getenv(var_name, default_value)
+    if env_var == DEFAULT_VAR_KEYWORD:
+        if default_value is None:
+            raise ValueError(
+                f"{'default value not present for SHELL env var: ' + var_name}"
```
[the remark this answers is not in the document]
mpryc: resolved in latest review

exporters/pelorus/utils.py
```diff
@@ -6,11 +6,14 @@
 import contextlib
 import dataclasses
 import logging
+import os
 from typing import Any, Optional, Union, overload
 
 # sentinel value for the default kwarg to get_nested
 __GET_NESTED_NO_DEFAULT = object()
 
+DEFAULT_VAR_KEYWORD = "default"
```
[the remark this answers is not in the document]
mpryc: @KevinMGranger I agree about this corner case and we should not disregard possibility of having "default" as valid PROJECT name or other env variable.

That is why I have introduced option to override "default" with any user specified keyword. This way user will have option to configure config map to use other then "default" keyword:
`
PELORUS_DEFAULT_KEYWORD
`

exporters/pelorus/__init__.py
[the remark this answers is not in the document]
KevinMGranger: But we give it a default value right above, don't we?

exporters/pelorus/utils.py
KevinMGranger: The return type should be `Optional[str]`. That can happen if `default_value` is `None` and the variable is unset.

exporters/pelorus/utils.py
```diff
@@ -142,3 +145,36 @@ def format(self, record):
...
+           export LOG_LEVEL="default"
+
+           In which case LOG_LEVEL is set to DEFAULT_LOG_LEVEL
+
+    This is required for the config map to define fallback vars in a consistent way.
+    """
+    default_keyword = DEFAULT_VAR_KEYWORD
+    custom_default_keyword = os.getenv("PELORUS_DEFAULT_KEYWORD")
+    if custom_default_keyword and len(custom_default_keyword) > 0:
+        default_keyword = custom_default_keyword
```
KevinMGranger, line 172: Since `None` and `""` are both falsy, this section can be simplified to `default_keyword = os.getenv("PELORUS_DEFAULT_KEYWORD") or DEFAULT_VAR_KEYWORD`

KevinMGranger (commented):
@mpryc any thoughts on the empty string thing we saw before? Does that mean we change the default configmap to just say "default" for each value?

At this point, all the default / empty string handling has me saying "let's just use a config file" 😆

exporters/pelorus/utils.py
mpryc: changed.

mpryc: Yes I will change the config map PR later to include "default" when we expect default value.

KevinMGranger:
Just made two small additions:

1. Added typing overloads for get_env_var. Now the type checkers knows it's guaranteed to be a string if you give a default.
2. Added a decision table to document our expectations.
3. 
If these look good to you, I'll hit approve!

mpryc:
@KevinMGranger I like the additional changes, so go ahead and merge if you are satisfied with the PR.

Issue #131, opened 2020-07-13
Document requirements for installing in an disconnected environment


prakritiverma:

mpryc:
Note: Use case is valid, to install Pelorus in the clusters that do not have access to the resources such as quay.io nor github.

PR #459, opened 2022-05-17
Airgap possiblities

ColonelBundy:
* Added airgap possiblities.

## Linked Issues?

None

## Testing Instructions

Override values in values.yaml.

### DIFF
--- a/charts/pelorus/values.yaml
+++ b/charts/pelorus/values.yaml
@@ -27,3 +27,12 @@ exporters:
     source_url: https://github.com/konveyor/pelorus.git
 
 snapshot_schedule: "@monthly"
+
+
+# Specify images and their versions
+# override for airgap possiblities.
+images:
+  oauth_proxy: image-registry.openshift-image-registry.svc.cluster.local:5000/openshift/oauth-proxy:v4.4
+  thanos: quay.io/thanos/thanos:v0.8.1
+  thanos_proxy: registry.redhat.io/openshift3/oauth-proxy:v3.11
+  web_terminal_tooling: quay.io/wto/web-terminal-tooling:1.2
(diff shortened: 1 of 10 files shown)


charts/pelorus/values.yaml
```diff
@@ -27,3 +27,12 @@ exporters:
...
 snapshot_schedule: "@monthly"
+
+
+# Specify images and their versions
+# override for airgap possiblities.
+images:
+  oauth_proxy: image-registry.openshift-image-registry.svc.cluster.local:5000/openshift/oauth-proxy:v4.4
+  thanos: quay.io/thanos/thanos:v0.8.1
+  thanos_proxy: registry.redhat.io/openshift3/oauth-proxy:v3.11
+  web_terminal_tooling: quay.io/wto/web-terminal-tooling:1.2
```
KevinMGranger, line 38: If these become mandatory values in `values.yaml`, then this becomes a breaking change. Right now, the configuration workflow is editing the `values.yaml` file in place instead of overriding it with various helm flags. We also want a minimalist `values.yaml` to still be valid.

To solve those problems, we should have these as defaults within the chart itself. In other words, `image: {{ .Values.images.thanos | default "quay.io/thanos/thanos:v0.8.1" }}`

KevinMGranger (changes requested):
Hi @ColonelBundy, we're looking into this, and grateful for your contribution! There's just a couple of things we should address:

- We want the defaults to live in the charts. This will be a better UX for the values file.
- We'll need to update the chart version(s). If you haven't already, you can set up the [development environment](https://pelorus.readthedocs.io/en/stable/Development/#updating-the-chart-versions) to do this automatically (and have it run a git commit hook to verify you've done this!)

mpryc:
PR triage:
 Still valid PR, need rework to adopt to the new pelorus-operator installation method.

mpryc:
Note: will this address https://github.com/konveyor/pelorus/issues/131 ?

Issue #262, opened 2021-03-16
Update code to support OpenShift Python client >= .12


themoosman:
The current code base will only work with the [openshift python client](https://github.com/openshift/openshift-restclient-python) of version <= .11.  The current version is .12.

mpryc:
Fixed with #460

Issue #446, opened 2022-05-09
committime exporters does not register app w/o helm upgrade


weshayutin:
1. walk through the steps outlined here
https://github.com/konveyor/pelorus/blob/ea770709f0ed8580d32e2524bd1c8fd324b3e95f/samples/todolist.md
2. Note that the committime exporter does not register the todolist-mongo app, but the deploytime exporters does.

committime:
```
# TYPE commit_timestamp gauge
commit_timestamp{app="committime-exporter",commit="16a0fc499de54eaf2fa03510f02268de6e73dee2",image_sha="sha256:3ddb964b3fcd09cca3907b7759510371406a14f06019e3c5238ad5f61e74a9ca",namespace="pelorus"} 1.651686863e+09
# HELP commit_timestamp Commit timestamp
# TYPE commit_timestamp gauge
commit_timestamp{app="committime-exporter",commit="16a0fc499de54eaf2fa03510f02268de6e73dee2",image_sha="sha256:3ddb964b3fcd09cca3907b7759510371406a14f06019e3c5238ad5f61e74a9ca",namespace="pelorus"} 1.651686863e+09
commit_timestamp{app="deploytime-exporter",commit="16a0fc499de54eaf2fa03510f02268de6e73dee2",image_sha="sha256:fc85f7a5563d505348878b6b9549f698db6061ff7d8b7f3a5b73d76ab10ead17",namespace="pelorus"} 1.651686863e+09
```
deploytime
```
# HELP deploy_timestamp Deployment timestamp
# TYPE deploy_timestamp gauge
deploy_timestamp{app="todolist",image_sha="sha256:1e9be94df5129c3b32ee97e35b0871823d4789f8942b8799d25237d5f7c65a4f",namespace="mongo-persistent"} 1.652118287e+09
```

Now uninstall pelorus
helm uninstall pelorus  --namespace pelorus 

Reinstall
 helm install pelorus charts/pelorus --namespace pelorus --values /var/tmp/PELORUS/my_sample_values.yaml

* now the todolist-mongo in mongo-persistent is found by committime.
```
# HELP commit_timestamp Commit timestamp
# TYPE commit_timestamp gauge
commit_timestamp{app="todolist",commit="0d5e0986628b72ff0c694c23e428eceeb9027e49",image_sha="sha256:1e9be94df5129c3b32ee97e35b0871823d4789f8942b8799d25237d5f7c65a4f",namespace="mongo-persistent"} 1.651758549e+09
[...]

PR #460, opened 2022-05-17
Use new python openshift version 0.13.1

mpryc:
Fixes Issue #262, which uses newer version of OpenShift python module.

related to #446 

@redhat-cop/mdt

### DIFF
--- a/exporters/committime/app.py
+++ b/exporters/committime/app.py
@@ -4,8 +4,6 @@ import os
 import time
 from distutils.util import strtobool
 
-from kubernetes import client
-from openshift.dynamic import DynamicClient
 from prometheus_client import start_http_server
 from prometheus_client.core import REGISTRY
 
@@ -45,10 +43,7 @@ class GitFactory:
 if __name__ == "__main__":
     pelorus.upgrade_legacy_vars()
 
-    pelorus.load_kube_config()
-    k8s_config = client.Configuration()
-    k8s_client = client.api_client.ApiClient(configuration=k8s_config)
-    dyn_client = DynamicClient(k8s_client)
+    dyn_client = pelorus.utils.get_k8s_client()
 
     username = os.environ.get("GIT_USER", "")
     token = os.environ.get("GIT_TOKEN", "")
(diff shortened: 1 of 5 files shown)


weshayutin (approved):
VERY NICE! WFM :)

exporters/pelorus/utils.py
```diff
@@ -198,3 +201,25 @@ def get_env_var(var_name: str, default_value: Optional[str] = None) -> Optional[
...
+        k8s_client = DynamicClient(k8sconfig)
+    except config.config_exception.ConfigException:
+        # Try load config from cluster
+        config.load_incluster_config()
+        file_namespace = open(
+            "/run/secrets/kubernetes.io/serviceaccount/namespace", "r"
+        )
+        if file_namespace.mode == "r":
+            namespace = file_namespace.read()
+            print("namespace: %s\n" % (namespace))
```
KevinMGranger, line 219: I'm wondering why this was here in the first place-- just printing what namespace the service account is in? Can't that just be checked in OpenShift directly?

If we keep this, then it looks like the old code had some issues:

1. It's checking that the file's mode is `r`, when it just opened it with `r`.
2. The file never gets closed.

We can just replace this whole part with `namespace = Path("/run/secrets/kubernetes.io/serviceaccount/namespace").read_text()`

exporters/requirements.txt
```diff
@@ -1,5 +1,5 @@
 # Used by all exporters
-openshift < 0.12       # module openshift
+openshift == 0.13.1    # module openshift
```
KevinMGranger, line 2: 🎉

exporters/committime/app.py
```diff
@@ -45,10 +43,7 @@ def getCollector(
 if __name__ == "__main__":
     pelorus.upgrade_legacy_vars()
 
-    pelorus.load_kube_config()
-    k8s_config = client.Configuration()
-    k8s_client = client.api_client.ApiClient(configuration=k8s_config)
-    dyn_client = DynamicClient(k8s_client)
+    dyn_client = pelorus.utils.get_k8s_client()
```
KevinMGranger, line 46: I like this, it really cleans things up!

mpryc:
Merging, got approval from @weshayutin comment from @KevinMGranger was resolved and 4.9 prow is a known issue not related to this change.

PR #467, opened 2022-05-25
Add readiness and liveness probes, associate exporter type label

mpryc:
This will allow to check exporter pod statuses in easy way.

@redhat-cop/mdt

### DIFF
--- a/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
+++ b/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
@@ -22,7 +22,8 @@ spec:
         app.kubernetes.io/name: {{ .app_name }}
         app: {{ .app_name }}
         deploymentconfig: {{ .app_name }}
-        application: {{ .app_name}}
+        application: {{ .app_name }}
+        pelorus.konveyor.io/exporter-type: {{ .exporter_type | default "generic-exporter" }}
     spec:
       containers:
       - name: {{ .app_name }}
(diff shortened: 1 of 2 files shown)


charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
```diff
@@ -51,6 +52,16 @@ spec:
...
+        readinessProbe:
+          tcpSocket:
+            port: 8081
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        livenessProbe:
+          tcpSocket:
+            port: 8080
+          initialDelaySeconds: 15
+          periodSeconds: 20
```
weshayutin, line 64: would threshholds be useful here?
successThreshold: 1 
failureThreshold: 3 
https://docs.openshift.com/container-platform/4.10/applications/application-health.html

mpryc: By default it's 1 and 3, so I think we are fine here. If exporter is not exposing 8080 on localhost, then we are in a trouble.

There is benefit of having those probes as this will kill pods and remove from the cluster to not occupy resources. There is also a drawback. Once the pod is killed it's gone and we can't simply get the logs from such pod to see what happened within python application, however we can always restart deployment and while the pod is running investigate the issue.

charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
```diff
@@ -22,7 +22,8 @@ spec:
         app.kubernetes.io/name: {{ .app_name }}
         app: {{ .app_name }}
         deploymentconfig: {{ .app_name }}
-        application: {{ .app_name}}
+        application: {{ .app_name }}
+        type: {{ .exporter_type | default "generic-exporter" }}
```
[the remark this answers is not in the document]
mpryc: @KevinMGranger nice idea, changed in the latest version.

KevinMGranger (approved):
/lgtm

PR #485, opened 2022-06-06
Make default values.yaml use default configmaps

KevinMGranger:

Lets a default install work without setting up configmaps first.

## Testing Instructions

1. Uninstall pelorus and any configmaps referenced by it.
2. Install pelorus using the default values.yml in this branch.
3. See if the deploytime exporter deploys correctly.

### DIFF
--- a/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
+++ b/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
@@ -29,17 +29,13 @@ spec:
       - name: {{ .app_name }}
         imagePullPolicy: Always
         envFrom:
-      {{- if .env_from_configmaps }}
-        {{- range .env_from_configmaps }}
+      {{- range .env_from_configmaps | default (list "default-pelorus-config" "default-deploytime-config") }}
         - configMapRef:
             name: {{ . }}
-        {{- end}}
       {{- end}}
-      {{- if .env_from_secrets }}
-        {{- range .env_from_secrets }}
+      {{- range .env_from_secrets }}
         - secretRef:
             name: {{ . }}
-        {{- end}}
       {{- end}}
         env:
 {{- if .exporter_type }}
--- a/charts/pelorus/charts/exporters/templates/default_configmaps.yaml
+++ b/charts/pelorus/charts/exporters/templates/default_configmaps.yaml
@@ -1,0 +1,21 @@
+# These configmaps are not meant to be edited directly.
+# To configure pelorus, you can create your own copies of these.
+# See the documentation for details:
+# https://pelorus.readthedocs.io/en/latest/Configuration/#configuring-exporters
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: default-pelorus-config
+data:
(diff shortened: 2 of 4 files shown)


charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
```diff
@@ -29,17 +29,13 @@ spec:
...
-      {{- if .env_from_configmaps }}
-        {{- range .env_from_configmaps }}
+      {{- range .env_from_configmaps | default (list "default-pelorus-config" "default-deploytime-config") }}
         - configMapRef:
             name: {{ . }}
-        {{- end}}
       {{- end}}
-      {{- if .env_from_secrets }}
-        {{- range .env_from_secrets }}
+      {{- range .env_from_secrets }}
```
KevinMGranger, line 36: Go Templates inherit Go's semantics for arrays: the "zero" / nil value is considered an empty array.

charts/pelorus/charts/exporters/templates/default_configmaps.yaml
```diff
@@ -0,0 +1,21 @@
+# These configmaps are not meant to be edited directly.
+# To configure pelorus, you can create your own copies of these.
+# See the documentation for details:
+# https://pelorus.readthedocs.io/en/latest/Configuration/#configuring-exporters
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: default-pelorus-config
```
KevinMGranger, line 9: I'm not married to these names, of course :)

weshayutin:
deploytime exporter 

---> Running application from Python script (deploytime/app.py) ...
  File "/opt/app-root/src/deploytime/app.py", line 14, in <module>
    import pelorus
  File "/opt/app-root/lib64/python3.9/site-packages/pelorus/__init__.py", line 23, in <module>
    raise ValueError("Invalid log level: %s" % loglevel)
ValueError: Invalid log level: default

weshayutin:
I like where this is going.  Is the more appropriate default for LOG_LEVEL = "INFO" ?
Also let's:
remove line 158 [here](https://github.com/konveyor/pelorus/blob/master/scripts/run-pelorus-e2e-tests#L158)
add "--values charts/pelorus/values.yaml" to line 166 [here](https://github.com/konveyor/pelorus/blob/master/scripts/run-pelorus-e2e-tests#L166)  /me notes that may be default and unrequired, I forget

mpryc:
@weshayutin the issue with the exception on the default log level, is because we did not yet create release and you are using default ref where code is not compatible with configmaps:

https://github.com/konveyor/pelorus/blob/master/charts/pelorus/charts/exporters/templates/_buildconfig.yaml#L20

To fix this you need to specify in your deployment config master branch for the exporter instance.

mpryc:
@KevinMGranger I like your approach. Question, can the default configmaps be a symlink to already defined ones in the configmap directory? It's kind of duplication. Also I am unsure if having one file with two configmaps is good practice, we can have multiple instances with different configmaps. I understand it's the "default" one, so maybe it's ok to have them in one ?

PR #495, opened 2022-06-09
Allow get_nested function to accept dotted paths

mpryc:
Enhance get_nested function to allow obtaining values from the nested paths that have dotted keys, e.g.:

`metadata.annotations.io.openshift.version = 4.9`

In such case we are unable to get the value from annotation:

`io.openshift.version`

## Testing Instructions

make unit-tests

@redhat-cop/mdt


KevinMGranger:
This is what the list form of `get_nested` is for. For example:

```python
get_nested(build, ["metadata", "annotations", "io.openshift.build.commit.id"], default=None)
```

PR #497, opened 2022-06-09
ensure deploytime timestamp is set to the deploytime

### LINKED ISSUE #245: Metric Deployment Frequency is not correct in the selected time range
Yesterday (February 1) I built a lab in opentlc (ocp 4.6) and installed pelorus (with helm install ...). and deploys example applications. I deployed example it for the last time at approximately 10pm. Today (February 2), when I select the range "Last 5 minutes" and without having done any deploy, I see the following image:

Pelorus Dashboard

is it correct?

weshayutin:
## Linked Issues?

resolves #245

## Testing Instructions

@redhat-cop/mdt


KevinMGranger:
Need to do some quick local testing first but this looks about right.

Issue #474, opened 2022-05-31
Tooling to validate the install of Pelorus


weshayutin:

weshayutin:
A first pass of this is done. 
@mpryc is rewriting the validation and incorporating it into the Pelorus Makefile
https://github.com/konveyor/pelorus/pull/472

mpryc:
@weshayutin Can we consider this done?

I think the tooling is in a good shape (scripts/run-pelorus-e2e-tests) as we have:
 - generic way of validating install and exporters
 - possibility to use custom deployment types by passing custom values.yaml

The use of the tooling and coverage such as additional JIRA exporter will be captured in #473.

weshayutin:
yes.. I think so :)
I'm still working https://github.com/openshift/release/pull/29693 but that is out of scope of this issue

PR #504, opened 2022-06-13
Allow to use custom filename in the CI deployment file for the e2e tests

mpryc:
Passing different filename to the run-pelorus-e2e-tests allows to create multiple different CI jobse based on the same deployment script.

This is just a base implementation, each job should have corresponding change in the Makefile to use this feature.

This is needed to allow implement #474 in a nice fashion.

@redhat-cop/mdt


weshayutin (approved):
/LGTM

PR #507, opened 2022-06-13
Nit change to inform that exporter is ran from image.

mpryc:
When the exporter is ran from the pre-built image, no OPENSHIFT_BUILT_* env variables are available.
There is also no information accessible from the image itself about the source code from which the image was built.

This change allows to include image name and tag, so we can print this information within exporter.

We also have knowledge about the image sha that we can easily identify the source code for it using standard oc describe commands:

To obtain information about the image, simple use:
```
oc describe pod -n pelorus <exporter_name>

oc describe pod -n pelorus deploytime-exporter
[...]
>     Image:          image-registry.openshift-image-registry.svc:5000/pelorus/deploytime-exporter@sha256:2574fb8ea74e7ceca6033a0bac75b3ca1581c34fc06c6b79bc755cbd9a6e6d15

```

@redhat-cop/mdt

### DIFF
--- a/charts/pelorus/Chart.yaml
+++ b/charts/pelorus/Chart.yaml
@@ -14,7 +14,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.6.1
+version: 1.6.2
 
 dependencies:
   - name: exporters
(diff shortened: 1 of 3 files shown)


KevinMGranger:
It's a little tricky to test this locally, but it looks good to me.



I tested this by setting OPENSHIFT_BUILD_COMMIT to an empty string, and manually setting PELORUS_IMAGE_TAG to some random string.

It was too tricky to create a proper build that wouldn't include the s2i info. It looks good enough to me!

Maybe we need a script to run a local image build?

PR #509, opened 2022-06-14
Proper exit code from timeout function

### LINKED ISSUE #508: e2e CI jobs do silently fails in the timeout func
When the e2e tests are executed they may silently fail due to the timeout being ran in a subshell and not propagating exit code properly. This causes the job to pass even if the exporters did fail to be deployed properly.

mpryc:
Fixes #508, where script was failing silently with exit code 0, causing CI jobs to report success, where it should fail.

@redhat-cop/mdt


weshayutin (approved):
thanks!

KevinMGranger:
Well, CI succeeded! :)

PR #513, opened 2022-06-21
Report resource timestamp as metric timestamp

### LINKED ISSUE #532: failure metrics are not set to their timestamps
Similar to #245, #497, but for the failure exporters.

### LINKED ISSUE #533: committime metrics are not set to their timestamps
Similar to #245, #497, but for the committime exporters.

KevinMGranger:
Wes fixed the deploytime issue with this in #497. This just does it for committime and failure.

~~Still need to double check what _type_ it is, as we can't guarantee that the prometheus library will handle non-floats correctly.~~ Wes tried it and said it was good. We can revisit this if there's some type weirdness somewhere.

resolves #532 
resolves #533 

## Testing Instructions

Set up committime and/or failure time exporters, and ensure that the values seen in grafana / prometheus are what you expect.

### DIFF
--- a/exporters/committime/collector_base.py
+++ b/exporters/committime/collector_base.py
@@ -92,6 +92,7 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
                     my_metric.image_hash,
                 ],
                 my_metric.commit_timestamp,
+                timestamp=my_metric.commit_timestamp,
             )
             yield commit_metric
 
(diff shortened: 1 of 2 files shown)


weshayutin:
Understand: WIP because: Still need to double check what type it is, as we can't guarantee that the prometheus library will handle non-floats correctly.

FWIW.. I deployed w/ the change and got the expected results for committime, deploytime, failure exporters.
Thanks!

PR #516, opened 2022-06-22
Move default labels from helm pelorus example values.yaml

mpryc:
Change that moves the default labels that are being used in the thanos-rbac.yaml inside that Thanos ClusterRoleBinding yaml file.

This is nothing more then a bit of cleanup for the values.yaml.

@redhat-cop/mdt


KevinMGranger:
Prow noob here. Are those failing just because the timeout is too aggressive? Or is something else causing an issue?

PR #517, opened 2022-06-22
Show logs from crashing pods for e2e tests

mpryc:
This is useful when e2e are failing, so we know why the pods are in crash state.

## Testing Instructions

Deploy e2e tests with wrong secret...

@redhat-cop/mdt


weshayutin (approved):
/LGTM

PR #522, opened 2022-06-23
Rename GIT_USER and GIT_TOKEN to be consistent across exporters

mpryc:
Change which renames GIT_USER to USER and GIT_TOKEN to TOKEN, however ensures backwards compatibility with the old value names.

This makes all the exporters use same USER and TOKEN value names rather than failure using USER and commit time using GIT_USER.

@redhat-cop/mdt


weshayutin:
WOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOT!!!!!!!!!!!!!!!!!!!!

PR #524, opened 2022-06-24
Update grafana to the newest available version from v4 channel

mpryc:
Grafana operator is installed from the community-operators that is available at:

https://operatorhub.io/operator/grafana-operator

Change to bump the version to the newest available version from the v4 channel which we are using.

@redhat-cop/mdt


weshayutin:
Screenshot from 2022-06-24 08-08-28

PR #534, opened 2022-06-29
Add shellcheck github workflow

### LINKED ISSUE #530: shellcheck not run as part of CI
We should use this to validate the quality of our shell scripts. 

Also, if we don't have it, local devs may encounter git hook errors from others' changes.

mpryc:
Fixes #530

@redhat-cop/mdt

### DIFF
--- a/.github/workflows/shellcheck.yaml
+++ b/.github/workflows/shellcheck.yaml
@@ -1,0 +1,44 @@
+name: Shellcheck Tests
+on:
+  push:
+    # Sync with Makefile and SHELL_SCRIPTS
+    paths:
+      - 'demo/demo-tekton'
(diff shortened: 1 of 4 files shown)


KevinMGranger:
I think this needs an `apt install shellcheck`, since it's not installed by default. Otherwise the makefile will just skip it.

KevinMGranger (changes requested):
I think this needs an apt install shellcheck, since it's not installed by default. Otherwise the makefile will just skip it.

KevinMGranger (approved):
Oh, it worked? Strange, shellcheck isn't in the [release's manifest](https://releases.ubuntu.com/22.04/ubuntu-22.04-live-server-amd64.manifest). Cool!

PR #535, opened 2022-06-30
[docs] Change the prefix for the failure metrics to sdp

### LINKED ISSUE #140: Change the prefix for the failure metrics to `sdp`
For consistency's sake, I would like to see all metrics under the `software delivery performance` group be prefixed by `sdp:`. Right now the three failure rules are prefixed with `fr`, I assume for `failure rate`, which doesn't really apply to all three.

The SDP dashboard queries will need to be updated as well as part of this change.

mpryc:
Change to reflect current implementation in docs.

Fixes #140

This is only doc review as implementation already uses "sdp:".

@redhat-cop/mdt


weshayutin (approved):
++ thank you!

PR #538, opened 2022-07-01
consolidate authentication for exporters in the Configuration.md doc

### LINKED ISSUE #537: Default install instructions imply a jira token is required for the "out of box" exporters, which is not the case.
* Move links to Jira / bitbucket token from Install.md -> Configuration.md

This *may* be confusing to a new user.

weshayutin:
* authentication to remote services is NOT required with the default
  install, so why confuse users in their first steps.
* consolidate the authentication doc in configuration.md

## Linked Issues?

resolves #537 

## Testing Instructions

@redhat-cop/mdt

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -87,6 +87,64 @@ data:
   NAMESPACES: "default"    # ""
 ```
 
+### Authentication to Remote Services
+
+Pelorus exporters make use of `personal access tokens` when authentication is 
+required.  It is recommended to configure the Pelorus exporters with authenticaion
+via the `TOKEN` key to avoid connection rate limiting and access restrictions.
+
+More information about personal access tokens:
+* [Github Personal Access Tokens](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
+
+* [Jira / Bitbucket Personal Access Tokens](https://confluence.atlassian.com/bitbucketserver/personal-access-tokens-939515499.html)
+
+* [Gitlab Personal Access Tokens](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html)
+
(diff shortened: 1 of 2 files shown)


docs/Configuration.md
```diff
@@ -87,6 +87,62 @@ data:
...
+
+* [Gitea Tokens](https://docs.gitea.io/en-us/api-usage/#generating-and-listing-api-tokens)
+
+To store a personal access token securely and to make the token available to Pelorus use
+Openshift secrets. Pelorus can utilize secrets for all of the exporters. The following
+is an example for the committime exporter.
+
+A simple Github example:
+```shell
+oc create secret generic github-secret --from-literal=TOKEN=<personal access token> -n pelorus
```
mpryc, line 113: I do wonder if we should also mention possibility of using example configs to create secrets rather than passing them on the command line only, see:

https://github.com/konveyor/pelorus/tree/master/charts/pelorus/secrets

weshayutin: I think I agree.. we should be pushing people towards the config maps :)  Will update in a bit

mpryc: It's not config map, it's secret which is defined in a file, but similar to config map.

weshayutin: oh ya.. sorry read that too quickly..
cli vs. template.. six of one, half dozen of the other.. worth mentioning though.

docs/Configuration.md
```diff
@@ -87,6 +87,62 @@ data:
...
 ```
 
+### Authentication to Remote Services
+
+Pelorus exporters make use of `personal access tokens` when authentication is 
+required.  It is recommended to configure the Pelorus exporters with authenticaion
+via the `TOKEN` key to avoid connection rate limiting and access restrictions.
+
+More information about personal access tokens:
+* [Github Personal Access Tokens](https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line)
```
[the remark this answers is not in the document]
weshayutin: fixed

docs/Configuration.md
weshayutin: added

PR #541, opened 2022-07-06
Disable helm checksum checking as it's flake in CI

mpryc:
Disabling helm checksum checking may improve pass ratio in the CI.

@redhat-cop/mdt


weshayutin (approved):
/LGTM

PR #556, opened 2022-07-13
Use API_USER instead of USER

### LINKED ISSUE #553: committime: GIT_USER should override USER when both are set
While `$USER` is cleaner for running exporters in a container, where the user running it doesn't matter, it makes for hard-to-spot errors in local testing. Setting `$GIT_USER` does nothing!

(This would be easier to spot with #554.)

In `upgrade_legacy_vars`, instead of
```python
git_user = utils.get_env_var("GIT_USER")
if not utils.get_env_var("USER"):
    os.environ["USER"] = git_user
```
we should have:
```python
git_user = utils.get_env_var("GIT_USER")
user = utils.get_env_var("USER")
os.environ["USER"] = git_user or user
```

KevinMGranger:
resolves #553

### DIFF
--- a/charts/pelorus/Chart.yaml
+++ b/charts/pelorus/Chart.yaml
@@ -14,7 +14,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.6.5
+version: 1.6.6
 
 dependencies:
   - name: exporters
--- a/charts/pelorus/configmaps/committime.yaml
+++ b/charts/pelorus/configmaps/committime.yaml
@@ -7,7 +7,7 @@ metadata:
   name: committime-config
   namespace: pelorus
 data:
-  USER: "default"            # ""  |  User's github username, can be overriden by env_from_secrets
+  API_USER: "default"            # ""  |  User's github username, can be overriden by env_from_secrets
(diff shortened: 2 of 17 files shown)


weshayutin:
I'm finding instances of just "USER" across some of the various docs.
labs/consultant/03-Exporters.md
mocks/README.md
samples/todolist_commit_deploy.md

AFAICT the intention is to globally replace s/USER/API_USER is that correct or no?

KevinMGranger:
Ah, thought I caught them all.

We typically don't manually squash, no? We have github do it for us?

mpryc:
@KevinMGranger one last nit:

docs/Development.md

```
diff --git a/docs/Development.md b/docs/Development.md
index 94b1526..7badb06 100644
--- a/docs/Development.md
+++ b/docs/Development.md
@@ -188,7 +188,7 @@ Code also comes with a nice debugger feature. Here is a starter configuration to
             "program": "${workspaceFolder}/exporters/committime/app.py",
             "console": "integratedTerminal",
             "env": {
-                "GITHUB_USER": "<github username here>",
+                "API_USER": "<github username here>",
                 "GITHUB_TOKEN": "<personal access token here>",
                 "LOG_LEVEL": "INFO",
                 "APP_LABEL": "app.kubernetes.io/name"

```

PR #576, opened 2022-07-21
Make e2e tests use simplified environment

mpryc:
Currently e2e tests are using dev-env, that contains all dependencies even those that are not required to run those tests. Adding separate makefile target allows to save time and space while creating env for running those tests.

This hopefully reduces flake issues in periodic CI jobs.

@redhat-cop/mdt


mpryc:
Depends-On: https://github.com/openshift/release/pull/30682

Issue #592, opened 2022-08-04
Gitea commit time exporter invalid API


mpryc:
While running e2e tests against Gitea provided URL, commit time exporter is giving traceback:

Provider repo URL:
   "https://try.gitea.io/mpryc/pelorus-gitea"
And build hash:
    "897f5c490442e88cc609c241da880128ad02e42b"

```
08-04-2022 12:07:03 ERROR    Error encountered while getting CommitMetric info:
08-04-2022 12:07:03 ERROR    Invalid URL 'try.gitea.io/api/v1/api/v1/repos/mpryc/pelorus-gitea/git/commits/897f5c490442e88cc609c241da880128ad02e42b': No scheme supplied. Perhaps you meant http://try.gitea.io/api/v1/api/v1/repos/mpryc/pelorus-gitea/git/commits/897f5c490442e88cc609c241da880128ad02e42b?
Traceback (most recent call last):
  File "/home/migi/Development/Konveyor/Pelorus/pelorus-minio/exporters/committime/collector_base.py", line 225, in get_metric_from_build
    metric = self._set_commit_timestamp(metric, errors)
  File "/home/migi/Development/Konveyor/Pelorus/pelorus-minio/exporters/committime/collector_base.py", line 359, in _set_commit_timestamp
    metric = self.get_commit_time(metric)
  File "/home/migi/Development/Konveyor/Pelorus/pelorus-minio/exporters/committime/collector_gitea.py", line 64, in get_commit_time
    response = requests.get(url, auth=(self._username, self._token))
  File "/home/migi/Development/Konveyor/Pelorus/pelorus-minio/.venv/lib64/python3.10/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
  File "/home/migi/Development/Konveyor/Pelorus/pelorus-minio/.venv/lib64/python3.10/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/migi/Development/Konveyor/Pelorus/pelorus-minio/.venv/lib64/python3.10/site-packages/requests/sessions.py", line 573, in request
    prep = self.prepare_request(req)
  File "/home/migi/Development/Konveyor/Pelorus/pelorus-minio/.venv/lib64/python3.10/site-packages/requests/sessions.py", line 484, in prepare_request
    p.prepare(
[...]

PR #593, opened 2022-08-04
Fix default FQDN for gitea committime collector.

mpryc:
Fixes issue #592, where gitea commit time default API URL contained wrong default API address resulting in double "api/v1".

Updated docs to match current implementation where GIT_API may be provided for Gitea, Azure DevOps and GitHub backends.

@redhat-cop/mdt


weshayutin (approved):
/LGTM

PR #587, opened 2022-08-02
update Tekton demo with optional build types

weshayutin:
## Linked Issues?

resolves #526 
resolves #478 

related to #<issue number> <-- Use this if it shouldn't

## Testing Instructions

check out pull request
./demo-tekton <git_url>

PR #597, opened 2022-08-08
Clean up tekton demo

KevinMGranger:
Switch to Universal Base Images

Move task to pipeline setup, namespace it

Clarify documentation

Make demo template formatting nicer

---

I had trouble getting #587 working for me. This should fix that. WIP until I have some more clarification on things from #587.


weshayutin (changes requested):

weshayutin:
Thanks for the clean up, everything worked for me.  We do need to keep the resource cleanup to be certain the test is hitting the built container and build numbers line up.

PR #600, opened 2022-08-09
Use correct ns and allow to inspect tkn info in the demo

mpryc:
Properly switch to the basic-python-tekton namespace before performing any oc actions in the demo.

Split the build-and-deploy from the apply and use temporary file for easy debugging.

Co-authored-by: Wesley Hayutin <weshayutin@gmail.com>

@redhat-cop/mdt


weshayutin:
@KevinMGranger need your review here since migi and I are co-authors.  Hold for Kevin's review

PR #604, opened 2022-08-17
Requests custom certs

### LINKED ISSUE #463: Support custom certificates for any exporter that uses requests
The simplest approach is to allow mounting the CA trust file in the container, and then have the environment variable `REQUESTS_CA_BUNDLE` point to that path.

However, we must take care because [prepared requests do not take the environment into account](https://requests.readthedocs.io/en/latest/user/advanced/#:~:text=When%20you%20are,taken%20into%20account.). This is true even if you set the verify path on the session object.

It's also not clear if specifying this _overrides_ or _augments_ the CA bundles provided by `certifi`. That's risky to mix in a project that might reach out to public instances as well as private ones.

One solution that can handle both use cases: detect if a custom CA bundle is specified, and if so, append it to the CA bundle from `certifi` at container startup.

KevinMGranger:
Add the ability to use custom certificates with `requests`-based API clients.

Docs coming in an upcoming PR, or if I finish them before you review and merge.

## Linked Issues?

related to #463

## Testing Instructions

`pytest exporters/tests/certs`


KevinMGranger:
@mpryc @weshayutin ready for a review of the _overall approach_, not so much the functionality. Will be testing it tomorrow, and then documenting it.

KevinMGranger:
I've removed the "resolves #463" because I think docs are really necessary for this.

I also just realized we'll have to integrate volumes into the deploymentconfig now. Trying to think about the best UX for that. You can see my early work in my branch: https://github.com/KevinMGranger/pelorus/tree/requests_custom_certs_docs_and_charts

PR #615, opened 2022-08-29
Troubleshooting script

KevinMGranger:
## Describe the behavior changes introduced in this PR
Add a script to check for mislabeled resources.

Docs will be in a separate PR.

## Linked Issues?

related to #549

## Testing Instructions

1. deploy some builds, pods, and replicationcontrollers. I used the usual todolist-mongo-go.
2. Run the script against `mongo-persistent` with `--for-exporter committime deploytime` to verify that there are no issues.
3. Manually remove the `app.kubernetes.io/name` label from a pod in that namespace.
4. Run the script and verify it complains about the missing label.
5. Add the label back, and then remove that label from the pod's replicationcontroller.
6. Run the script and verify it complains about the missing label.
7. Add the label back, and remove the label from the build that created the image for that pod.
8. Run the script and verify it complains about the missing label.

PR #633, opened 2022-09-14
Make logging set up part of exporter startup, not import side effect

KevinMGranger:
Addressing feedback from https://github.com/konveyor/pelorus/pull/615#pullrequestreview-1107138110 (and solving personal annoyances)

The logging system is now configured as part of exporter startup. Previously it was a side-effect of importing pelorus (bad!)

### DIFF
--- a/exporters/committime/app.py
+++ b/exporters/committime/app.py
@@ -156,6 +156,7 @@ class GitCommittimeConfig:
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


mpryc (changes requested):
I am missing change for the exporters/extra/releasetime exporter. I know it's experimental, but it should still work.

mpryc (approved):
/lgtm

PR #640, opened 2022-09-15
Add option to configure Prometheus retention

mpryc:
Allow to set Prometheus retention to configurable value, which defaults to 1y.

@redhat-cop/mdt


KevinMGranger (approved):
/lgtm

PR #647, opened 2022-09-16
verify both the app and pelorus w/ tekton

weshayutin:
* verify the basic-python app is running
* verify the latest commit deployed is in the committime log
  * update the pipeline permissions w/ route access
* s2i requires committime exporter w/ provider=image
* Update Readme

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -18,11 +18,11 @@ Help:
 ./demo-tekton.sh -h
 ```
 
-|Build Type   |Status         |
-|:------------|:--------------|
-| binary      | supported     |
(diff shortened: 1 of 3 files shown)


weshayutin:
@mpryc @KevinMGranger I think this *may* be helpful when testing locally.  Certainly I've recreated the issue w/ the committime image provider.  I'll retest with this once we've removed timestamps on everything other than deploytime.   I suspect we are still vulnerable to timezone errors when users annotate the commit.date something to revisit on Monday hopefully.

Thanks

mpryc (approved):

PR #649, opened 2022-09-19
Add option to configure Prometheus maximum storage capacity

mpryc:
Option to configure maximum storage capacity for the Prometheus.
This is addition to already configurable time retention.

@redhat-cop/mdt


weshayutin:
Screenshot from 2022-09-19 17-27-00

PR #654, opened 2022-09-20
Adds continous metric to the deploytime

### LINKED ISSUE #657: Deploytime exporter reports too many same rows
If the deploytime monitors multiple namespaces and apps, data can grow really fast.

For each subsequent unique pair of namespace/app there is a new deploy_timestamp with the previous apps already reported.

In case we have 1 pair, there will be 1 deploy_timestamp. If we have 2 pairs, then we will see 1 + (1+1) = 3 rows.
For 3 pairs, we will get 1 + (1+1) + (1+1+1) = 6 rows, meaning we are looking here at sum of an = n+1 numerical sequence.

Attached file with 528 deploy_timestamp, where only 32 should be reported.

For example at 180 apps we would have 16290 lines instead of 180, that would make our deploy time exporter pretty unusable, for 300 apps 45150 lines.
[rows.txt](https://github.com/konveyor/pelorus/files/9608423/rows.txt)

mpryc:
Change that adds to the deploytime continous time series for the active deployments.

Deployment with a timestamp creates an event, which may be in the past. Having just an event we are unable to make calculations based on the currently active deployments, because one may be gone from the cluster.

Adding deployment_active metric to already existing deploy_timestamp will allow to better visualize Lead Time for Change, where the "dots" are connected to each other, plus in the future we may be able to answer additional questions such as how many deployed apps are still working.

On top of that this change ensures one metric is presented only one time in the series rather then multiple times. The metrics were gathered several times multiplied by the number of found unique deployments.

Fixes #657 

@redhat-cop/mdt

### DIFF
--- a/exporters/deploytime/app.py
+++ b/exporters/deploytime/app.py
@@ -26,35 +26,51 @@ class DeployTimeCollector:
 
     def collect(self) -> Iterable[GaugeMetricFamily]:
         logging.info("collect: start")
-        metric = GaugeMetricFamily(
-            "deploy_timestamp",
-            "Deployment timestamp",
-            labels=["namespace", "app", "image_sha"],
-        )
 
         namespaces = self.get_and_log_namespaces()
         if not namespaces:
             return []
 
         metrics = generate_metrics(namespaces, self.client)
-        # TODO: when merging 654, just ignore this whole section (overwrite with 654)
-        # all the code should be self-contained in the DeployTimeMetric anyway.
+
+        deploy_timestamp_metric = GaugeMetricFamily(
+            "deploy_timestamp",
+            "Deployment timestamp",
+            labels=["namespace", "app", "image_sha"],
+        )
+        deployment_active_metric = GaugeMetricFamily(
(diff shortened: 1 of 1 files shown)


KevinMGranger:
We're inheriting the mistake the previous code made-- I'll have a proposed patch to clean that up in a second.

KevinMGranger:
Thoughts? https://github.com/KevinMGranger/pelorus/commit/1dec37aad2cc09631d9acc939449b467dd8f7b6e

Compare: https://gist.github.com/KevinMGranger/78bf9c10becd8d7ad9b5fe267626a68c

Not only are the metric readings now side-by-side, the statistics about python garbage collection differ by an order of magnitude. Do you accept these changes?

weshayutin:
not required for MVR.. let's so a follow up 2.0.1

KevinMGranger:
Looks like a prow / AWS error, right?

mpryc:
@KevinMGranger yeah

KevinMGranger (approved):

PR #658, opened 2022-09-20
ensure CUSTOM_DATE_FORMAT does not crash committime exporter

weshayutin:
fix image-exporter date, and update tekton commit timestamp

The image exporter is crashing when the date format
and commit.date image annotation does not match.

currently the commit time is read in human readable from
git log.  It is safer to use the epoch time format
and set the committime-image-exporter to use epoch
as the COMMIT_DATE_FORMAT

### DIFF
--- a/demo/tekton-demo-setup/03-build-and-deploy.yaml
+++ b/demo/tekton-demo-setup/03-build-and-deploy.yaml
@@ -94,7 +94,8 @@ objects:
             set -ex
             find /workspace
             cd /workspace/debug 
-            git log -1 --format=%cd > "$(results.commit-time.path)"
+            epoch_time=`git log -1 --format=%ct`
+            TZ=UTC date -d @`echo $epoch_time` | tee "$(results.commit-time.path)"
         - name: debug-time-last-commit
           image: ubi9/toolbox
           script: |
--- a/exporters/committime/collector_image.py
+++ b/exporters/committime/collector_image.py
@@ -120,10 +120,18 @@ class ImageCommitCollector(AbstractCommitCollector):
     def _set_commit_timestamp(self, metric, errors) -> CommitMetric:
         # Only convert when commit_time is in metric, previously should be
         # found from the Label with fallback to annotation
-        if metric.commit_time:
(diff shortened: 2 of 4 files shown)


demo/tekton-demo-setup/03-build-and-deploy.yaml
```diff
@@ -94,7 +94,8 @@ objects:
             set -ex
             find /workspace
             cd /workspace/debug 
-            git log -1 --format=%cd > "$(results.commit-time.path)"
+            epoch_time=`git log -1 --format=%ct`
+            TZ=UTC date -d @`echo $epoch_time` | tee "$(results.commit-time.path)"
```
KevinMGranger, line 98: I don't understand this part, I thought the whole point was using epochs?

Also, this line can be simplified into `TZ=UTC date -d @"$epoch_time" | tee "$(results.commit-time.path)"`

weshayutin: @KevinMGranger you are not wrong about epochs.  Perhaps after discussing today we'll go for it.  We didn't want to change too much at this time.  I am concerned a little about changing the convert timestamp across all the exporters at this time.   Changing so it just works w/ the committime image exporter is a more smaller change IMHO.

exporters/committime/collector_image.py
```diff
@@ -120,10 +120,18 @@ def commit_metric_from_image(self, app: str, image, errors: list) -> CommitMetri
...
+        try:
+            if metric.commit_time:
+                metric.commit_timestamp = pelorus.convert_date_time_to_timestamp(
+                    metric.commit_time, self._timedate_format
+                )
+        except ValueError:
+            errors.append(
+                "COMMIT_DATE_FORMAT does not match annotation commit.date on the image\n"
             )
+            errors.append("Removing the offending timestamp")
```
KevinMGranger, line 132: This is supposed to be a list of separate errors, not a logger. This should be the same call.

KevinMGranger:
This looks like it will be included in https://github.com/konveyor/pelorus/pull/720. Let's see the status on that before merging / closing this.

mpryc:
Closing, there was time format rework. May be not needed.

PR #665, opened 2022-09-21
Allow to use EPOCH timestamp in the image annotation

mpryc:
For the images that were build without commit time or hash data we allow to annotate them for the Commit Time exporter to work properly. This change allows to annotate with EPOCH time format.

@redhat-cop/mdt

### DIFF
--- a/exporters/committime/collector_image.py
+++ b/exporters/committime/collector_image.py
@@ -20,7 +20,7 @@ from typing import Iterable, Optional
 
 import pelorus
 from committime import CommitMetric
-from pelorus.timeutil import parse_guessing_timezone_DYNAMIC
+from pelorus.timeutil import parse_guessing_timezone_DYNAMIC, to_epoch_from_string
 from pelorus.utils import collect_bad_attribute_path_error, get_env_var, get_nested
 
 from .collector_base import COMMIT_DATE_ANNOTATION_ENV, AbstractCommitCollector
@@ -108,7 +108,10 @@ class ImageCommitCollector(AbstractCommitCollector):
             metric = self._set_commit_time_from_annotations(metric, errors)
 
         if not metric.commit_hash:
-            # We ignore all the errors by passing "None" as a string, because commit hash isn't required.
+            # We ignore all the errors by passing [], because commit hash isn't required.
+            metric = self._set_commit_hash_from_annotations(metric, [])
+        if not metric.commit_hash:
+            # We ensure None is passed as string
             metric.commit_hash = "None"
         metric = self._set_commit_timestamp(metric, errors)
 
@@ -121,9 +124,15 @@ class ImageCommitCollector(AbstractCommitCollector):
         # Only convert when commit_time is in metric, previously should be
         # found from the Label with fallback to annotation
         if metric.commit_time:
-            metric.commit_timestamp = parse_guessing_timezone_DYNAMIC(
-                metric.commit_time, format=self._timedate_format
-            ).timestamp()
+            try:
+                metric.commit_timestamp = to_epoch_from_string(
+                    metric.commit_time
+                ).timestamp()
+            except (ValueError, AttributeError):
+                # Do nothing here as we tried with EPOCH timestamp
+                metric.commit_timestamp = parse_guessing_timezone_DYNAMIC(
+                    metric.commit_time, format=self._timedate_format
+                ).timestamp()
         return metric
 
     def get_commit_time(self, metric) -> Optional[CommitMetric]:
--- a/exporters/pelorus/timeutil.py
+++ b/exporters/pelorus/timeutil.py
@@ -62,6 +62,24 @@ def parse_guessing_timezone_DYNAMIC(timestring: str, format: str) -> datetime:
         return parsed.replace(tzinfo=timezone.utc)
 
 
+def to_epoch_from_string(timestring: str) -> datetime:
+    """
+    It's really EPOCH to EPOCH conversion with datetime return object.
+
+    If timestring matches expected EPOCH format a proper datetime object is returned,
+    otherwise raises ValueError and consumer of this function probably want's to use
+    other format of timestamp.
+    """
+    epoch_date_time = timestring.split(".")[0]
+    # Try to convert to an EPOCH, but only if it's 10 digit
+    if len(epoch_date_time) != 10:
+        raise ValueError(
+            f"Tried to get epoch from not allowed string length: {timestring}"
+        )
+    else:
+        return datetime.fromtimestamp(int(epoch_date_time))
+
+
 def second_precision(dt: datetime) -> datetime:
     """
     Change the datetime to have second precision (removing microseconds).
--- a/exporters/tests/test_datetime.py
+++ b/exporters/tests/test_datetime.py
@@ -1,10 +1,13 @@
 from datetime import datetime, timedelta, timezone
 
+import pytest
+
 import provider_common.github as github
 from pelorus.timeutil import (
     parse_assuming_utc,
     parse_guessing_timezone_DYNAMIC,
     parse_tz_aware,
+    to_epoch_from_string,
 )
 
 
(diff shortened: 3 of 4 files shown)


exporters/pelorus/timeutil.py
```diff
@@ -62,6 +62,26 @@ def parse_guessing_timezone_DYNAMIC(timestring: str, format: str) -> datetime:
...
+    other format of timestamp.
+    """
+    epoch_date_time = timestring.split(".")[0]
+    # Try to convert to an EPOCH, but only if it's 10 or 13 digit
+    if len(epoch_date_time) not in allowed_length:
+        raise ValueError(
+            f"Tried to get epoch from not allowed string length: {timestring} / {allowed_length}"
+        )
+    else:
+        return datetime.fromtimestamp(int(epoch_date_time))
```
KevinMGranger, line 82: I think we said we'd treat 13 digit as milliseconds, right? So we'd need to divide by 1000?

exporters/pelorus/timeutil.py
```diff
@@ -62,6 +62,26 @@ def parse_guessing_timezone_DYNAMIC(timestring: str, format: str) -> datetime:
         return parsed.replace(tzinfo=timezone.utc)
 
 
+def to_epoch_from_string(
+    timestring: str, allowed_length: list[int] = [10, 13]
```
KevinMGranger, line 66: Any reason to have customizable lengths?

exporters/pelorus/timeutil.py
mpryc: No, everything works fine here. Function allows to pass even 1 number and then expands that to *0

exporters/pelorus/timeutil.py
mpryc: just to make function nicer :)

exporters/pelorus/timeutil.py
```diff
@@ -62,6 +62,26 @@ def parse_guessing_timezone_DYNAMIC(timestring: str, format: str) -> datetime:
...
+) -> datetime:
+    """
+    It's really EPOCH to EPOCH conversion with datetime return object.
+
+    If timestring matches expected EPOCH format a proper datetime object is returned,
+    otherwise raises ValueError and consumer of this function probably want's to use
+    other format of timestamp.
```
KevinMGranger, line 73: Since this is just a best guess and they'll want to fall back, maybe this should just return `None` if it doesn't work? Then you could do `(to_epoch_from_string(ts) or parse_guessing_timezone_DYNAMIC(ts, format)).timestamp()`

exporters/committime/collector_image.py
```diff
@@ -121,9 +121,15 @@ def _set_commit_timestamp(self, metric: CommitMetric, errors) -> CommitMetric:
...
         # found from the Label with fallback to annotation
         if metric.commit_time:
-            metric.commit_timestamp = parse_guessing_timezone_DYNAMIC(
-                metric.commit_time, format=self._timedate_format
-            ).timestamp()
+            try:
+                metric.commit_timestamp = to_epoch_from_string(
+                    metric.commit_time
+                ).timestamp()
+            except (ValueError, AttributeError):
```
KevinMGranger, line 128: When would an AttributeError happen?

mpryc: When you pass integer and not str(int) as input for example.

exporters/pelorus/timeutil.py
mpryc: I prefer to give explicit error rather then silently fall with None in case someone will want to use that function later, but of course can change if you disagree.

exporters/pelorus/timeutil.py
KevinMGranger: I don't understand, what would it mean if someone passed something like `[5]`? It's not like we enforce leading zeroes-- it's just to determine milliseconds versus seconds, right?

exporters/committime/collector_image.py
KevinMGranger: Shouldn't we check for that and raise a `TypeError` ourselves, so we know where it came from?

exporters/pelorus/timeutil.py
KevinMGranger: I think this depends on the length discussion-- I'll make a bigger followup comment.

KevinMGranger:
I'm still confused by the epoch function, but I might be misremembering what we wanted.

I think we said:

- assume it's an epoch
- assume a long epoch string means it's in milliseconds

If we're allowing 13-digit epochs, that means it's in milliseconds, or it refers to the year `33658` at the earliest. Which will cause an exception from `fromtimestamp`.

I also really don't get why customizing the number of digits is necessary. Where to do use any other values than the defaults?

Finally, there's one gotcha: it needs to be `fromtimestamp(timestamp, timezone.utc)`, otherwise we get a naive datetime again. Those are really annoying. We should have a unit test to catch that.

exporters/pelorus/timeutil.py
mpryc: I am trying to ensure some annotation is not consisting of numbers that are 6,7,8,9,22 in length see couple of examples:

```
>>> from datetime import datetime
>>> datetime.fromtimestamp(int("123"))
datetime.datetime(1970, 1, 1, 1, 2, 3)
>>> datetime.fromtimestamp(int("1234567"))
datetime.datetime(1970, 1, 15, 7, 56, 7)
>>> datetime.fromtimestamp(int("123456789012"))
datetime.datetime(5882, 3, 11, 1, 30, 12)
>>> datetime.fromtimestamp(int("00000000000013900000434"))
datetime.datetime(2410, 6, 22, 17, 13, 54)
>>> datetime.fromtimestamp(int("00000000000001600000434"))
datetime.datetime(2020, 9, 13, 14, 33, 54)
```

So we are limiting here stupid human errors a bit. See the last example which happens to be quite fine date...

mpryc: actually we should allow only seconds, see:

```
>>> datetime.fromtimestamp(int("1663772656490"))
  File "<stdin>", line 1, in <module>
ValueError: year 54692 is out of range
```

weshayutin:
Screenshot from 2022-09-21 10-13-13
Screenshot from 2022-09-21 10-13-28

exporters/committime/collector_image.py
KevinMGranger: Now that I think about it-- when would `metric.commit_time` ever be anything but a string?

exporters/pelorus/timeutil.py
```diff
@@ -62,6 +62,24 @@ def parse_guessing_timezone_DYNAMIC(timestring: str, format: str) -> datetime:
...
+    """
+    epoch_date_time = timestring.split(".")[0]
+    # Try to convert to an EPOCH, but only if it's 10 digit
+    if len(epoch_date_time) != 10:
+        raise ValueError(
+            f"Tried to get epoch from not allowed string length: {timestring}"
+        )
+    else:
+        return datetime.fromtimestamp(int(epoch_date_time))
```
KevinMGranger, line 80: While we still get the correct data, I really think we should be returning a timezone-aware datetime. For instance, since it's in localtime, what happens if there's a DST transition? I don't know, but if we add `timezone.utc`, we don't _need_ to know.

exporters/tests/test_datetime.py
```diff
@@ -102,3 +105,29 @@ def test_github():
...
+        ("1652305808", "1652305808.0"),
+        ("1652305808.0", "1652305808.0"),
+    ],
+)
+def test_to_epoch_from_string(timestamps, expected):
+    epoch_from_str = to_epoch_from_string(timestamps)
+    assert str(epoch_from_str.timestamp()) == expected
+
+
+@pytest.mark.xfail(raises=ValueError)
```
KevinMGranger, line 124: `xfail` is meant for tests that are _temporarily_ failing, because you want to acknowledge the existence of a bug, or because something isn't implemented yet.

To test that an exception is raised, you'd do

```python
with pytest.raises(ValueError):
    to_epoch_from_string(timestamps)
```

Though  it seems like a nit, it _does_ affect test output.

KevinMGranger (changes requested):

KevinMGranger (approved):

KevinMGranger (approved):

weshayutin (approved):

PR #666, opened 2022-09-21
update tekton to use epoch

weshayutin:
Use epoch for commit timestamp
Updated README
Added an example values.yaml file

### DIFF
--- a/demo/tekton-demo-setup/03-build-and-deploy.yaml
+++ b/demo/tekton-demo-setup/03-build-and-deploy.yaml
@@ -93,8 +93,13 @@ objects:
           script: |
             set -ex
             find /workspace
-            cd /workspace/debug 
-            git log -1 --format=%cd > "$(results.commit-time.path)"
+            cd /workspace/debug
+            printf "epoch timestamps are recommended: git log -1 --format=%ct"
+            epoch=`git log -1 --format=%ct`
+            human=`date -d @$epoch`
+            printf "epoch time: " + $epoch
+            printf "human readable time: " + $human
+            echo $epoch > "$(results.commit-time.path)"
         - name: debug-time-last-commit
           image: ubi9/toolbox
           script: |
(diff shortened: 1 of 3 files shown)


demo/tekton-demo-setup/03-build-and-deploy.yaml
```diff
@@ -93,8 +93,13 @@ objects:
           script: |
             set -ex
             find /workspace
-            cd /workspace/debug 
-            git log -1 --format=%cd > "$(results.commit-time.path)"
+            cd /workspace/debug
+            printf "epoch timestamps are recommended: git log -1 --format=%ct"
```
KevinMGranger, line 97: These printf calls won't have newlines, any problem with `echo`?

Also, what if `%c` is interpreted by printf? Could be brittle.

mpryc: should be:

```
printf "epoch timestamps are recommended: git log -1 --format=%%ct\n"
```

or as @KevinMGranger suggested:
```
echo "epoch timestamps are recommended: git log -1 --format=%ct"
```

weshayutin: it works and looks fine imho, here's the output.
```
+ cd /workspace/debug
+ printf 'epoch timestamps are recommended: git log -1 --format=%ct'
epoch timestamps are recommended: git log -1 --format=t++ git log -1 --format=%ct
+ epoch=1663770043
++ date -d @1663770043
+ human='Wed Sep 21 14:20:43 UTC 2022'
epoch time: + printf 'epoch time: ' + 1663770043
human readable time: + printf 'human readable time: ' + Wed Sep 21 14:20:43 UTC 2022
+ echo 1663770043
STEP-DEBUG-TIME-LAST-COMMIT

1663770043
```

KevinMGranger: `epoch timestamps are recommended: git log -1 --format=t`
It swallowed the `%c`. You only see the correct output because it's in debug mode where it echoes each line (`set -x`).

Since it's just for debugging I guess it's fine, but this is brittle.

mpryc (approved):

PR #677, opened 2022-09-29
clean up pv && pvc prior to execution

weshayutin:
* not having the pv and pvc cleaned up will cause the demo to fail
* message user and delete both the pvc and pv

Note: this is getting cumbersome to clean $everything to keep the pipeline history but I suppose we're done now.  This is to be compared to just nuking the namespace.


mpryc:
this one is ok, I think we should consider adding some cleanup of OCP resources in the cleanup_and_exit function https://github.com/konveyor/pelorus/blob/master/demo/demo-tekton.sh#L134, but that's not what this PR is for.

/lgtm

mpryc (approved):

PR #686, opened 2022-10-11
Helm chart operator preparation

KevinMGranger:

This will reorganize the existing helm charts in a way that the operator-sdk can work with.

It will also document our approach.

/hold until we have fully separated everything.


KevinMGranger:
Superseded by #696

Issue #594, opened 2022-08-04
Documentation Edits for Centralization/Standardization


scoonRH:
Editing existing Pelorus documentation to standardize across all tools for eventual move to a centralized documentation website (https://konveyor.github.io/) and repo (https://github.com/konveyor/konveyor.github.io/tree/main).

scoonRH:
Do not have privileges to apply a label. Please apply doc label.

PR #691, opened 2022-10-13
Switch documentation to material theme for mkdocs

etsauer:

I propose switching from the default readthedocs theme to the [Material for MkDocs theme](https://squidfunk.github.io/mkdocs-material/). This theme in my opinion makes the documentation look and read much cleaner, as well as includes several built-in UX enhancements we can start to take advantage of, including codeblock captions, annotations, buttons, enhanced icon support, and more.

## Linked Issues?

resolves #<issue number>

related to #<issue number>

## Testing Instructions

Check that the readthedocs PR build passes, and view the demo content via readthedocs or `mkdocs serve`

@redhat-cop/mdt


mpryc (changes requested):
I really like the new theme, but it's personal, so will also allow @weshayutin and @KevinMGranger to comment.

I am requesting change to add the following python dependencies to the `exporters/requirements-dev.txt` file:

```
# Required by mkdocs material theme
pymdown-extensions
mkdocs-material
```

Without above the `mkdocs serve` won't work on the local dev setup.

etsauer:
@mpryc yes, I can add those dependencies. However, just one note that, in order for readthedocs to build properly, it requires an mkdocs-specific requirements file it can use to build, which I've placed in `docs/requirements.txt`. Should we include those dependencies in two places?

mpryc:
> @mpryc yes, I can add those dependencies. However, just one note that, in order for readthedocs to build properly, it requires an mkdocs-specific requirements file it can use to build, which I've placed in `docs/requirements.txt`. Should we include those dependencies in two places?

You are right it doesn't make sense to put those in 2 different places. Makefile should be updated to install also those when preparing dev-env: https://github.com/konveyor/pelorus/blob/master/Makefile#L58-L59

mpryc (approved):

mateusoliveira43:
Really liked the dark mode support of the theme. I just noticed one of the images, `docs/img/architecture.png`, does not look so good in dark mode
no-background

A question: should Pelorus use the same theme as some of the other Konveyor projects? (related to #594)

etsauer:
@mateusoliveira43 good catch. Updated both diagrams on that page to look better in dark mode.

KevinMGranger (approved):

PR #693, opened 2022-10-17
fix: Allow missing projects in JIRA exporter

### LINKED ISSUE #455: JIRA exporter does not give results if one of the projects is missing
If we configure JIRA exporter in a way that scope of the projects is defined:

`PROJECTS=firstproject,secondproject,thirdproject`

and one of the projects is missing in JIRA, e.g. only secondproject is missing.

Then none of the issues are returned and JIRA exporter fails to feed data properly.

mateusoliveira43:

Ignores missing projects when exporting JIRA issues. 

## Linked Issues?

resolves #455

## Testing Instructions

Uncomment the test in `exporters/tests/test_failure_exporter.py` with your credentials and run integration tests.

@redhat-cop/mdt

### DIFF
--- a/exporters/README.md
+++ b/exporters/README.md
@@ -5,4 +5,4 @@ Before deploying, you must export your GITHUB_REPOS, GITHUB_USER and GITHUB_TOKE
     oc process -f templates/github-secret.yaml -p GITHUB_USER=${GITHUB_USER} -p GITHUB_TOKEN=${GITHUB_TOKEN} -p GITHUB_REPOS=${GITHUB_REPOS} | oc apply -f-
     oc process -f templates/exporter.yaml -p APP_NAME=commit-exporter | oc apply -f-
 
-Configuration can be found in the [config guide](./docs/Configuration.md)
+Configuration can be found in the [config guide](https://pelorus.readthedocs.io/en/latest/Configuration/).
--- a/exporters/README.md
+++ b/exporters/README.md
@@ -5,4 +5,4 @@ Before deploying, you must export your GITHUB_REPOS, GITHUB_USER and GITHUB_TOKE
     oc process -f templates/github-secret.yaml -p GITHUB_USER=${GITHUB_USER} -p GITHUB_TOKEN=${GITHUB_TOKEN} -p GITHUB_REPOS=${GITHUB_REPOS} | oc apply -f-
     oc process -f templates/exporter.yaml -p APP_NAME=commit-exporter | oc apply -f-
 
-Configuration can be found in the [config guide](https://pelorus.readthedocs.io/en/latest/Configuration/).
+Configuration can be found in the [config guide](https://pelorus.readthedocs.io/en/latest/GettingStarted/configuration/PelorusExporters/).
--- a/exporters/failure/collector_jira.py
+++ b/exporters/failure/collector_jira.py
@@ -21,6 +21,9 @@ from typing import Any, Optional, cast
 from attrs import define, field
 from jira import JIRA
 from jira.client import ResultList
+from typing import List, Optional
+
+from jira import JIRA, Issue
 from jira.exceptions import JIRAError
 
 from failure.collector_base import AbstractFailureCollector, TrackerIssue
@@ -29,9 +32,6 @@ from pelorus.config.converters import comma_or_whitespace_separated
 from pelorus.config.log import REDACT, log
 from pelorus.timeutil import parse_tz_aware, second_precision
 
-# One query limit, exporter will query multiple times.
-# Do not exceed 100 as JIRA won't return more.
-JIRA_SEARCH_RESULTS = 50
 QUERY_RESULT_FIELDS = (
     "summary,labels,created,resolutiondate,status,statuscategorychangedate"
 )
@@ -39,16 +39,15 @@ DEFAULT_JQL_SEARCH_QUERY = 'type in ("Bug") AND priority in ("Highest")'
 JQL_SEARCH_QUERY_ENV = "JIRA_JQL_SEARCH_QUERY"
 # User specified JIRA comma separated statuses for resolved issue
 RESOLVED_STATUS_ENV = "JIRA_RESOLVED_STATUS"
-
+NON_EXISTING_PROJECT_ERROR_START = "The value '"
+NON_EXISTING_PROJECT_ERROR_END = "' does not exist for the field 'project'."
 
 _DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%f%z"
 
 
 @define(kw_only=True)
 class JiraFailureCollector(AbstractFailureCollector):
-    """
-    Jira implementation of a FailureCollector
-    """
+    """JIRA implementation of a FailureCollector."""
 
     username: str = field(default="", metadata=env_vars(*env_var_names.USERNAME))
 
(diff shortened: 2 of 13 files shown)


mateusoliveira43:
I would change the logic of the `__init__` to only call JIRA connection once.

I did not test the integration of this fix with the whole project.

I did not write new tests (I was in a hurry, but I can do it in the following days).

mateusoliveira43:
I updated the logic to only call JIRA connection once. But still did not write new tests.

exporters/failure/collector_jira.py
```diff
@@ -41,108 +38,128 @@
...
-                    issue.key, created_ts, resolution_ts, self.get_app_name(issue)
-                )
-                critical_issues.append(tracker_issue)
+            jira_client = self._connect_to_jira()
+            self._set_jql_queries_strings(jira_client)
+            logging.debug("JIRA JQL query: %s", self.jql_query_string)
+            jira_issues = jira_client.search_issues(
+                self.jql_query_string,
+                startAt=0,
+                maxResults=False,
```
mateusoliveira43, line 147: I looked through the JIRA python library documentation and code, and I believe this way is better. I compared the time execution this way and the old way and it was about 0.5 seconds faster (2.9 against 3.4, approximately)

exporters/failure/collector_jira.py
```diff
@@ -41,108 +38,128 @@
...
             if error.status_code == 400:
                 logging.error(
                     "Status: %s, Error Response: %s", error.status_code, error.text
                 )
                 logging.info("JIRA query: %s", self.jql_query_string)
-            else:
-                raise
-
-        return critical_issues
+                return []
```
mateusoliveira43, line 158: Here I would change to raise the error and do not return an empty list

exporters/tests/test_failure_exporter.py
```diff
@@ -126,6 +108,7 @@ def mock_jira_connect(self):
         projects=None,
     )
     collector.search_issues()
+    # TODO test?
```
mateusoliveira43, line 111: This test is testing something? There is  o assert

KevinMGranger (commented):
I'm liking what I'm seeing so far! I think we need to figure out what to do about that `400` error case, though. The old behavior, as unfortunate as it was, protected users against _any_ 400 error. Removing it would be more correct if it was just for missing projects, but then we might be breaking people's setups.

(This is the kind of thing telemetry would be good for, but we don't have that.)

exporters/failure/collector_jira.py
KevinMGranger: This whole section is much cleaner, thank you.

I just want to note for later (not asking you to change this): if someone has a _ton_ of issues (relatable), then the code will look like it's hanging. The old way had some log messages interspersed. If this becomes an issue later, we could make a small abstraction over the pagination ourselves. No need to complicate things now, though!

exporters/tests/test_failure_exporter.py
mateusoliveira43: Got it. A suggestion would be to check if nothing was raised, I believe we can do it with something like this https://github.com/mateusoliveira43/cly/blob/main/tests/unit/cly/test_config.py#L36

exporters/failure/collector_jira.py
```diff
@@ -41,108 +38,128 @@
...
+        # Create the JiraFailureMetric
+        created_tz = parse_tz_aware(issue.fields.created, _DATETIME_FORMAT)
+        created_ts = second_precision(created_tz).timestamp()
+        resolution_ts = self._get_resolved_timestamp(issue, self.jira_resolved_statuses)
+        return TrackerIssue(
+            issue.key, created_ts, resolution_ts, self.get_app_name(issue)
+        )
 
-        critical_issues = []
+    def search_issues(self) -> List[Optional[TrackerIssue]]:
```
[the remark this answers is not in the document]
mateusoliveira43: Here I made a mistake, I put the `Optional` trying to tell the list could be empty, but the correct is just `List[TrackerIssue]`. I will fix it.

exporters/failure/collector_jira.py
```diff
@@ -41,108 +38,128 @@
...
+        Set JQL query string for projects.
+
+        Parameters
+        ----------
+        jira_client : JIRA
+            JIRA instance.
+        """
+        if self.jql_query_string == DEFAULT_JQL_SEARCH_QUERY and self.projects:
+            _projects = set(self.projects.split(","))
+            existing_projects = {project.name for project in jira_client.projects()}
```
mpryc, line 102: I do wonder about performance impact here. If API allows to access many projects it may be heavy on that call?

mpryc (commented):
One question about performance impact.

Second is:
 What happens when the exporter starts in a pod at the time when JIRA project is not in jira and then project is being added to the JIRA?

exporters/failure/collector_jira.py
mateusoliveira43: Good point, I did not test and do not know the answer. I think we can test it out in a big example. If it is heavy, I believe we could change this part (lines 102 until 109) to
```python
            projects_str = ""
            for project in _projects:
                try:
                    jira_client.project(project)
                    projects_str += f'","{project}' if projects_str else project
                except JIRAError:
                    logging.error(
                        'Project of name "%s" does not exist in server %s',
                        project,
                        self.server,
                    )
```
I did not test if the `project` function works passing the name of the JIRA project.

mpryc: @mateusoliveira43 or even something else. Keep the previous logic (get all of the specified projects) and only if it fails apply the above. Above loop is much nicer then getting all imo.

mateusoliveira43: Alright, I will change it

mateusoliveira43: Here there is a problem: to use the `project` function we need to pass the key of the project, not its name. For example: in your example we have to pass "FIRST" instead of "Test". What you suggest to overcome this?

mateusoliveira43: I used the error message to get them, I do not like this solution because if the error message is updated by the the JIRA Python library, our code breaks.

mpryc: I think it's what we need to live with. Possibly some test to cover such scenario?

Here is a bit of code that is ugly and of course could be in recurrent function to have only logic once,  but falls-back to per project query if something goes wrong (maybe err 400 is enough here?):

```
from jira import JIRA
from jira.exceptions import JIRAError

api_key="***************"
email="***************"
server="https://pelorustest.atlassian.net"

projects_str="TEST,MyProject"

jql_query_string = ('type in ("Bug") AND priority in ("Highest") AND project in ("{}")').format(projects_str)

jira_client = JIRA(options={"server": server}, basic_auth=(email, api_key))
jira_search_results = []

try:
    jira_search_results = jira_client.search_issues(jql_query_string, startAt=0, maxResults=50, fields="summary",)
except JIRAError as jira_error:
    if jira_error.status_code == 400 and \
        "does not exist for the field 'project'" in jira_error.text:
        for project in projects_str.split(','):
            jql_query_string = ('type in ("Bug") AND priority in ("Highest") AND project in ("{}")').format(project.strip())
            try:
                jira_search_results += jira_client.search_issues(jql_query_string, startAt=0, maxResults=50, fields="summary",)
            except JIRAError as jira_error:
                if jira_error.status_code == 400 and \
                    "does not exist for the field 'project'" in jira_error.text:
                    pass

jira_search_results

```

mateusoliveira43: I implement something just like that, just added a logic to do the query maximum 2 times, instead of 1 + `n` (`n` being the number of projects passed by the user).

Yes, I still need to write tests, and I forgot to apply the `strip` function to the strings

KevinMGranger:
We should check if this fix is included in #720, or if we should rebase on top of that once it's merged.

exporters/tests/test_failure_exporter.py
```diff
@@ -72,58 +76,61 @@ def test_jira_pass_connection(server, username, apikey):
...
-    [
-        (
-            "https://pelorustest.atlassian.net",
-            "fake@user.com",
-            "WIEds4uZHiCGnrtmgQPn9E7D",
-        )
-    ],
-)
+# TODO write mockoon or integration to test success (and success without finding any issue)
+# @pytest.mark.parametrize("projects", ["non_existing,Test,wrong_name", "Test"])
```
mateusoliveira43, line 80: There is a way to pass real credentials in a safe way through github actions?

mateusoliveira43: we tried to add but did not work. Gone a open an issue to fix this later

exporters/README.md
```diff
@@ -5,4 +5,4 @@ Before deploying, you must export your GITHUB_REPOS, GITHUB_USER and GITHUB_TOKE
     oc process -f templates/github-secret.yaml -p GITHUB_USER=${GITHUB_USER} -p GITHUB_TOKEN=${GITHUB_TOKEN} -p GITHUB_REPOS=${GITHUB_REPOS} | oc apply -f-
     oc process -f templates/exporter.yaml -p APP_NAME=commit-exporter | oc apply -f-
 
-Configuration can be found in the [config guide](./docs/Configuration.md)
+Configuration can be found in the [config guide](https://pelorus.readthedocs.io/en/latest/Configuration/).
```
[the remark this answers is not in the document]
mpryc: Correct, we don't

mpryc:
I had final round of testing. Everything looks great, tested works as expected.

One comment which is also on the main branch, so could be fixed as separate PR:
 - if we configure `APP_LABEL` together, `APP_LABEL` is ignored and all issues are returned. For our test server it should return 2 open/close items so 4 items, but all of the labels are returned:
 
```shell
export TOKEN=<token>
export API_USER=<user>
export SERVER=https://pelorustest.atlassian.net/
export APP_LABEL='my.app.label/name'
unset JIRA_JQL_SEARCH_QUERY
unset JIRA_RESOLVED_STATUS
python exporters/failure/app.py
```

mpryc (approved):

PR #695, opened 2022-10-19
Create cleaner social cards for link sharing

etsauer:
Signed-off-by: Eric Sauer <esauer@redhat.com>

When we changed to the material theme doe the docs site, it changed the logo image we use, which doesn't look nice when we share links to the pelorus site. The social cards feature from Material fixes that: https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/#built-in-social-plugin

## Testing Instructions

@redhat-cop/mdt

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -51,5 +51,4 @@ nav:
   - Support:
     - Upstream Support: UpstreamSupport.md
 plugins:
-  - social:
-      cards_dir: img
+  - social


etsauer:
The social cards aren't currently being served as expected. Opened a RTD issue to get some input: https://github.com/readthedocs/readthedocs.org/issues/9680

etsauer:
I got a comment back saying that there's currently not a way to make the build URLs dynamically replace our site URL, but that it may be comein in the future via some environment variable enhancements being discussed. With that, and considering that, by default our docs site point to the latest release, and not actually `latest` (master), I think we could just merge this now, and then check to see if the thumbnails in chat look better

KevinMGranger:
@etsauer could we define an environment variable for `site_url`? https://www.mkdocs.org/user-guide/configuration/#environment-variables

etsauer:
@KevinMGranger I looked into that, but since we don't control the hosting environment, we don't have access at the point where that would need to be set. That environment variable would already have to exist (which it doesn't) and then we could tap into it.

etsauer:
@KevinMGranger we can already verify that the images are being generated. Let's take the index page in this PR build as an example:

https://pelorus--695.org.readthedocs.build/en/695/

If you inspect the source and look in the `head` element, you'll see this tag:

<meta property="og:image" content="https://pelorus.readthedocs.io/en/latest/assets/images/social/index.png">

And you try to open the link there you'll get a 404:

https://pelorus.readthedocs.io/en/latest/assets/images/social/index.png

HOWEVER, if you replace the base URL of https://pelorus.readthedocs.io/en/latest/ with the PR base URL of https://pelorus--695.org.readthedocs.build/en/695/, you'll see the image pop up:

https://pelorus--695.org.readthedocs.build/en/695/assets/images/social/index.png

So everything is working right, except the `meta` tag is only ever going to point to the production URL, because that's whats in mkdocs.yml

PR #694, opened 2022-10-19
Fancy new homepage and restructured navigation

etsauer:
Signed-off-by: Eric Sauer <esauer@redhat.com>

## Linked Issues?

resolves #<issue number>

related to #<issue number>

## Testing Instructions

Please include any additional commands or pointers in addition to our [standard PR testing process](/docs/Development.md#testing-pull-requests).

@redhat-cop/mdt

PR #696, opened 2022-10-19
Helm chart rework

mpryc:
Helm chart rework for operator-sdk

Modifications of the existing charts and proper symlinks to allow working with operator-sdk and helm install commands.

To generate helm operator use scrit from #714

@redhat-cop/mdt


mpryc:
@KevinMGranger we could make it work with backwards-compatible change by simply having pelorus folder without dependencies on pelorus-operators and then another folder pelorus-operator from which operator is generated. Do you think it's important to keep backwards compatibility here ?

weshayutin:
Going to try this out in a moment.  During the design session we spoke to "backwards compatibility", but at the time what we meant by backwards compatibility was that a user could execute the helm install  or choose to perform an operator install.
With this change you have collapsed the helm install from two commands to one command.

IMHO users would not have a problem running one helm command vs. two in the original install. 
This is fine and very nicely done.

mpryc:
This works now with the: https://github.com/konveyor/pelorus/pull/696. Installed pelorus using operator ! :)

mpryc:
Documentation will be updated in separate PR as I don't want to cause problems with #694

KevinMGranger:
#728 is merged, this is now obsolete 🎉

PR #704, opened 2022-11-04
use epoch vs. object for sn issue date

### LINKED ISSUE #703: service-now issue collection fails on type error
https://termbin.com/d8ka

11-04-2022 15:31:57 INFO     Collected failure_creation_timestamp{ app=unknown, issue_number=INC0007002 } 2018-10-16 22:47:51+00:00
  File "/usr/lib64/python3.9/wsgiref/handlers.py", line 137, in run
    self.result = application(self.environ, self.start_response)
  File "/opt/app-root/lib64/python3.9/site-packages/prometheus_client/exposition.py", line 128, in prometheus_app
    status, headers, output = _bake_output(registry, accept_header, accept_encoding_header, params, disable_compression)
  File "/opt/app-root/lib64/python3.9/site-packages/prometheus_client/exposition.py", line 104, in _bake_output
    output = encoder(registry)
  File "/opt/app-root/lib64/python3.9/site-packages/prometheus_client/openmetrics/exposition.py", line 63, in generate_latest
    floatToGoString(s.value),
  File "/opt/app-root/lib64/python3.9/site-packages/prometheus_client/utils.py", line 9, in floatToGoString
    d = float(d)
TypeError: ("float() argument must be a string or a number, not 'datetime.datetime'", Metric(failure_creation_timestamp, Failure Creation Timestamp, gauge, , [Sample(name='failure_creation_timestamp', labels={'app': 'unknown', 'issue_number': 'INC0000060'}, value=datetime.datetime(2016, 12, 12, 7, 19, 57, tzinfo=datetime.timezone.utc), timestamp=None, exemplar=None), Sample(name='failure_creation_timestamp', labels={'app': 'unknown', 'issue_number': 'INC0009002'}, value=datetime.datetime(2018, 9, 16, 5, 49, 23, tzinfo=datetime.tim

weshayutin:
SN was using a datetime object for the issue start / resolution time.  Prometheus expects a float.  Looking at the github issue failure exporter, we use an epoch timestamp there.  So, this is easy we just need to convert the original timestamp to an epoch float.

resulting log from the sn failure exporter: https://termbin.com/jz0f
There are possible other issues, and there is no need to convert the timestamp twice. Just getting this up asap.

resolves #703

## Testing Instructions
Create a dev environment for service now


KevinMGranger:
Just added type annotations so this kind of thing would be more evident in the future. They've already caught some spots where we don't handle the `None` case!

KevinMGranger:
Using `%s` is recommended against because it's technically nonstandard. It's not documented in the [datetime module format codes](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes), and thus isn't guaranteed by python. it only works because they're passed to `strftime(3)`.

Does this really matter? Eh. We don't really support running this on non-POSIX systems.

As for the use of `second_precision`, that's mainly used for logging, so we don't need to worry about using that.

mpryc (approved):

mpryc:
passed all CI jobs, merging

PR #709, opened 2022-11-07
Pin macOS oc version to 4.10, download ARM on ARM macs

### LINKED ISSUE #707: dev env setup handle golang cert issues on macOS
There's an [outstanding issue](https://github.com/golang/go/issues/52010) with go's x509 library that broke invalid certificate handling on macOS. If you're using some local cluster with self-signed certs, ignoring their validity won't work.

If you're using OpenShift Local / CRC, then their bundled `oc` version actually works around this! But I always forget about that when remaking my dev environment 😅.

A reminder would be nice.

KevinMGranger:

Pin `oc` on macs to 4.10 to work around a golang x509 bug.

Also install the arm version of `oc` on arm macs.

## Linked Issues?

resolves #707

## Testing Instructions

1. Delete `.venv/bin/oc`
2. Run `make .venv/bin/oc`
3. Verify it installs correctly.

If on a mac:

1. Follow the above
2. try connecting to an OpenShift cluster with untrusted certs
3. run `file .venv/bin/oc` and verify it is arm formatted.

### DIFF
--- a/scripts/get_tool_dl_url.py
+++ b/scripts/get_tool_dl_url.py
@@ -7,16 +7,30 @@ import enum
 import platform
 import re
 import sys
-from typing import Callable, Iterable, NamedTuple
+from typing import Callable, Iterable, Literal, NamedTuple, cast
 
 import requests
 
 GITHUB_RELEASE_TEMPLATE = "https://api.github.com/repos/{}/releases/latest"
 
-OS = platform.system()
-ARCH = platform.machine()
+
+SUPPORTED_SYSTEMS = {"Linux", "Darwin"}
+
+_os = platform.system()
+if _os not in SUPPORTED_SYSTEMS:
+    sys.exit(f"Unsupported OS {_os}")
+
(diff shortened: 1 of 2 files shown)


scripts/get_tool_dl_url.py
```diff
@@ -39,7 +53,7 @@ class StandardTool:
             ARCH_PATTERN = re.compile("all|arm64")
         else:
             sys.exit(f"Unsupported architecture {ARCH}")
-    elif OS == "Linux":
+    elif OS == "Linux" and ARCH in X86_64_ARCH_NAMES:
```
KevinMGranger, line 56: Oopsie. Want y'all to see that this passed CI before I fix this.

mpryc (approved):

scripts/get_tool_dl_url.py
mpryc: I don't think Linux / Arm combo will be used here, so it's ok to merge.

Issue #708, opened 2022-11-07
Automatic Type Checking: view but don't block


KevinMGranger:
We've seen a few issues where type checking would have helped catch them before a runtime failure. I've been using type checking as it's built in to the vs code python extension.

We should add a type check as a pre-commit and CI step. While we're just starting to use it, it should just be a warning, not an error. We may decide to move towards errors some day. Even if it's just for certain files, for starters.

We do need to decide between MyPy and Pyright though.

|pyright | mypy |
|-------|-------|
| - more dependencies (I think it needs node.js as well as python) | + easier environment set up  |
 | + more featureful, on the forefront of some type checking evolution | - less featureful |
| + the default choice for the vs code python extension | - not the default |

I lean towards pyright personally, but we can experiment.

mateusoliveira43:
Really like the idea of type checking! 
About the VS Code comparison, I believe it should not be part of discussion. But to use `mypy` with it, I just add
```json
    "python.linting.mypyEnabled": true,
    "python.linting.mypyArgs": [
        "--show-column-numbers",
        "--strict",
    ],
```
to `settings.json` (and `mypy` should be installed in the virtual environment to it to work).

KevinMGranger:
I want to focus on a good development experience, so I think defaults matter. That's the only reason I brought up VS Code. But it is definitely easy to turn on mypy instead!

KevinMGranger:
Closed with #710 and #713.

PR #710, opened 2022-11-07
Introduce the start of automatic type checking

KevinMGranger:
We're not running it as part of pre-commit yet so we don't get error fatigue. ~~Unfortunately, it doesn't seem like there's an easy way to mark a check as a "warning": https://github.com/actions/toolkit/issues/399~~. Never mind! We can just mark the check as not required for merging.

Adds a type check target and runs type checking in CI.

Type checking is not a hard requirement, and is just informational right now.

## Linked Issues?

related to #708

## Testing Instructions

See CI logs.

To run it yourself, update your venv, make sure you have node installed on your system, and run `pyright`.

If you use VS Code and the python extension, you're probably already using pyright!

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -18,6 +18,12 @@ if ! make -k lint; then
     failed=1
 fi
 
+# echo "📋 Type Checking, expect failures for now"
+# if ! make -k typecheck; then
+#     printf "\n\n"
+#     echo "Type checking failed, although this is to be expected for now." >&2
+# fi
+
 printf "\n\n"
 
 if [[ failed -ne 0 ]]; then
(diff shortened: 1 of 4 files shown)


mpryc (dismissed):

scripts/pre-commit
```diff
@@ -18,6 +18,12 @@ if ! make -k lint; then
     failed=1
 fi
 
+# echo "📋 Type Checking, expect failures for now"
```
mpryc, line 21: Why it's commented out?

KevinMGranger: I have some work in a separate branch to completely remove existing type-check complaints. That's more involved work across all of committime.

Until then, I don't want the pre-commit check to be spammed with errors-- that's how one gets error fatigue and starts ignoring it.

I can remove this if it's cleaner than commenting it out. But I think of it as a feature flag, y'know? (Could also put it behind an environment variable).

mpryc (approved):

PR #712, opened 2022-11-07
Update GH actions, cache python deps

### LINKED ISSUE #711: Update outdated github actions
Some of them are causing deprecation warnings to be emitted: https://github.com/konveyor/pelorus/actions/runs/3414501849

KevinMGranger:

Updates the checkout and python github actions we use. This is to avoid bitrot, and get rid of deprecation warnings.

This also now lets us cache python dependencies-- now if you have actions run again for the same PR, it will reuse those dependencies unless any of the requirements files changed.

## Linked Issues?

resolves #711

## Testing Instructions

See GitHub actions and verify that they are working properly.

### DIFF
--- a/.github/workflows/build-pelorus.yaml
+++ b/.github/workflows/build-pelorus.yaml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       # Setup S2i and Build container image
       - name: Setup and Build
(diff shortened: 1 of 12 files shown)


mateusoliveira43:
It would be good to put all workflow that depends on `pip install` together in one file? (to only run it once) (with the cache this already improves time, but I think it would improve more seconds) 
I think adding a `continue-on-error: true` on it, would run all of then, even if one of them break

KevinMGranger:
I think the cache step accomplishes the same thing as putting them all in the same file. It's not that slow to begin with, this was more of a "why not?" change, since it already needed updating.

Keeping them as separate files reduces complexity, and means they're listed separately at the bottom of PRs.

mpryc (approved):

PR #720, opened 2022-11-14
clean up committime exporter

KevinMGranger:
(Depends on #713, wait until that is merged.)

There were several typing issues and unclear APIs in the committime exporter.

`CommitMetric` was used as a big blob.
It carried all inputs and outputs, with almost no guarantees for
when and where things would be set. It also had several unused fields.

The way it was meant to be handled by implementors was also unclear:
- do they modify it on the passed in `CommitMetric`? 
- do they return a `CommitMetric` with the committime updated?
- what if they return None?
- what if the information needed is missing?

There are several spots where a type checker would complain.
Including some spots where the variable being used could be unbound!

This PR will help things be cleaner, and handled more consistently.

PR #726, opened 2022-11-21
Deserialization framework

KevinMGranger:
Separates out the deserialization framework used to support #720.


exporters/pelorus/deserialization/__init__.py
```diff
@@ -0,0 +1,392 @@
...
+    return {_NESTED_PATH_KEY: path}
+
+
+def _is_attrs_class(cls: type) -> Optional[type["attr.AttrsInstance"]]:
+    """
+    Return the class if it is an attrs class, else None.
+
+    This exists because `attrs.has` is not a proper TypeGuard.
+    """
+    return cast("type[attr.AttrsInstance]", cls) if attrs.has(cls) else None
```
mpryc, line 77: Where the `attr.AttrsInstance" comes from? It's not part of the attr (at least in my venv). Does attr need minimum version which includes AttrsInstance ?

exporters/pelorus/deserialization/__init__.py
```diff
@@ -0,0 +1,392 @@
...
+
+# generic types such as dict[str, int] have two parts:
+# the "origin" and the "args".
+# The origin in this case is `dict`, and the args are `(str, int)`.
+# This lets us inspect these types properly.
+
+# in the future, these should be made into `typing.TypeGuard`s.
+
+
+def _is_dict_type(type_: type) -> Optional[tuple[type, type]]:
```
mpryc, line 90: Laud thinking here, why this isn't enough ?
```
type(type_) is dict
```

exporters/pelorus/deserialization/__init__.py
```diff
@@ -0,0 +1,392 @@
...
+        target_name = target_name if target_name else str(target_type)
+
+        src_name = src_name if src_name else type(src).__name__
+
+        self.structured_field_name_path.append(target_name)
+        self.unstructured_data_path.append(src_name)
+        try:
+            return self._deserialize(src, target_type)
+        finally:
+            self.unstructured_data_path.pop()
```
mpryc, line 180: .pop() gets exactly one element, could you explain how this works if `self.unstructured_data_path` contains more then one element? I don't see recursion to ensure all elements are removed from this stack.

mpryc (commented):
Some questions.

mpryc:
Also files `*.pyi` could be added to the .gitignore?

mpryc (approved):

PR #728, opened 2022-11-23
Helm chart rework for operator-sdk

mpryc:
Modifications of the existing charts and proper symlinks to allow working with operator-sdk and helm install commands.

NOTE: This is separate PR to the #696. One of the PRs will be abandoned once we set on one approach.

The #696 revealed an issue with extra parameters from the values.yaml that required sub-level values for sub-charts and so we would be incompatible with current values.yaml structure for things like thanos setup.

This approach moves back the pelorus-core under standard pelorus application and makes subchart for the exporters which should be compatible with operator-sdk.

@redhat-cop/mdt


KevinMGranger (approved):
🎉

Issue #732, opened 2022-11-29
operator manager pod is throwing error log about grafana post install hook


mpryc:
Currently the manager pod is presenting many errors around grafana post-install hook, that tries to install grafana-api serviceaccount which is already existing. This may be a cause of other problems such as #731 ?

There is `Hook post-install pelorus/templates/grafana-post-install-hook.yaml` hook which should be reworked as manager is complaining that `serviceaccounts "grafana-api" already exists`

_Originally posted by @mpryc in https://github.com/konveyor/pelorus/issues/715#issuecomment-1329000462_

mpryc:
Resolved with #740

PR #740, opened 2022-12-05
Deprecate Grafana snapshot job

mpryc:
Grafana snapshot job was aimed to give access to the metrics without login to the grafana dashboard.

This feature isn't working as expected and currently can be deprecated.

If needed we should consider:
 1. Creating anonymous access to the grafana instance
 2. Rework current logic to create such snapshots

@redhat-cop/mdt


mpryc:
Fixes #732

mpryc:
@KevinMGranger can we merge this without conftest? I have disabled already those, however they are in the list of checks as disable happened after it failed?

weshayutin (approved):

PR #745, opened 2022-12-09
Feat/check available bumps pelorus operator

mateusoliveira43:

Create scheduled work to check if pelorus operator can be bumped.

## Linked Issues?

No linked issues.

## Testing Instructions

To see the new command help, run `make help`
To run new command, run `make check-bump-pelorus-operator`

### DIFF
--- a/.github/workflows/check-update-pelorus-operator.yml
+++ b/.github/workflows/check-update-pelorus-operator.yml
@@ -1,0 +1,25 @@
+name: Check for updates for pelorus operator
+
+on:
+  schedule:
+    - cron:  '0 0 * * 1' # Every Monday at 00:00 UTC
+
+jobs:
+  project-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v3
+        with:
+          python-version: "3.11"
+          cache: "pip"
+          cache-dependency-path: |
+            **/requirements*.txt
+            pyproject.toml
+
+      - name: Check if Pelorus operator was bumped
+        run: |
+          make check-bump-pelorus-operator
+          git restore pelorus-operator/helm-charts/pelorus/Chart.lock
+          test -z "$(git status --short)"
--- a/scripts/create_pelorus_operator
+++ b/scripts/create_pelorus_operator
@@ -40,18 +40,20 @@ function print_help() {
     printf "\t  -d\tpath to EMPTY destination folder, default: %s\n" "${DEFAULT_PELORUS_OPERATOR_DIR}"
     printf "\t  -p\tpath to DIR with operator patches to be applied, default: %s\n" "${DEFAULT_PELORUS_OPERATOR_PATCHES_DIR}"
     printf "\t  -v\toperator version to be used in a format x.y.z. If not set z+1 is used.\n"
+    printf "\t  -n\tdo NOT update version (just for development) %s\n"
     exit 0
 }
 
 ### Options
 OPTIND=1
-while getopts "s:d:p:h?v:" option; do
+while getopts "s:d:p:n?:h?v" option; do
     case "$option" in
     h|\?) print_help;;
     s)    source_dir=$OPTARG;;
     d)    destination_dir=$OPTARG;;
     p)    patches_dir=$OPTARG;;
     v)    operator_ver=$OPTARG;;
+    n)    no_update=true;;
     esac
 done
 
@@ -85,8 +87,12 @@ fi
 # Check if destination directory is empty and exists
 if [ -d "${destination_dir}" ]; then
 	if [ "$(ls -A "${destination_dir}")" ]; then
-        echo "ERROR: Destination directory ${destination_dir} is not empty, can not continue."
-        exit 2
+        # echo "ERROR: Destination directory ${destination_dir} is not empty, can not continue."
+        # exit 2
+        file_version=$(cat ${destination_dir}/Makefile | grep "VERSION ?= ")
+        file_version=${file_version:11}
+        echo "WARNING: Destination directory ${destination_dir} is not empty, deleting files."
+        rm -rf ${destination_dir} && mkdir ${destination_dir}
 	fi
 else
 	echo "ERROR: Destination directory ${destination_dir} do not exists, please create one and re-run script."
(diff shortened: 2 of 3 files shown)


.github/workflows/check-update-pelorus-operator.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  schedule:
+    - cron:  '0 0 * * 1' # Every Monday at 00:00 UTC
```
mateusoliveira43, line 11: I think weekly is enough, what you think?
I just entered a random date, we can choose when it is best to run it

.github/workflows/check-update-pelorus-operator.yml
```diff
@@ -0,0 +1,31 @@
+name: Check for updates for pelorus operator
+
+on:
+  push:
```
mateusoliveira43, line 4: just run on schedule?

.github/workflows/check-update-pelorus-operator.yml
```diff
@@ -0,0 +1,31 @@
...
+          cache: "pip"
+          cache-dependency-path: |
+            **/requirements*.txt
+            pyproject.toml
+
+      - name: Check if Pelorus operator was bumped
+        run: |
+          make bump-pelorus-operator
+          git restore pelorus-operator/helm-charts/pelorus/Chart.lock
+          test -z "$(git status --short)"
```
mateusoliveira43, line 31: I will check how to check if the job failed, and if so, inform a message like "Updates can be made to pelorus operator. Run `make bump-pelorus-operator` to update it."

.github/workflows/check-update-pelorus-operator.yml
```diff
@@ -0,0 +1,31 @@
...
+          python-version: "3.10"
+          cache: "pip"
+          cache-dependency-path: |
+            **/requirements*.txt
+            pyproject.toml
+
+      - name: Check if Pelorus operator was bumped
+        run: |
+          make bump-pelorus-operator
+          git restore pelorus-operator/helm-charts/pelorus/Chart.lock
```
mateusoliveira43, line 30: here I am ignoring chart.lock changes, but there is still the false positive changes we have to treat (when the script changes yaml sections position)

scripts/create_pelorus_operator
```diff
@@ -115,15 +121,23 @@ pushd "${destination_dir}" || exit 2
     # Correct operator version, to be latest +1
     OPERATOR_VERSION=$(curl -H "Authorization: Bearer XYZ" -X GET "https://quay.io/api/v1/repository/${OPERATOR_ORG_NAME}/${OPERATOR_PROJECT_NAME}/tag/" | jq .tags[].name | head -1 | sed -e 's|\"||g')
     echo "INFO: Current operator version: ${OPERATOR_VERSION}"
-    NEW_OPERATOR_VERSION=$(echo ${OPERATOR_VERSION} | awk -F. -v OFS=. '{$NF += 1 ; print}')
+    if [ ! -z ${file_version} ] && [ ${file_version} != ${OPERATOR_VERSION} ]; then
+        echo "ERROR: file version ${file_version} does not match released version ${OPERATOR_VERSION}."
+        exit 2
+    fi
+    if $no_update; then
```
mateusoliveira43, line 128: @mpryc added the logic to not update version, but there are still false positives

scripts/create_pelorus_operator
```diff
@@ -115,15 +121,23 @@ pushd "${destination_dir}" || exit 2
     # Correct operator version, to be latest +1
     OPERATOR_VERSION=$(curl -H "Authorization: Bearer XYZ" -X GET "https://quay.io/api/v1/repository/${OPERATOR_ORG_NAME}/${OPERATOR_PROJECT_NAME}/tag/" | jq .tags[].name | head -1 | sed -e 's|\"||g')
     echo "INFO: Current operator version: ${OPERATOR_VERSION}"
-    NEW_OPERATOR_VERSION=$(echo ${OPERATOR_VERSION} | awk -F. -v OFS=. '{$NF += 1 ; print}')
+    if [ ! -z ${file_version} ] && [ ${file_version} != ${OPERATOR_VERSION} ]; then
+        echo "ERROR: file version ${file_version} does not match released version ${OPERATOR_VERSION}."
```
mateusoliveira43, line 125: @mpryc added this to avoid duplicated versions in repo

KevinMGranger (commented):
The stuff I commented on looks good. But I'll wait to hear from @mpryc since he's more experienced with the operator stuff.

.github/workflows/check-update-pelorus-operator.yml
mateusoliveira43: I will change the rule to apply only for schedule and related operator stuff

mateusoliveira43:
discuss after https://github.com/konveyor/pelorus/issues/841 is resolved if this PR still makes sense

mateusoliveira43:
Alright, I think this PR is not necessary

I believe what made me want to create it was that each time you run the create-operator script, you can get a different operator sdk version from [here](https://github.com/operator-framework/operator-sdk/releases). So, we only need to check if there is a newer version of it, instead of creating all this logic.

But I think that our logic is wrong in our scripts today. For example, the latest releases of operator sdk today are 1.26.1 and 1.27.0, and our scripts install 1.26.1 because its newer. We should check which is bigger, right?

My idea is: 
- create a CI job to check if the operator sdk is the newest (and bigger) possible when there are changes in pelorus-operator folder; if not, break and warn the about the newer version available.
- update our installation scripts to install the newest (and bigger) possible versions of each software.

Do you agree? @mpryc @KevinMGranger @weshayutin 

If so, I will close this PR and open issues to address each point.

mateusoliveira43:
@mpryc  but we are not forced to just release bigger versions, are we?

For example, current Python is 3.11, but this does not mean they will not release patches to 3.10, right? I believe this is not a mistake from the software versioning.

mpryc:
> @mpryc but we are not forced to just release bigger versions, are we?
> 
> For example, current Python is 3.11, but this does not mean they will not release patches to 3.10, right? I believe this is not a mistake from the software versioning.

That is correct, however latest tag should in this case point to the 3.11 and not the 3.10-update which may have happened after 3.11 release.

mpryc:
I am closing this PR.

In my opinion it is not required. There was an issue with `operator-sdk` release where the following order was made:


```
$ curl --silent "https://api.github.com/repos/operator-framework/operator-sdk/releases" | jq -r '.[].tag_name'
v1.28.0
v1.26.1
v1.27.0
v1.25.4
v1.26.0
v1.25.3
v1.25.2
v1.25.1
v1.25.0
v1.24.1
v1.24.0
v1.23.0
v1.22.2
v1.22.1
v1.22.0
v1.21.0
v1.20.1
v1.20.0
v1.19.1
v1.19.0
v1.18.1
v1.18.0
v1.17.0
v1.16.0
v1.15.0
v1.14.0
v1.13.1
v1.13.0
v1.12.0
v1.11.0
```

As you can see  there were two times where the latest tag was not the latest number:

```
v1.26.1       <- After  v1.27.0
v1.25.4       <- After  v1.26.0
```

But it's not our responsibility to decide what is the `latest` tag for a particular repository, maybe there was some decision behind making.

Logged but in the operator-sdk project: https://github.com/operator-framework/operator-sdk/issues/6403

Closing this one, we should monitor the answers of the above issue and see if this one will need to be re-opened.

With the latest tag v1.28.0 we are safe, but we need to get answers to not hit this again in the future.

PR #749, opened 2022-12-12
Deserialization: allow retaining source, fix getitem handling

KevinMGranger:

1. kubernetes dynamic `ResourceField`s support `__getitem__`. However, you can't do a `issubclass` check on a `Mapping`, unlike some other "magic" checks. This replaces the `issubclass` / `isinstance` work with a simple attribute lookup.
2. Sometimes, keeping the source of the deserialization is useful. A new metadata option has been added for that.

## Linked Issues?

This helps support the work in #720 .

## Testing Instructions

Run unit tests as usual.

### DIFF
--- a/exporters/pelorus/deserialization/__init__.py
+++ b/exporters/pelorus/deserialization/__init__.py
@@ -59,19 +59,21 @@ import typing
 from typing import (
     Any,
     Iterable,
+    Literal,
     Mapping,
-    MutableMapping,
     MutableSequence,
+    NamedTuple,
     Optional,
     Sequence,
     TypeVar,
     Union,
-    cast,
+    get_type_hints,
     overload,
 )
 
 import attr
 import attrs
+from typing_extensions import TypeGuard
 
 from pelorus.deserialization.errors import (
     DeserializationErrors,
@@ -86,6 +88,7 @@ from pelorus.utils._attrs_compat import NOTHING
 
 # metadata
 _NESTED_PATH_KEY = "__pelorus_structure_nested_path"
+_RETAIN_SOURCE_KEY = "__pelorus_structure_retain_source"
 
 T = TypeVar("T")
 
@@ -100,14 +103,73 @@ def nested(path: Union[str, Sequence[str]]) -> dict[str, Sequence[str]]:
     return {_NESTED_PATH_KEY: split_path(path)}
 
 
-def _is_attrs_class(cls: type) -> Optional[type["attr.AttrsInstance"]]:
+def retain_source(retain: bool = True) -> dict[str, bool]:
+    """
+    Retain the source of the deserialization in this field.
+    """
+    return {_RETAIN_SOURCE_KEY: retain}
+
+
+# python has some magic to make classes work with `issubclass` and instances work with `isinstance`,
+# even if they don't actually inherit from them. Some so-called "abstract base classes" just check
+# for the presence of certain methods. Cool!
+# However, some of them don't support that, for reasons outlined in their docs (see the first section):
+# https://docs.python.org/3.11/library/collections.abc.html
+
+# thus, we need to make our own checks. These have the same downsides as in the docs,
+# but they're "good enough". "Practicality beats purity".
+
+# we need this for mappings specifically, because a kubernetes.dynamic.resource.ResourceField has
+# __getitem__, which is actually all we care about.
+
+
+def _type_has_getitem(type_: type) -> bool:
+    "See if this type has a getitem method (supports item[key])"
+    return hasattr(type_, "__getitem__")
+
+
+# region: attrs fixes
+
+# type annotations can be represented as the types themselves, or as strings.
+# attrs seems to use strings for fields in inherited classes, which can break
+# things for us. This may be a bug that has to be fixed for them.
+# Until then, here's a small alternative.
+
+
+def _is_attrs_class(cls: type) -> TypeGuard[type["attr.AttrsInstance"]]:
     """
     Return the class if it is an attrs class, else None.
 
     This exists because `attrs.has` is not a proper TypeGuard.
     """
-    return cast("type[attr.AttrsInstance]", cls) if attrs.has(cls) else None
+    return attrs.has(cls)
 
+
+class Field(NamedTuple):
+    """
+    Relevant field information for an attrs class.
+    """
+
+    name: str
+    type: type
+    default: Union[Any, Literal[NOTHING]]
+    metadata: dict[Any, Any]
+
+
+def _fields(cls: type["attr.AttrsInstance"]) -> Iterable[Field]:
+    "Get all field information for this attrs class."
+    hints = get_type_hints(cls)
+    attrs_fields = attrs.fields(cls)
+    for field in attrs_fields:
+        yield Field(
+            field.name,
+            hints[field.name],
+            default=field.default,
+            metadata=field.metadata,
+        )
+
+
+# endregion
 
 # region: generic types
 
@@ -116,13 +178,13 @@ def _is_attrs_class(cls: type) -> Optional[type["attr.AttrsInstance"]]:
 # The origin in this case is `dict`, and the args are `(str, int)`.
 # This lets us inspect these types properly.
 
-# in the future, these should be made into `typing.TypeGuard`s.
-
 
 def _extract_dict_types(type_: type) -> Optional[tuple[type, type]]:
     """
     If this type annotation is a dictionary-like, extract its key and value types.
     Otherwise return None.
+
+    dictionary-like is defined as "has __getitem__ and two type arguments".
 
     >>> assert _extract_dict_types(dict[str, int]) == (str, int)
     >>> assert _extract_dict_types(list[int]) is None
@@ -131,7 +193,9 @@ def _extract_dict_types(type_: type) -> Optional[tuple[type, type]]:
     if origin is None:
         return None
 
-    if not issubclass(origin, MutableMapping):
+    if not _type_has_getitem(type_):
+        return None
+    if len(args) != 2:
         return None
 
     return args[0], args[1]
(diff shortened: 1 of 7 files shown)


exporters/pelorus/deserialization/__init__.py
```diff
@@ -79,54 +83,119 @@
...
 PRIMITIVE_TYPES: tuple[type, ...] = (int, float, str, bool)
 
 
-def nested(path: str) -> dict[str, Any]:
-    "Put the nested data path in a metadata dict."
-    return {_NESTED_PATH_KEY: path}
+def nested(path: Union[str, Sequence[str]]) -> dict[str, Sequence[str]]:
+    """
+    Put the nested data path in a metadata dict.
+    Like get_nested, use a list if a path segment has a dot in it.
```
mateusoliveira43, line 101: since the type is sequence, should we change the "list" word in the docstring?

exporters/pelorus/deserialization/__init__.py
```diff
@@ -79,54 +83,119 @@
...
+# even if they don't actually inherit from them. Some so-called "abstract base classes" just check
+# for the presence of certain methods. Cool!
+# However, some of them don't support that, for reasons outlined in their docs (see the first section):
+# https://docs.python.org/3.11/library/collections.abc.html
+
+# thus, we need to make our own checks. These have the same downsides as in the docs,
+# but they're "good enough". "Practicality beats purity".
+
+# we need this for mappings specifically, because a kubernetes.dynamic.resource.ResourceField has
+# __getitem__, which is actually all we care about.
```
mateusoliveira43, line 123: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#standard-duck-types type them as sequence would fix the problem? (kubernetes.dynamic.resource.ResourceField should have len as well?)

exporters/pelorus/deserialization/__init__.py
```diff
@@ -79,54 +83,119 @@
...
+
+# region: attrs fixes
+
+# type annotations can be represented as the types themselves, or as strings.
+# attrs seems to use strings for fields in inherited classes, which can break
+# things for us. This may be a bug that has to be fixed for them.
+# Until then, here's a small alternative.
+
+
+def _is_attrs_class(cls: type) -> TypeGuard[type["attr.AttrsInstance"]]:
```
mateusoliveira43, line 139: did not get the idea for this function

KevinMGranger: Discussed in meeting :+1:

exporters/pelorus/deserialization/__init__.py
KevinMGranger: Discussed in meeting :+1:

exporters/pelorus/deserialization/__init__.py
KevinMGranger: Changed in a dev docs branch, PR coming soon!

exporters/pelorus/deserialization/__init__.py
mateusoliveira43: TypeGurad is only available in Python 3.10, but are we not accepting 3.9 as well in our code? from https://github.com/konveyor/pelorus/blob/master/exporters/setup.py (and we have a badge in docs https://pelorus.readthedocs.io/en/latest/)

exporters/requirements.txt
```diff
@@ -3,6 +3,7 @@ openshift == 0.13.1    # module openshift
 prometheus_client      # module prometheus_client
 attrs                  # module attrs
 exceptiongroup         # module exceptiongroup
+typing-extensions      # module typing_extensions
```
mateusoliveira43, line 6: what you think of using [Poetry](https://python-poetry.org/) (or other similar tool) for managing the project's dependencies (and fixing the versions)? (but better as separate PR)

exporters/pelorus/deserialization/__init__.py
```diff
@@ -79,54 +83,119 @@
...
     If this type annotation is a dictionary-like, extract its key and value types.
     Otherwise return None.
 
+    dictionary-like is defined as "has __getitem__ and two type arguments".
+
     >>> assert _extract_dict_types(dict[str, int]) == (str, int)
     >>> assert _extract_dict_types(list[int]) is None
     """
     origin, args = typing.get_origin(type_), typing.get_args(type_)
     if origin is None:
```
mateusoliveira43, line 193: can it be efficient to do just one `if`? with something like
```python
if all(origin is not None, _type_has_getitem(type_), len(args) == 2):
    return args[0], args[1]
return None
```

exporters/pelorus/deserialization/__init__.py
```diff
@@ -311,14 +387,12 @@ def _deserialize_field(self, src: Mapping[str, Any], field: attrs.Attribute):
 
     def _deserialize_attrs_class(self, src: Mapping[str, Any], to: type[T]) -> T:
         "Initialize an attrs class field-by-field."
-        cls = _is_attrs_class(to)
-        assert cls is not None, f"class was not an attrs class: {to.__name__}"
+        assert _is_attrs_class(to), f"class was not an attrs class: {to.__name__}"
```
mateusoliveira43, line 390: should we change the assert here? since the compiler can ignore it in an optimized build

exporters/pelorus/deserialization/__init__.py
```diff
@@ -414,9 +500,15 @@ def _deserialize_list(
...
+                unstructured_src
+                if isinstance(unstructured_src, str)
+                else format_path(unstructured_src)
+            )
             raise DeserializationErrors(
                 errors,
-                src_name=self.unstructured_data_path[-1],
+                src_name=src_name,
                 target_name=self.structured_field_name_path[-1],
             )
```
mateusoliveira43, line 514: I would separate this file in small ones (or just another one) and just have an `__all__` here, just because this is a `__init__` file

exporters/tests/test_deserialization.py
```diff
@@ -58,6 +60,16 @@ class Nested:
     assert actual.nested_int == 2
 
 
+def test_multi_nested():
+    @define
+    class NestedNested:
+        nested_int: int = field(metadata=nested(["foo", "com.example.int"]))
```
mateusoliveira43, line 66: I think I did not understand how this work. changing to `["foo", "com", "example.int"]` should not work as well?

exporters/pelorus/utils/nested.py
```diff
@@ -0,0 +1,191 @@
...
+        return tuple(part for part in path.split(".") if part)
+    else:
+        return path
+
+
+def format_path(path: Sequence[str]) -> str:
+    """
+    Format a path for readability.
+    >>> assert format_path("foo.bar".split(".")) == "foo.bar"
+    >>> assert format_path(["foo", "has.dots", "bar"]) == "foo[has.dots].bar"
```
mateusoliveira43, line 131: this doctests are breaking. Do we run them in the CI?

exporters/requirements.txt
KevinMGranger: The last time I checked, `poetry` is not supported by the `ubi8-python` image we use.

We also decided a while back that we _wanted_ to have dependencies upgrade on their own, to make sure we can keep up with security releases without manual intervention.

I also have other criticisms of poetry, but that's a discussion for another time.

exporters/pelorus/deserialization/__init__.py
KevinMGranger: Good point, I could do `origin is not None and _type_has_getitem(type_) and len(args) == 2`. I've never seen `all` used this way before, but... I don't dislike it, actually 🤔 

That might be for a later "cleanup" because for every code change, we have to wait for CI and whatnot.

exporters/pelorus/deserialization/__init__.py
KevinMGranger: I don't think we use optimizations. And this is more for peace of mind-- if I've written the code right, this case should never happen anyway.

exporters/pelorus/deserialization/__init__.py
KevinMGranger: The only reason this is an `__init__` instead of just `deserialization.py` is to put errors somewhere else, just to keep things a little more manageable. I don't know if `__init__` needs to just be re-exports. [Even the python stdlib has meaningful code in `__init__` sometimes!](https://github.com/python/cpython/blob/702a5bc4637c82dc011e98b84f0cede98eb08dda/Lib/json/__init__.py)

exporters/tests/test_deserialization.py
KevinMGranger: That would mean something different!

I should add a doctest-style example to `get_nested` and `nested`, but here's the short version:

```python
x: SomeDeeplyNestedDict
assert get_nested(x, "foo.com.example.int") == x["foo"]["com"]["example"]["int"]
assert get_nested(x, ["foo", "com.example.int"]) == x["foo"]["com.example.int"]
```

The first form is a convenient way to replace what we used to have, something like `build_config.spec.source.git.uri`, with `get_nested(build_config, "spec.source.git.uri")`. But when looking at labels and annotations in OpenShift, they commonly have dots in them-- and users can configure what they are, too. So we need the other form to handle that.

exporters/pelorus/utils/nested.py
mateusoliveira43: since there are some, I would remove them from the doctrings and add to the tests

KevinMGranger: Fixed in the architecture docs PR

exporters/requirements.txt
KevinMGranger: We should discuss this at sync. And we should document this in our architecture docs!

mpryc:
/lgtm
please resolve all @mateusoliveira43 comments before merging.

PR #750, opened 2022-12-16
Pelorus community operator fixes

mpryc:
A number of fixes in this revision:
 - Changed logo
 - Added descriptions and fields to the openAPIV3Schema
 - Updated example to include all 3 exporters
 - added "enabled:" field for the exporter to allow enabling or disabling exporter from the values.yaml or the openAPIV3Schema
 - Added metadata and categories for the Marketplace

## Testing Instructions

`operator-sdk run bundle quay.io/migtools/pelorus-operator-bundle:v0.0.38 --namespace pelorus`

@redhat-cop/mdt

### DIFF
--- a/pelorus-operator/bundle/manifests/charts.pelorus.konveyor.io_pelorus.yaml
+++ b/pelorus-operator/bundle/manifests/charts.pelorus.konveyor.io_pelorus.yaml
@@ -15,7 +15,7 @@ spec:
   - name: v1alpha1
     schema:
       openAPIV3Schema:
-        description: Pelorus is the Schema for the pelorus API
+        description: Configure a running instance of Pelorus
         properties:
           apiVersion:
             description: 'APIVersion defines the versioned schema of this representation
@@ -23,16 +23,48 @@ spec:
               internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this
+            description: |-
+              Kind is a string value representing the REST resource this
               object represents. Servers may infer this from the endpoint the client
-              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
             type: string
           metadata:
             type: object
           spec:
-            description: Spec defines the desired state of Pelorus
+            description: 'PelorusSpec defines state of the Pelorus application and
+              allows to configure for the desired workflow. More information about
+              Pelorus configuration that can be used in the Spec section is available
+              at: https://pelorus.readthedocs.io/en/latest/Configuration/'
+            properties:
+              exporters:
+                description: |-
+                  References configuration for the Pelorus exporters. More info about exporters configuration: https://pelorus.readthedocs.io/en/latest/Configuration/#configuring-exporters-details
+
+
+                    Example:
+                      instances:
+                        - app_name: < instance_name >
+                          enabled: < won't be created if set to false >
+                          exporter_type: < deploytime, committime, failure >
+                type: object
+                x-kubernetes-preserve-unknown-fields: true
+              prometheus_retention:
+                description: 'Prometheus Retention time. More information: https://pelorus.readthedocs.io/en/latest/Install/#configure-prometheus-retention'
+                type: string
+              prometheus_retention_size:
+                description: 'Prometheus Retention Size. More information: https://pelorus.readthedocs.io/en/latest/Install/#configure-prometheus-retention'
+                type: string
+              prometheus_storage:
+                description: 'Use Prometheus Persistent Volume. More information:
+                  https://pelorus.readthedocs.io/en/latest/Install/#configure-prometheus-persistent-volume-recommended'
+                type: boolean
+              prometheus_storage_pvc_capacity:
+                description: Prometheus Persistent Volume capacity.
+                type: string
+              prometheus_storage_pvc_storageclass:
+                description: Prometheus Persistent Volume storage class to be used.
+                type: string
             type: object
-            x-kubernetes-preserve-unknown-fields: true
           status:
             description: Status defines the observed state of Pelorus
             type: object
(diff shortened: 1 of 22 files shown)


pelorus-operator/bundle/manifests/charts.pelorus.konveyor.io_pelorus.yaml
```diff
@@ -15,24 +15,56 @@ spec:
   - name: v1alpha1
     schema:
       openAPIV3Schema:
-        description: Pelorus is the Schema for the pelorus API
+        description: Pelorus is the Schema for pelorus API to adopt instance to the
```
mateusoliveira43, line 18: Grafana operator's description are like this, but would not be better to have something like "Configure a running instance of Pelorus" (like Prometheus operator)?

pelorus-operator/bundle/manifests/charts.pelorus.konveyor.io_pelorus.yaml
```diff
@@ -15,24 +15,56 @@ spec:
...
               of an object. Servers should convert recognized schemas to the latest
               internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this
+            description: |-
+              Kind is a string value representing the REST resource this
               object represents. Servers may infer this from the endpoint the client
-              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
```
mateusoliveira43, line 30: would be better if we rephrase  to something like "Kind is a string value, **in CamelCase**, representing the REST resource this..."?

mpryc: This one comes from operator-sdk, so you can file issue to their project.

scripts/pelorus-operator-patches/07_spec_description.diff
```diff
@@ -0,0 +1,89 @@
+--- config/crd/bases/charts.pelorus.konveyor.io_pelorus.yaml.original	2022-12-16 11:41:22.683297131 +0100
++++ config/crd/bases/charts.pelorus.konveyor.io_pelorus.yaml	2022-12-16 11:40:22.725983634 +0100
+@@ -15,29 +15,76 @@
+   - name: v1alpha1
+     schema:
+       openAPIV3Schema:
+-        description: Pelorus is the Schema for the pelorus API
++        description: >-
```
mateusoliveira43, line 8: I do not know how the editing of the diff files works, but I would change the `>-`s for `|`s, for breaking lines in yaml, just for standardization

scripts/pelorus-operator-patches/07_spec_description.diff
```diff
@@ -0,0 +1,89 @@
...
+             type: object
+-            x-kubernetes-preserve-unknown-fields: true
++            properties:
++              exporters:
++                description: >-
++                  References configuration for the Pelorus exporters. More
++                  info about exporters configuration:
++                  https://pelorus.readthedocs.io/en/latest/Configuration/#configuring-exporters-details
++
++                    Example:
```
mateusoliveira43, line 55: here I would remove the tab from "examples:" (consequently, remove one from each of the following lines)

scripts/pelorus-operator-patches/07_spec_description.diff
```diff
@@ -0,0 +1,89 @@
...
++                  Prometheus Retention Size. More information:
++                  https://pelorus.readthedocs.io/en/latest/Install/#configure-prometheus-retention
++                type: string
++              prometheus_storage:
++                description: >-
++                  Use Prometheus Persistent Volume. More information:
++                  https://pelorus.readthedocs.io/en/latest/Install/#configure-prometheus-persistent-volume-recommended
++                type: boolean
++              prometheus_storage_pvc_capacity:
++                description: Prometheus Persitent Volume capacity.
```
mateusoliveira43, line 77: TYPO in "persistent"? (missing second S) (also in line 80)

scripts/pelorus-operator-patches/07_spec_description.diff
mateusoliveira43: Instead of the description, could it be a object in another **properties** section?

scripts/pelorus-operator-patches/07_spec_description.diff
mpryc: That is what OpenShift editor gives and I prefer to use ">-" as it's easier to copy-paste from the OCP editor.

scripts/pelorus-operator-patches/07_spec_description.diff
mpryc: That can't be done. We have problem when we want to use properties section in here. The OCP evaluates every entry in properties and it's not matching openAPIV3Schema because it's Helm values.yaml which is in fact json.

I have tried with:
```
              properties:
                exporters:
                  description: >-
                    References configuration for the Pelorus exporters. More
                    info about exporters configuration:
                    https://pelorus.readthedocs.io/en/latest/Configuration/#configuring-exporters-details
                  type: object
                  properties:
                    instances:
                      description: >-
                        Exporter instances configuration. More information:
                        https://pelorus.readthedocs.io/en/latest/Configuration/
                      type: array
                      items:
                        type: object
                      properties:
                        app_name:
                          description: 'Exporter name, must start with "- ".'
                          type: string
                        enabled:
                          description: 'If set to false, exporter won''t be provisioned.'
                          type: boolean
                        env_from_configmaps:
                          description: >-
                            ConfigMap name(s) with configuration used by the
                            exporter instance.
                          type: object
                        env_from_secrets:
                          description: >-
                            Secret name(s) with secrets used by the exporter
                            instance.
                          type: object
                        exporter_type:
                          description: >-
                            Exporter type as in the:
                            https://pelorus.readthedocs.io/en/latest/Configuration/#configuring-exporters-overview
                          type: string
```

But OCP was creating malformed values, because there is no such type that can fit our list (problematic - at the beginning of the line):
```
- app_name: first
- app_name: second
```

scripts/pelorus-operator-patches/07_spec_description.diff
mpryc: Fixed

scripts/pelorus-operator-patches/07_spec_description.diff
mateusoliveira43: alright, no problem

scripts/pelorus-operator-patches/07_spec_description.diff
mateusoliveira43: :( alright, I think the way you did is fine, and in the docs it is well formatted as well

mateusoliveira43 (dismissed):

scripts/pelorus-operator-patches/07_spec_description.diff
mpryc: We can always find better solution in separate PR

mpryc: Resolving as it's now tracked as an issue

weshayutin (approved):

PR #752, opened 2022-12-16
Pelorus Operator v0.0.1 - Community Operator candidate

mpryc:
Pelorus Operator which is the candidate for the Community Operator

To install:

```operator-sdk run bundle quay.io/pelorus/pelorus-operator-bundle:v0.0.1 --namespace pelorus```

@redhat-cop/mdt

### DIFF
--- a/scripts/create_pelorus_operator
+++ b/scripts/create_pelorus_operator
@@ -16,7 +16,7 @@
 #
 
 OPERATOR_PROJECT_NAME=pelorus-operator
-OPERATOR_ORG_NAME=migtools
+OPERATOR_ORG_NAME=pelorus
 OPERATOR_PROJECT_DOMAIN=pelorus.konveyor.io
 
 # Get the full absolute path to the script. Needed while calling the script
@@ -31,9 +31,6 @@ DEFAULT_PELORUS_CHARTS_DIR="${SCRIPT_DIR}/${DEFAULT_PELORUS_CHARTS_SUBDIR}/"
 DEFAULT_PELORUS_OPERATOR_DIR="${SCRIPT_DIR}/../pelorus-operator/"
 DEFAULT_PELORUS_OPERATOR_PATCHES_DIR="${SCRIPT_DIR}/pelorus-operator-patches/"
 
-# Needed for symlink inside pelorus-operator folder
-DEFAULT_OPERATOR_HELM_CHART_SUBDIR="helm-charts/pelorus"
-
 function print_help() {
     printf "\nUsage: %s [OPTION]\n\n" "$0"
     printf "\tStartup:\n"
@@ -42,17 +39,19 @@ function print_help() {
     printf "\t  -s\tpath to source pelorus helm charts folder, default: %s\n" "${DEFAULT_PELORUS_CHARTS_DIR}"
     printf "\t  -d\tpath to EMPTY destination folder, default: %s\n" "${DEFAULT_PELORUS_OPERATOR_DIR}"
     printf "\t  -p\tpath to DIR with operator patches to be applied, default: %s\n" "${DEFAULT_PELORUS_OPERATOR_PATCHES_DIR}"
+    printf "\t  -v\toperator version to be used in a format x.y.z. If not set z+1 is used.\n"
     exit 0
 }
 
 ### Options
 OPTIND=1
-while getopts "s:d:p:h?" option; do
+while getopts "s:d:p:h?v:" option; do
     case "$option" in
     h|\?) print_help;;
     s)    source_dir=$OPTARG;;
     d)    destination_dir=$OPTARG;;
     p)    patches_dir=$OPTARG;;
+    v)    operator_ver=$OPTARG;;
     esac
 done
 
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: scripts/create_pelorus_operator [hunk 1]
```diff
@@ -32,7 +32,7 @@ DEFAULT_PELORUS_OPERATOR_DIR="${SCRIPT_DIR}/../pelorus-operator/"
 DEFAULT_PELORUS_OPERATOR_PATCHES_DIR="${SCRIPT_DIR}/pelorus-operator-patches/"
 
 # Needed for symlink inside pelorus-operator folder
-DEFAULT_OPERATOR_HELM_CHART_SUBDIR="helm-charts/pelorus"
+# DEFAULT_OPERATOR_HELM_CHART_SUBDIR="helm-charts/pelorus"
```
mateusoliveira43, line 35: why comment this line?
(can we delete it?)

scripts/create_pelorus_operator
```diff
@@ -115,7 +115,10 @@ pushd "${destination_dir}" || exit 2
     # Correct operator version, to be latest +1
     OPERATOR_VERSION=$(curl -H "Authorization: Bearer XYZ" -X GET "https://quay.io/api/v1/repository/${OPERATOR_ORG_NAME}/${OPERATOR_PROJECT_NAME}/tag/" | jq .tags[].name | head -1 | sed -e 's|\"||g')
     echo "INFO: Current operator version: ${OPERATOR_VERSION}"
-    NEW_OPERATOR_VERSION=$(echo ${OPERATOR_VERSION} | awk -F. -v OFS=. '{$NF += 1 ; print}')
+    NEW_OPERATOR_VERSION=$(echo "${OPERATOR_VERSION}" | awk -F. -v OFS=. '{$NF += 1 ; print}')
+    if [ "$NEW_OPERATOR_VERSION" -eq "1" ]; then
```
mateusoliveira43, line 119: did not understand this. The version will not always be in the `major.minor.patch` style?

scripts/create_pelorus_operator
mpryc: It applies to lines 103->113, so we would need to remove all of that, which is marked TBD.

I have commented this out because shellcheck was failing on my local box complaining that the variable is unused.

mateusoliveira43:
More a question then a request: should the operator stuff live in a separated repo?

mpryc:
> More a question then a request: should the operator stuff live in a separated repo?

We did talk about this previously and we may end up moving it to separate repo in the future. Currently it's easier to have it in same repo, because it uses helm-charts from pelorus one.

mateusoliveira43:
Now that we will version it, I believe is important to check if the version is correct (I have already bumped a personal project from 0.0.1 to 0.1.1, for example) . I believe there is a better solution for this, but I use this in one project https://github.com/mateusoliveira43/cly/blob/main/tests/test_version.py#L84

And we should configure an action to automatically updated the operator in the marketplace as well, when version change occurs

I would wait to have these two features before merging this PR

mpryc:
This isn't yet passing the https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/1967

Some labels are missing

scripts/create_pelorus_operator
```diff
@@ -42,17 +39,19 @@ function print_help() {
     printf "\t  -s\tpath to source pelorus helm charts folder, default: %s\n" "${DEFAULT_PELORUS_CHARTS_DIR}"
     printf "\t  -d\tpath to EMPTY destination folder, default: %s\n" "${DEFAULT_PELORUS_OPERATOR_DIR}"
     printf "\t  -p\tpath to DIR with operator patches to be applied, default: %s\n" "${DEFAULT_PELORUS_OPERATOR_PATCHES_DIR}"
+    printf "\t  -v\toperator version to be used in a format x.y.z. If not set z+1 is used.\n"
```
mateusoliveira43, line 42: instead of x, y and z, would be better to use major, minor and patch?

scripts/create_pelorus_operator
```diff
@@ -100,23 +99,19 @@ pushd "${destination_dir}" || exit 2
...
+    if [ -z ${operator_ver+x} ]; then
+    destination_dir="${DEFAULT_PELORUS_OPERATOR_DIR}"
+        # Correct operator version, to be latest +1
+        OPERATOR_VERSION=$(curl -H "Authorization: Bearer XYZ" -X GET "https://quay.io/api/v1/repository/${OPERATOR_ORG_NAME}/${OPERATOR_PROJECT_NAME}/tag/" | jq .tags[].name | head -1 | sed -e 's|\"||g')
+        echo "INFO: Current operator version: ${OPERATOR_VERSION}"
+        NEW_OPERATOR_VERSION=$(echo "${OPERATOR_VERSION}" | awk -F. -v OFS=. '{$NF += 1 ; print}')
+        if [ "$NEW_OPERATOR_VERSION" -eq "1" ]; then
+          NEW_OPERATOR_VERSION=0.0.1
+        fi
+    else
```
mateusoliveira43, line 111: if code enters the else block, does it break line 166? because `OPERATOR_VERSION` was not defined

scripts/create_pelorus_operator
mpryc: I think it's better to stick with x.y.z because SemVer allows to have some other parts then just a non negative number and we really do want to stick with numbers only.

https://semver.org/#spec-item-2

scripts/create_pelorus_operator
mpryc: You are correct and the logic is slightly incorrect. Will fix that.

mpryc: Plus I see line 103 is left over from copy-paste...

scripts/create_pelorus_operator
mpryc: removed commented out block in the newest version of this PR.

scripts/create_pelorus_operator
mpryc: When this line gives error, meaning the operator was not found in this org, then the following line will always result to "1":
```
OPERATOR_VERSION=$(curl -H "Authorization: Bearer XYZ" -X GET "https://quay.io/api/v1/repository/${OPERATOR_ORG_NAME}/${OPERATOR_PROJECT_NAME}/tag/" | jq .tags[].name | head -1 | sed -e 's|\"||g')

# This will result to "1"
NEW_OPERATOR_VERSION=$(echo "${OPERATOR_VERSION}" | awk -F. -v OFS=. '{$NF += 1 ; print}')
```

In the above case we want to have version in the major.minor.patch, so we set the first one as `0.0.1`

mpryc:
The upstream community operator PR (which is copy-paste from the generated bundle folder): https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/1967

scripts/create_pelorus_operator
mateusoliveira43: Instead of passing the version we want (that can lead to versioning errors), should we accept a list of 3 options (major, minor and patch) and apply versioning rules according to the option? (and I would let patch as default like already is)

KevinMGranger:
Looks like this is released?

mpryc:
> Looks like this is released?

Yes, my hope was that this PR got merged prior to https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/1967 but looks like the community operators got first in. If we can get this PR merged that would ensure we are 1-1 with what's in the community-operators-prod. I will propose some process to ensure we always get those two together. The community-operators-prod is at higher priority as there is number of CI jobs already implemented to ensure installation is correct for number of OCP releases.

scripts/create_pelorus_operator
KevinMGranger: If @mateusoliveira43 is satisfied then this looks mergeable!

KevinMGranger (approved):

PR #753, opened 2022-12-19
Add architecture docs

KevinMGranger:


mateusoliveira43:
@KevinMGranger can you update the PR description to link which PR this one depends on? (put `depends on #NUMBER` in it so dpulls job knows about it)

Issue #754, opened 2022-12-19
Consider running doctests


KevinMGranger:
They're useful for small utility functions, and make sure we're not "lying" in the documentation.

Pytest even has some support for them, although I'm not fully familiar with how they do it: https://docs.pytest.org/en/7.1.x/how-to/doctest.html

mpryc:
Adding CI label as I believe this should be covered by the new github action?

KevinMGranger:
Yes, but we can and should integrate it into our unit test strategy also.

PR #757, opened 2022-12-20
fix: add more tests to deserialization

mateusoliveira43:

Add more tests to deserialization module and also running doctests when running the test suite.

## Linked Issues

Resolves https://github.com/konveyor/pelorus/issues/754

## Testing Instructions

run test and coverage command 
```
coverage run -m pytest -rap -m "not integration and not mockoon" && coverage html
```
to check tests and coverage.

### DIFF
--- a/exporters/pelorus/deserialization/__init__.py
+++ b/exporters/pelorus/deserialization/__init__.py
@@ -185,9 +185,6 @@ def _extract_dict_types(type_: type) -> Optional[tuple[type, type]]:
     Otherwise return None.
 
     dictionary-like is defined as "has __getitem__ and two type arguments".
-
-    >>> assert _extract_dict_types(dict[str, int]) == (str, int)
-    >>> assert _extract_dict_types(list[int]) is None
     """
     origin, args = typing.get_origin(type_), typing.get_args(type_)
     if origin is None:
@@ -205,9 +202,6 @@ def _extract_list_type(type_: type) -> Optional[type]:
     """
     If this type annotation is a list-like, extract its value type.
     Otherwise return None.
-
-    >>> assert _extract_list_type(list[int]) == int
-    >>> assert _extract_list_type(dict[str, str]) is None
     """
     origin, args = typing.get_origin(type_), typing.get_args(type_)
     if origin is None:
@@ -223,9 +217,6 @@ def _extract_optional_type(type_: type) -> Optional[type]:
     """
     If this type annotation is Optional[T], then returns the type T.
     Otherwise, returns None.
-
-    >>> assert _extract_optional_type(Optional[int]) == int
-    >>> assert _extract_optional_type(int) is None
     """
     # this is weird because Optional is just an alias for Union,
     # where the other entry is NoneType.
--- a/exporters/pelorus/deserialization/__init__.py
+++ b/exporters/pelorus/deserialization/__init__.py
@@ -185,6 +185,9 @@ def _extract_dict_types(type_: type) -> Optional[tuple[type, type]]:
     Otherwise return None.
 
     dictionary-like is defined as "has __getitem__ and two type arguments".
+
+    >>> assert _extract_dict_types(dict[str, int]) == (str, int)
+    >>> assert _extract_dict_types(list[int]) is None
     """
     origin, args = typing.get_origin(type_), typing.get_args(type_)
     if origin is None:
@@ -202,6 +205,9 @@ def _extract_list_type(type_: type) -> Optional[type]:
     """
     If this type annotation is a list-like, extract its value type.
     Otherwise return None.
+
+    >>> assert _extract_list_type(list[int]) == int
+    >>> assert _extract_list_type(dict[str, str]) is None
     """
     origin, args = typing.get_origin(type_), typing.get_args(type_)
     if origin is None:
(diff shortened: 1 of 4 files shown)


exporters/tests/test_deserialization.py
```diff
@@ -50,10 +115,11 @@ class Missing:
     print(e.value)
 
 
-def test_nested_field_positive():
+@pytest.mark.parametrize("nested_path", ["foo.bar", ("foo", "bar")])
```
KevinMGranger, line 118: :+1:

exporters/pelorus/deserialization/__init__.py
```diff
@@ -223,9 +217,6 @@ def _extract_optional_type(type_: type) -> Optional[type]:
     """
     If this type annotation is Optional[T], then returns the type T.
     Otherwise, returns None.
-
-    >>> assert _extract_optional_type(Optional[int]) == int
-    >>> assert _extract_optional_type(int) is None
```
KevinMGranger, line 228: I don't think we should remove the doctests. They're just as much documentation as they are tests.

What we can look into later is pytest's doctest support: https://docs.pytest.org/en/7.1.x/how-to/doctest.html

exporters/tests/test_deserialization.py
```diff
@@ -9,12 +9,77 @@
...
     nested,
     retain_source,
 )
 
 
+@pytest.mark.parametrize(
+    "type_",
+    [list[int], str, set[str], Iterable[str], Container[str], tuple[float, float]],
+)  # Iterable and container should not have __getitem__, tuple should pass?
+def test_extract_dict_types_from_non_dict(type_: type):
```
KevinMGranger, line 25: I don't think we should test implementation details here. This should be tested as part of the regular testing.

There's this idea of "the testing pyramid", where:
-  integration tests are large, harder to write (and sometimes maintain), but very valuable even for one test.
-  unit tests are small, simple, and easy to maintain, but less valuable individually.

The deserialization unit tests for classes are "bigger" unit tests, and will hit this code path regardless.

exporters/tests/test_deserialization.py
```diff
@@ -9,12 +9,77 @@
...
     deserialize,
     nested,
     retain_source,
 )
 
 
+@pytest.mark.parametrize(
+    "type_",
+    [list[int], str, set[str], Iterable[str], Container[str], tuple[float, float]],
+)  # Iterable and container should not have __getitem__, tuple should pass?
```
KevinMGranger, line 24: I don't know if it's worth it to catch every edge case. This was mainly for working with openshift objects, and I don't know when we'd actually expect a tuple!

KevinMGranger (changes requested):
I like the nested and Any tests you added! But let's discuss or implement my notes about the doctests and what kind of unit tests we should write.

exporters/pelorus/deserialization/__init__.py
mateusoliveira43: I look into it

exporters/tests/test_deserialization.py
mateusoliveira43: I would add a comment about not supporting tuples then

exporters/tests/test_deserialization.py
mateusoliveira43: this one I was trying to hit one of the ifs of the function (And I was only able with set), but I will remove the more unusual types

exporters/pelorus/utils/nested.py
```diff
@@ -151,21 +155,21 @@ class BadAttributePathError(Exception):
...
     path: Sequence[str]
     path_slice: slice
+    key: str
     value: Any
     root_name: Optional[str] = None
 
     @property
     def message(self) -> str:
-        msg = f"{self.root_name + ' is missing' if self.root_name else 'Missing'} {'.'.join(self.path)}"
+        msg = f"{self.root_name + ' is missing' if self.root_name else 'Missing'} {self.key}"
```
mateusoliveira43, line 166: I changed here because the errors message where not meaningful for me

THEN
```
exporters/tests/test_deserialization.py::test_nested_missing Error while deserializing <class 'tests.test_deserialization.test_nested_missing.<locals>.Nested'> from dict:
  nested_int: Missing foo.bar foo.bar because foo was {}


exporters/tests/test_deserialization.py::test_multi_nested_missing Error while deserializing <class 'tests.test_deserialization.test_multi_nested_missing.<locals>.Nested'> from dict:
  nested_int: Missing foo.com.example.int foo.com.example.int because foo was {}
```
NOW
```
exporters/tests/test_deserialization.py::test_nested_missing Error while deserializing <class 'tests.test_deserialization.test_nested_missing.<locals>.Nested'> from dict:
  nested_int: Missing bar in foo.bar because foo was {}

exporters/tests/test_deserialization.py::test_multi_nested_missing Error while deserializing <class 'tests.test_deserialization.test_multi_nested_missing.<locals>.Nested'> from dict:
  nested_int: Missing com in foo.com[example.int] because foo was {}
```

exporters/pelorus/deserialization/__init__.py
mateusoliveira43: added it :)

KevinMGranger:
I'm going to revisit the unit tests part and clean it up myself (perhaps even cleaning up the code as we go along).

/lgtm

KevinMGranger (approved):

PR #760, opened 2023-01-03
Mockoon tests for image collector

KevinMGranger:
and other small mockoon improvements.

Hopefully would catch the issues noticed in #720.

- [x] Open question: is this test sufficient?

The test data came from a basic python tekton demo image, and I made a [small script](https://github.com/KevinMGranger/pelorus-extra/blob/f2f1357cae7ab522171e748b54a1db7d53dc79eb/image.py) to alter the data a bit.

## Testing Instructions

Run `make mockoon-tests`


mpryc:
@KevinMGranger maybe I am missing something, but where do you actually load/use the ```mocks/commitexporter_github.json```
Secondly would it make sense to put those in the ```exporters/tests/data/``` or move the files which are there in your new folder?

KevinMGranger:
They're used by mockoon, not by my code. And yes, I think it would be nicer to have them where you said-- but that's for another pull request.

mocks/commitexporter_github.json
```diff
@@ -1,6 +1,6 @@
 {
```
mateusoliveira43, line 1: how we create/edit this files?

KevinMGranger: Very carefully.

Just kidding! We edit it with [Mockoon](https://mockoon.com/).

We should probably document how we use mockoon for new tests. Here's what we do:

We set it up as an intercepting proxy, and record responses from openshift, github, or other API providers. We can then reuse those responses for mocked integration tests.

Using the GUI is good for manual testing, and is how we set up new data. But for automated testing, you can run it from the CLI. That's what our script does.

mateusoliveira43:
I believe the mockoon jobs are failing... probably the problem that happened for me (using a new version of the container image resolved). I would wait to fix the job before merging the PR (I will focus on this now)

mpryc:
I had to modify the change:

1. point to it's full registry, otherwise was getting error from `podman` run:
    `Error: short-name resolution enforced but cannot prompt without a TTY`
2.  bump the version of mockoon container, otherwise I was getting errors in the mockoon logs:
    ```shell
     ›   Error: These environment's data are too recent and cannot be run with the 
     ›   CLI
     ›   Please update the CLI with the following command 'npm install -g 
     ›   @mockoon/cli'
    ```

The diff:
```diff
index 9c06377..073c801 100755
--- a/scripts/run-mockoon-tests
+++ b/scripts/run-mockoon-tests
@@ -26,7 +26,7 @@
 # removed from the system. 
 
 CONTAINER_ENGINE="podman"
-MOCKOON_CONTAINER="mockoon/cli:1.5.1"
+MOCKOON_CONTAINER="docker.io/mockoon/cli:2.3.0"
 MOCKOON_PORT="3000"
 MOCKOON_TIMEOUT=30
```

Attached is the log from the running script `./scripts/run-mockoon-tests
[mockoon_run.txt](https://github.com/konveyor/pelorus/files/10478589/mockoon_run.txt)
`

I see from the collected `committimeexporter_github.json` the date/time that does have positive test:
 -  `io.openshift.build.commit.date: Tue Jan 03 22:53:46 2023 +0000`
 -  `io.openshift.build.commit.id: 16ef25d531ce169f5ae36990ef7b050920d2a395` 
 -  `io.openshift.build.source-location: https://github.com/kevinmgranger/pelorus`
 - `name": "sha256:9031ad13cdc840f029acd3d000dcf5e1287fc1a6cf47ed8ac4015e7c47a14aae`


The rest have negative tests, like the following image is missing commit date, which has warning in the mockoon run:
- sha256:dd26951cea41468f3b37a0d60d29d22a2a9192f6229af766f0e56daa5c011157

Overall I am happy with the above tests.

KevinMGranger:
Good call! I'll make those changes.

mateusoliveira43:
the bitbucket mockoon test is not running, with the reason it does not work on macos (but I tried, and also breaks on Linux). Should with fix it in this PR?

KevinMGranger:
Let's make that a separate PR. But we should definitely fix it.

KevinMGranger:
@mpryc negative tests are hard to add. I think every exporter just logs when something is missing data, and testing for log line presence is pretty annoying.

We could revisit that if we did structured logging, or if we wanted to add a configurable callback for when things are missing. But that probably needs structured logging to look nice anyway.

mateusoliveira43 (approved):

PR #763, opened 2023-01-04
Fix/tekton demo

mateusoliveira43:

Some small improvements to tekton demo script.

## Linked Issues

No.

## Testing Instructions

Install pelorus operator and create instance and run the script with `./demo/demo-tekton.sh`.

### DIFF
--- a/demo/demo-tekton.sh
+++ b/demo/demo-tekton.sh
@@ -2,28 +2,9 @@
 #Assumes User is logged in to cluster
 set -euo pipefail
 
+# Defaults
 app_name="basic-python-tekton"
 app_namespace="basic-python-tekton"
-
-Help()
-{
-   # Display Help
-   echo "Execute a tekton pipeline with various build types."
-   echo
-   echo "Syntax: scriptTemplate [-h|g|b|n]"
-   echo "options:"
-   echo "a     application name, default ${app_name}"
-   echo "n     namespace to be used, default ${app_namespace}"
-   echo "g     the git url"
-   echo "r     git branch reference, use this for Pull Requests. e.g. refs/pull/587/head"
-   echo "h     Print this Help."
-   echo "b     build type [buildconfig, binary, s2i]"
-   echo "t     Time in seconds to sleep between subsequent deployments, default 300 (enables non-interaction mode)"
-   echo "c     Number of deployments, default 1 (enables non-interaction mode)"
-   echo
-}
-
-# Defaults
 current_branch="$(git symbolic-ref HEAD)"
 current_branch=${current_branch##refs/heads/}
 url=""
@@ -35,6 +16,26 @@ REMOTE_BRANCH_EXISTS=false
 PELORUS_WORKING_DIR=""
 PELORUS_DEMO_TMP_DIR=""
 TMP_DIR_PREFIX="pelorus_tkn_tmp_"
+tekton_setup_dir="$(dirname "${BASH_SOURCE[0]}")/tekton-demo-setup"
+python_example_txt="$(dirname "${BASH_SOURCE[0]}")/python-example/response.txt"
+
+Help()
+{
+   # Display Help
+   echo "Execute a tekton pipeline with various build types."
+   echo
+   echo "Syntax: scriptTemplate [-h|g|b|n]"
+   echo "options:"
+   echo "a     application name, default ${app_name}"
+   echo "n     namespace to be used, default ${app_namespace}"
+   echo "g     the git url"
+   echo "r     git branch reference, use this for Pull Requests. e.g. refs/pull/587/head, default current branch"
+   echo "h     Print this Help."
+   echo "b     build type [buildconfig, binary, s2i], default ${build_type}"
+   echo "t     Time in seconds to sleep between subsequent deployments, default ${sleep_between} (enables non-interaction mode)"
+   echo "c     Number of deployments, default ${no_deployments} (enables non-interaction mode)"
+   echo
+}
 
 # Get the options
 while getopts ":hg:b:r:n:t:c:a:" option; do
@@ -46,7 +47,7 @@ while getopts ":hg:b:r:n:t:c:a:" option; do
          url=$OPTARG;;
       r) # the git ref
          current_branch=$OPTARG;;
-      b) # Enter the build type 
+      b) # Enter the build type
          build_type=$OPTARG;;
       t) # Sleep between subsequent deployments
          sleep_between=$OPTARG && \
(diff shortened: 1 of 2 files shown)


demo/demo-tekton.sh
```diff
@@ -307,7 +343,7 @@ if [ "${NO_HUMAN}" == false ]; then
           * ) echo "I'm not sure what $a means, please give 1 or 2" >&2 ;;
        esac
     done
-elif [ "${NO_HUMAN}" == true ]; then
+else
```
mpryc, line 346: I would stay explicit here. else means we would need to add comparison to check if the NO_HUMAN var exists and so on...

demo/demo-tekton.sh
```diff
@@ -258,8 +292,10 @@ oc process -f "$tekton_setup_dir/05-build-and-deploy.yaml" -p PROJECT_URI="${GIT
            -p NAMESPACE="${app_namespace}" -p APPLICATION_NAME="${app_name}" \
            -n default > /tmp/05-build-and-deploy.yaml.out 2>/tmp/05-build-and-deploy.yaml.err
 oc apply -n "${app_namespace}" -f /tmp/05-build-and-deploy.yaml.out
+# should delete files from /tmp/?
```
mpryc, line 295: We do this in the function which is in the trap() that is invoked when the script exits:
```
function cleanup_and_exit() {
```

demo/demo-tekton.sh
```diff
@@ -258,8 +292,10 @@ oc process -f "$tekton_setup_dir/05-build-and-deploy.yaml" -p PROJECT_URI="${GIT
...
 
 route=$(oc get -n "${app_namespace}" "route/${app_name}" --output=go-template='http://{{.spec.host}}')
```
mpryc, line 297: No, this is the route to the application that is being built.

demo/demo-tekton.sh
```diff
@@ -221,17 +239,33 @@ printf "Templates: "
...
+# oc process -f "$tekton_setup_dir/04-service-account_template.yaml" -p PROJECT_NAMESPACE="${app_namespace}" -n default | oc delete -f -
+# oc delete -f "$tekton_setup_dir/03-rbac-pipeline-role.yaml"
+# oc delete -f "$tekton_setup_dir/02-tekton-operator.yaml" # not enough :(
+# exit 0
 
 echo "Setting up resources:"
 
 echo "2. Installing tekton operator"
 oc apply -f "$tekton_setup_dir/02-tekton-operator.yaml"
+# I think we need to wait here
```
mpryc, line 268: For what?

demo/demo-tekton.sh
```diff
@@ -221,17 +239,33 @@ printf "Templates: "
 oc delete --all template.template.openshift.io -n "${app_namespace}" 2> /dev/null || echo "...done"
 printf "RBAC Authorization: "
 oc delete "clusterrolebinding.rbac.authorization.k8s.io/pipeline-role-binding-${app_namespace}" 2> /dev/null || echo "...done"
-pv="$(oc get pv | grep "${app_name}" | cut -d " " -f 1)"
-pvc="$(oc get pvc -n "${app_name}" | cut -d " " -f 1 | tail -n 1)"
-printf "Remove PV's and PVC's associated w/ the tekton demo: %s and %s\n" "$pv" "$pvc" 
-oc delete pv "$pv" --grace-period=0 --wait=false || true
-oc delete pvc "$pvc" -n "${app_namespace}" --grace-period=0 --wait=false || true
-oc patch pvc "$pvc" -p '{"metadata":{"finalizers":null}}' -n "${app_namespace}" || true
+if oc get pv | grep "${app_name}"; then
```
mpryc, line 242: We can be smarter here to explicitly get the pv with known name?

demo/demo-tekton.sh
```diff
@@ -221,17 +239,33 @@ printf "Templates: "
...
+    pv="$(oc get pv | grep "${app_name}" | cut -d " " -f 1)"
+    printf "Remove PV's associated w/ the tekton demo: %s\n" "$pv"
+    oc delete pv "$pv" --grace-period=0 --wait=false || true
+else
+    echo "No PV's to remove"
```
mpryc, line 247: Is it warning or error? If second should the script exit with error ?

demo/demo-tekton.sh
```diff
@@ -221,17 +239,33 @@ printf "Templates: "
...
+fi
+app_pvcs=$(oc get pvc -n "${app_name}" 2>&1)
+if [[ $app_pvcs == "No resources found in ${app_name} namespace." ]]; then
+    echo "No PVC's to remove"
+else
+    pvc="$(oc get pvc -n "${app_name}" | cut -d " " -f 1 | tail -n 1)"
+    printf "Remove PVC's associated w/ the tekton demo: %s\n" "$pvc"
+    oc delete pvc "$pvc" -n "${app_namespace}" --grace-period=0 --wait=false || true
+    oc patch pvc "$pvc" -p '{"metadata":{"finalizers":null}}' -n "${app_namespace}" || true
```
mpryc, line 256: After it's removed we could check if they are really gone from the cluster. We are manipulating finalizers here so we want to be sure it's clean.

demo/demo-tekton.sh
mateusoliveira43: But its defined in the defaults (so it is always available), and the user can not change its value (only the script changes it to true). Do you still prefer to be explicit?

demo/demo-tekton.sh
mateusoliveira43: It did not remove neither `/tmp/05-build-and-deploy.yaml.out` or `/tmp/05-build-and-deploy.yaml.err` to me in all my runs. Can you check if it deletes for you?

demo/demo-tekton.sh
mateusoliveira43: I believe tekton operator does build immediately. The first time I run with the update 02 file, it still broke, then it passed in the second time. I suspect because it was still building

demo/demo-tekton.sh
mateusoliveira43: I do not know, I just modified here to stop breaking for me

demo/demo-tekton.sh
mateusoliveira43: I think it is just a clean up before start, I would not exit. If we exit, first time someone runs the scripts, it will always fail, no?

demo/demo-tekton.sh
mateusoliveira43: I think it would be nice, but I do not know how to do this :grimacing:

demo/demo-tekton.sh
mateusoliveira43: But shouldn't we print Grafana route too? Since its a demo, a believe the most important part is to show the dashboards to the user

demo/tekton-demo-setup/02-tekton-operator.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: openshift-pipelines-operator
   namespace: openshift-operators
 spec:
-  channel: stable
-  name: openshift-pipelines-operator-rh 
-  source: redhat-operators 
-  sourceNamespace: openshift-marketplace 
+  channel: latest
```
mateusoliveira43, line 7: @mpryc did you test if it breaks for you too the way it was?

mateusoliveira43:
There is no citation of tekton demo in our docs, right? (there is the the `demo/README.md` only, I think) Should we add a citation there?

demo/demo-tekton.sh
```diff
@@ -221,17 +239,33 @@ printf "Templates: "
 oc delete --all template.template.openshift.io -n "${app_namespace}" 2> /dev/null || echo "...done"
```
mateusoliveira43, line 239: first step (clean up prior to execution) does not delete everything `oc get all --namespace basic-python-tekton`
```
        NAME                          TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
        service/basic-python-tekton   ClusterIP   172.30.55.112   <none>        8080/TCP   9m33s

        NAME                                           HOST/PORT                                                                                        PATH   SERVICES              PORT    TERMINATION   WILDCARD
        route.route.openshift.io/basic-python-tekton   basic-python-tekton-basic-python-tekton.apps.cluster-wdh11092022b.wdh11092022b.mg.dog8code.com          basic-python-tekton   <all>                 None
```
does this matter?

demo/demo-tekton.sh
```diff
@@ -221,17 +239,33 @@ printf "Templates: "
...
+    pvc="$(oc get pvc -n "${app_name}" | cut -d " " -f 1 | tail -n 1)"
+    printf "Remove PVC's associated w/ the tekton demo: %s\n" "$pvc"
+    oc delete pvc "$pvc" -n "${app_namespace}" --grace-period=0 --wait=false || true
+    oc patch pvc "$pvc" -p '{"metadata":{"finalizers":null}}' -n "${app_namespace}" || true
+fi
+# oc delete -n "${app_namespace}" -f /tmp/05-build-and-deploy.yaml.out
+# oc process -f "$tekton_setup_dir/04-service-account_template.yaml" -p PROJECT_NAMESPACE="${app_namespace}" -n default | oc delete -f -
+# oc delete -f "$tekton_setup_dir/03-rbac-pipeline-role.yaml"
+# oc delete -f "$tekton_setup_dir/02-tekton-operator.yaml" # not enough :(
+# exit 0
```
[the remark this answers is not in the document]
mpryc: I don't think we should ever delete tekton resources on the cluster after running it, maybe as another flag to the script?

Think of use case:
 I want to run demo and then review pods/objects created by the pipeline or show the dashboard to someone. So I ran the script once and the pelorus has data provided by the demo.

After I'm done I will clean everything up.

demo/demo-tekton.sh
mpryc: It's demo so to show the application is up and running purpose only.

demo/demo-tekton.sh
weshayutin: There is a difference between PV's and PVC's  and they both need to be cleared.   I'll update this.

demo/demo-tekton.sh
weshayutin: I'll review the pv / pvc cleanup and change as needed

demo/demo-tekton.sh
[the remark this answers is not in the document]
weshayutin: correct.. we want to leave the pipeline in place w/ the execution history.

weshayutin:
good news :)
recent:///895cd2d4b98659d82f789eea63bde440

weshayutin (approved):

PR #766, opened 2023-01-05
fix: mockoon script failing silently

mateusoliveira43:

Remove the behavior of silently fail from the mockoon script.

## Linked Issues

No.

## Testing Instructions

Check CI logs for the mockoon job and run `make mockoon-tests` locally.

### DIFF
--- a/.github/workflows/mockoon-tests.yml
+++ b/.github/workflows/mockoon-tests.yml
@@ -7,6 +7,7 @@ on:
       - '!mocks/README.md'
       - '.github/workflows/mockoon-tests.yml'
       - 'Makefile'
+      - 'scripts/run-mockoon-tests'
 
   pull_request:
     paths:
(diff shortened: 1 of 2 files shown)


mpryc:
@mateusoliveira43 what is the delete_me.txt file for?

mateusoliveira43:
> @mateusoliveira43 what is the delete_me.txt file for?

Just to trigger the CI heheheeh I was planning to ask Kevin to delete it in his PR, but now thinking, I believe is better to change the job yaml file to also trigger when there is changes to the mockoon script. What you think?

mateusoliveira43:
@KevinMGranger can you check if this ok so we can review your mockoon PR?

mpryc:
This works, however there is small issue which can be addressed in separate PR.

Running the `make mockoon-tests` without prior running `make dev-env` will cause the mockoon tests to fail.

PR #781, opened 2023-01-16
bump python version check to 3.11

weshayutin:
checked the following:
* unit tests
* integration tests
* lint

seems to work :)

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10']
+        python-version: ['3.9', '3.10', '3.11']
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/scripts/python-version-check.py
+++ b/scripts/python-version-check.py
@@ -16,7 +16,7 @@
 from sys import executable, exit, version_info
 
 PYTHON_VER_MIN = (3, 9)
-PYTHON_VER_MAX = (3, 10)
+PYTHON_VER_MAX = (3, 11)
 
 SYS_PYTHON_VER = (version_info.major, version_info.minor)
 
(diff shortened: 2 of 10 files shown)


.github/workflows/unittests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10']
+        python-version: ['3.9', '3.10', '3.11']
```
mateusoliveira43, line 23: there are other files (like `python-linting.yml`, `python-formatting.yml`, ...) with 3.9 and 3.10. Should we also update them?

And there are some files (like `shellcheck.yaml` and `conftest.yml`) that just have one python version. Should we update them to be 3.11 (because is faster)? If so, I would change the jobs their from  matrix to explicit, like this https://github.com/mateusoliveira43/python-project-template/blob/main/.github/workflows/cd.yml

scripts/python-version-check.py
```diff
@@ -16,7 +16,7 @@
 from sys import executable, exit, version_info
 
 PYTHON_VER_MIN = (3, 9)
-PYTHON_VER_MAX = (3, 10)
+PYTHON_VER_MAX = (3, 11)
```
mateusoliveira43, line 19: Should we update `/docs/Development.md `line 147?

"After cloning the repo, you'll need a python version that's >= 3.9 **but < 3.11**."

.github/workflows/unittests.yml
weshayutin: fixed, thanks

scripts/python-version-check.py
weshayutin: updated.. thanks

mateusoliveira43 (approved):

etsauer:
Just tested and lgtm as well.

PR #784, opened 2023-01-18
New layout for the exporter docs

mpryc:
Current layout for the Pelorus Exporters docs causes the page to be very long. Split into per exporter configuration make it easier to focus on a particular part.

Also we don't need to inform user about different exporter installation paths (source code, quay images, hashes etc). This is more of an development doc, so moving that part to the Development section.

@redhat-cop/mdt


docs/Development.md
```diff
@@ -140,6 +140,125 @@ The following is a recommended directory structure for a Pelorus exporter `<NAME
...
+  - app_name: failure-github
+    exporter_type: deploytime
+    image_tag: d6f6e6fa1c9d48ca1deeaf1c72585b94964cbf31 # Specific merge build
+    env_from_secrets:
+    - github-credentials
+    env_from_configmaps:
+    - pelorus-config
+    - my-failure-github-config
+
+- app_name: jira-failure-exporter
```
mateusoliveira43, line 191: wrong indentation

mkdocs.yml
```diff
@@ -55,7 +55,11 @@ nav:
     - Installation: GettingStarted/Installation.md
     - Configuration:
       - Pelorus Core: GettingStarted/configuration/PelorusCore.md
-      - Pelorus Exporters: GettingStarted/configuration/PelorusExporters.md
+      - Pelorus Exporters:
+        - Overview: GettingStarted/configuration/PelorusExporters.md
+        - Deploy Time: GettingStarted/configuration/ExporterDeploytime.md
+        - Commit Time: GettingStarted/configuration/ExporterCommittime.md
+        - Failure Time: GettingStarted/configuration/ExporterFailure.md
```
mateusoliveira43, line 62: it is **failure time** or just **failure**?

mpryc: In the docs we have elsewhere "Failure Time Exporter", e.g.:

"The job of the failure time exporter is to capture the timestamp at which a failure occurs in a production environment and when it is resolved."

"Failure Time Exporter may be deployed with one of three backends, such as JIRA, GitHub Issues"

So I think it should be failure time ?

mateusoliveira43 (approved):

PR #792, opened 2023-01-19
feat: Automatically add replaces/skips to operator csv file

### LINKED ISSUE #788: Adjust create_pelorus_operator to include replaces and/or skips in our CSV
Currently lemon/orange CI jobs (see links below of the description of those jobs) are failing. We need to include `replaces` and/or `skips` in our CSV for the pelorus-operator, meaning our `create_pelorus_operator` script needs to have logic to create those sections dynamically.

Couple of links:
https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/

https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/how-to-update-operators.md#replaces--channels

https://operator-framework.github.io/community-operators/tests-in-pr/

https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/2083

https://github.com/konveyor/pelorus/pull/787

Example of failed job without replaces section:
https://github.com/redhat-openshift-ecosystem/community-operators-prod/actions/runs/3953180125/jobs/6769206949

mateusoliveira43:

Automatically add replaces/skips sections to Pelorus Operator ClusterServiceVersion file when updating/creating the  Operator.

## Linked Issues

resolves #788 

## Testing Instructions

run `rm -rf test_folder && mkdir test_folder && scripts/create_pelorus_operator -d test_folder` and check if the `test_folder/bundle/manifests/pelorus-operator.clusterserviceversion.yaml` has the desired sections from pelorus operator tags.

### DIFF
--- a/scripts/specify_operator_update.py
+++ b/scripts/specify_operator_update.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+"""
+Script to add replaces and skips sections to Pelorus Operator
+ClusterServiceVersion file, located in
+
+../pelorus-operator/bundle/manifests/pelorus-operator.clusterserviceversion.yaml
+
+More info: https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/
+"""
+
+import argparse
+import sys
+from pathlib import Path
+from typing import List
+
+YAML_TAB = "  "
+
+
+def path_type(file_path: str) -> Path:
+    _file_path = Path(file_path).resolve()
+    if _file_path.exists():
+        return _file_path
+    raise argparse.ArgumentTypeError(f"Folder {_file_path} does not exist.")
+
+
+def exit_error(message: str) -> None:
+    print(f"ERROR: {message}", file=sys.stderr)
+    raise SystemExit(1)
+
+
+def main(
+    new_version: str, cluster_service_version_file: Path, versions: List[str]
+) -> None:
+    if new_version in versions:
+        exit_error(f"version {new_version} already exists in Pelorus Operator tags.")
+
+    with open(cluster_service_version_file, encoding="utf-8") as file_content:
+        cluster_service_version_file_content = file_content.readlines()
+
+    cluster_service_version_file_content.append(
+        f"{YAML_TAB}replaces: pelorus-operator.v{versions[0]}\n"
+    )
+    cluster_service_version_file_content.append(f"{YAML_TAB}skips:\n")
+    for version in versions:
+        cluster_service_version_file_content.append(
+            f"{YAML_TAB*2}- pelorus-operator.v{version}\n"
+        )
+
+    with open(cluster_service_version_file, mode="w", encoding="utf-8") as file_content:
+        file_content.writelines(cluster_service_version_file_content)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Updates ClusterServiceVersion of new Pelorus Operator."
+    )
+    parser.add_argument(
+        "new_version", type=str, help="New version of Pelorus Operator."
+    )
+    parser.add_argument(
+        "destination_path", type=path_type, help="Folder of Pelorus Operator."
+    )
+    parser.add_argument(
+        "versions", nargs="*", help="Pelorus Operator versions, filtered and sorted."
+    )
+
+    cluster_service_version_file = (
+        Path(parser.parse_args().destination_path).resolve()
+        / "bundle/manifests/pelorus-operator.clusterserviceversion.yaml"
+    )
+    if not cluster_service_version_file.exists():
+        exit_error(f"{cluster_service_version_file} does not exist.")
+
+    main(
+        parser.parse_args().new_version,
+        cluster_service_version_file,
+        parser.parse_args().versions,
+    )
--- a/scripts/create_pelorus_operator
+++ b/scripts/create_pelorus_operator
@@ -100,7 +100,8 @@ pushd "${destination_dir}" || exit 2
     operator-sdk create api --helm-chart="${source_dir}" || exit 2
 
     # Correct operator version, to be latest +1
-    OPERATOR_VERSION=$(curl -H "Authorization: Bearer XYZ" -X GET "https://quay.io/api/v1/repository/${OPERATOR_ORG_NAME}/${OPERATOR_PROJECT_NAME}/tag/" | jq .tags[].name | head -1 | sed -e 's|\"||g')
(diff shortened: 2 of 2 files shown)


scripts/operator-versions.json
```diff
@@ -0,0 +1 @@
+{"versions": ["0.0.1", "0.0.2"]}
```
mateusoliveira43, line 1: I believe this is the simplest/easiest way, but if this information is on the openshift catalog (or any place in the web), I think is better to grab it from there

mateusoliveira43: alright, I will get that info from there, instead of adding in a json file

mateusoliveira43: alright, I could filter with `jq`: `".tags[] | select(.end_ts == null) | .name"` but why there multiple tags?

mpryc: I suspect the multiple tags are in the API to ensure consistency of the images when they are getting deleted. Also probably for the history of what happened with the images in the registry.

mateusoliveira43: when what about the "dirty" tags: '0.0.1nowatchers', '0.0.1fix777nowatchers', '0.0.1fix777' ?

mateusoliveira43: ok, ignored them

mpryc: I can confirm that:

```diff
@@ -460,7 +460,8 @@ spec:
   provider:
     name: Red Hat
     url: https://redhat.com
-  version: 0.0.2
-  replaces: pelorus-operator.v0.0.1
-  skips:
-    - pelorus-operator.v0.0.1
+  version: 0.0.3
+  replaces: pelorus-operator.v0.0.2
+    skips:
+      - pelorus-operator.v0.0.2
+      - pelorus-operator.v0.0.1
```

mateusoliveira43: oh wait, I think I putted wrong indentation

mpryc (changes requested):
There is a traceback:

I have created pelorus-operator repository with extra tag: "ONE" and also created tags, which are not in order :
```shell
$ curl -H "Authorization: Bearer XYZ" -X GET "https://quay.io/api/v1/repository/migi/pelorus-operator/tag/" | jq ".tags[] | select(.end_ts == null) | .name" | sed -e 's|\"||g'
0.0.5
0.0.6
ONE
0.0.4
0.0.3
0.0.1
```

See that `0.0.5` was created after `0.0.6`, so it appears first. I know the previous logic was `head -1`, however we may want to adjust that in this PR?

Also what about use case where one of the tag is missing or major bump happened, e.g.:

```
1.0.1
0.0.9
0.0.5
0.0.6
ONE
0.0.4
0.0.3
0.0.1
```


The traceback, where I pointed in the `./scripts/create_pelorus_operator` to the `migi` org:

```diff
diff --git a/scripts/create_pelorus_operator b/scripts/create_pelorus_operator
index 93ca65e..7a483d7 100755
--- a/scripts/create_pelorus_operator
+++ b/scripts/create_pelorus_operator
@@ -16,7 +16,7 @@
 #
 
 OPERATOR_PROJECT_NAME=pelorus-operator
-OPERATOR_ORG_NAME=pelorus
+OPERATOR_ORG_NAME=migi
 OPERATOR_PROJECT_DOMAIN=pelorus.konveyor.io
 
 # Get the full absolute path to the script. Needed while calling the script
```
Output
```shell
[...]
INFO: Adding replaces and skips entries to /home/migi/Development/Upstream/mpryc-pelorus/scripts/../charts/pelorus//home/migi/Development/Upstream/mpryc-pelorus/scripts/../pelorus-operator//bundle/manifests/charts.pelorus.konveyor.io_pelorus.yaml
  File "/home/migi/Development/Upstream/mpryc-pelorus/scripts/specify_operator_update.py", line 75, in <module>
    main(
  File "/home/migi/Development/Upstream/mpryc-pelorus/scripts/specify_operator_update.py", line 34, in main
    parsed_versions = [
  File "/home/migi/Development/Upstream/mpryc-pelorus/scripts/specify_operator_update.py", line 35, in <listcomp>
    version for version in versions if version.split(".", maxsplit=2)[2].isdigit()
IndexError: list index out of range

**Comment by mateusoliveira43** [comment]
Yeah, we can order the list

but I would not check here if the versions are in the correct format, this has to be done before sending them

File: scripts/specify_operator_update.py [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+def exit_error(message: str) -> None:
+    print(f"ERROR: {message}", file=sys.stderr)
+    raise SystemExit(1)
+
+
+def main(
+    new_version: str, cluster_service_version_file: Path, versions: List[str]
+) -> None:
+    parsed_versions = [
+        version
```
mpryc, line 35: Should we use here:

https://python-semver.readthedocs.io/en/stable/usage.html#checking-for-a-valid-semver-version

+ check that `prerelease` and `build` are None:

https://python-semver.readthedocs.io/en/stable/usage.html#accessing-parts-of-a-version-through-names

Then we can use max() to find out the newest version...

mpryc: This way we can ignore `head -1` and use your script to parse output versions ?

mateusoliveira43: the script is called at the end of the create script, it would had to be another one in my opinion. But calling two python scripts in the middle of a bash one, does not seem good, it is better to write everything in Python (and it would be much easier to filter the tags with Python, like the example you gave, I had to use regex with bash)

mpryc: @mateusoliveira43 now it's really working great. One corner case which I found and don't know if it's worth updating at this point:

If we want to re-spin (recreate) already existing tag, for example during our review process we found that 0.0.2 needs fixing, then we would use:

```shell
./scripts/create_pelorus_operator -v 0.0.2
```

pelorus-operator files are properly generated (as expected), but then at the end we get:
```shell
ERROR: version 0.0.2 already exists in Pelorus Operator tags.
```

So, our spec looks like:
```diff
@@ -461,6 +461,3 @@ spec:
     name: Red Hat
     url: https://redhat.com
   version: 0.0.2
-  replaces: pelorus-operator.v0.0.1
-  skips:
-    - pelorus-operator.v0.0.1
```

I think we should ideally error *before* we generate any files or allow to run the script and generate `skips/replaces` giving at the end `WARNING`, that this tag is already in the repo. This is valid case as I wrote, when someone would like to spin off same version for any reason.

mpryc: Ok I am happy with this approach.

mpryc (approved):
Happy with the current implementation. Tested and properly adds desired fields.

PR #798, opened 2023-01-20
Small format update to the exporter failure doc

mpryc:
This updates the configuration options sections and it's descriptions to the layout/format that is used in other parts of the documentation.

@redhat-cop/mdt

### DIFF
--- a/docs/GettingStarted/configuration/ExporterFailure.md
+++ b/docs/GettingStarted/configuration/ExporterFailure.md
@@ -63,107 +63,105 @@ This is the list of options that can be applied to `env_from_secrets`, `env_from
 | [JIRA_RESOLVED_STATUS](#jira_resolved_status) | no | - |
 | [GITHUB_ISSUE_LABEL](#github_issue_label) | no | bug |
 
-### PROVIDER
+###### PROVIDER
 
 - **Required:** no
     - **Default Value:** jira
 - **Type:** string
 
-Set the Issue Tracker provider for the failure exporter. One of `jira`, `github`, `servicenow`.
+: Set the Issue Tracker provider for the failure exporter. One of `jira`, `github`, `servicenow`.
 
-### LOG_LEVEL
+###### LOG_LEVEL
 
 - **Required:** no
     - **Default Value:** INFO
 - **Type:** string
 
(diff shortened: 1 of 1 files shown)


docs/GettingStarted/configuration/ExporterFailure.md
```diff
@@ -63,107 +63,105 @@ This is the list of options that can be applied to `env_from_secrets`, `env_from
...
 
-### PROVIDER
+###### PROVIDER
 
 - **Required:** no
     - **Default Value:** jira
 - **Type:** string
 
-Set the Issue Tracker provider for the failure exporter. One of `jira`, `github`, `servicenow`.
+: Set the Issue Tracker provider for the failure exporter. One of `jira`, `github`, `servicenow`.
```
mateusoliveira43, line 72: This change is very nice, but I never saw the usage of `:` in markdown syntax. I would use tabs.

mpryc: There is a difference between using:
1.```    Set the Issue Tracker provider for the failure exporter. One of `jira`, `github`, `servicenow`.```
2. ```: Set the Issue Tracker provider for the failure exporter. One of `jira`, `github`, `servicenow`.```

The first image corresponds to (1) and second to (2). I actually prefer (2), because:
- It's easier to identify which part of the doc is shifted/aligned. I don't like counting spaces/tabs and here we would need to have 4 spaces.
- The alignment does gives a bit of gap between previous line so in this case description looks like belongs to `PROVIDER` section and not to the `Type: string`.

This is *valid* Markdown syntax that defines definition list with one item:
https://www.markdownguide.org/extended-syntax/#definition-lists

image

Screenshot from 2023-01-23 13-37-27

mpryc: @mateusoliveira43 we can change this to Tabs, but other exporter docs also uses definition lists, so we want to be same across other docs.

mpryc: Approved, merging, with the new issue to consider changing: https://github.com/konveyor/pelorus/issues/803

PR #819, opened 2023-01-26
docs: Add precedence information to Secret, ConfigMaps and ExtraEnv

### LINKED ISSUE #468: If Secret and ConfigMap is provided, values from ConfigMap are being used
If both secret and configmap is used for any exporter, secret is used at lower priority so, configmap overrides values from secrets which should not happen.

We even have that in the comments:
https://github.com/konveyor/pelorus/blob/master/charts/pelorus/configmaps/committime.yaml#L10-L12

### LINKED ISSUE #802: Document the values priority
I did not had time to test priority

I will add the example

_Originally posted by @mateusoliveira43 in https://github.com/konveyor/pelorus/pull/799#discussion_r1083930888_

mateusoliveira43:

Check the precedence of Secret, ConfigMaps and ExtraEnv, and document it.

## Linked Issues

resolves #468

## Testing Instructions

Create a pelorus instance with the content of `test.yaml`, create secret and configmap with `oc apply -f file`, with the `secret.yaml` and `config.yaml` files, and check the logs of each deployexporter, with `oc logs -n pelorus name` (you can get the name with `oc get pod -n pelorus`).

### DIFF
--- a/secret.yaml
+++ b/secret.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: debug-secret
+  namespace: pelorus
+type: Opaque
+stringData:
+  LOG_LEVEL: "DEBUG"
--- a/test.yaml
+++ b/test.yaml
@@ -1,0 +1,166 @@
+kind: Pelorus
+apiVersion: charts.pelorus.konveyor.io/v1alpha1
+metadata:
+  name: pelorus-test-precedence
+  namespace: pelorus
+spec:
+  exporters:
+    global: {}
+    instances:
+      - app_name: d31
+        exporter_type: deploytime
+        env_from_secrets:
+        - debug-secret
+        env_from_configmaps:
+        - warning-configmap
+        extraEnv:
+          - name: NAMESPACES
+            value: pelorus
+          - name: LOG_LEVEL
+            value: ERROR # won
+
+      - app_name: d32
+        exporter_type: deploytime
+        env_from_secrets:
+        - debug-secret
+        extraEnv:
+          - name: NAMESPACES
+            value: pelorus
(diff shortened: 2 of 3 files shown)


secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: debug-secret
+  namespace: pelorus
+type: Opaque
```
mateusoliveira43, line 6: the type influences in anything?

test.yaml
```diff
@@ -0,0 +1,166 @@
+kind: Pelorus
```
mateusoliveira43, line 1: should we add a test for this?

mateusoliveira43:
add tests to check with 2 (or more) secrets (or configmaps) are passed, which has the precedence, as talked to @mpryc

mpryc:
Ok so in short:

```
extraEnv - casino
secret - 2nd
configmap - looser
```

mpryc:
So it's reversed order to what's here:
  https://github.com/konveyor/pelorus/blob/master/charts/pelorus/subcharts/exporters/templates/_deploymentconfig.yaml#L45-L68

I suspect that when two configmaps are present then the bottom one wins, however I am unsure if the list is properly ordered for the `{{- range }}` helm option: https://github.com/konveyor/pelorus/blob/master/charts/pelorus/subcharts/exporters/templates/_deploymentconfig.yaml#L46

mpryc:
Here is some helm issue where it's non-deterministic... but that's something we need to confirm and possibly fix by having different method or document properly https://github.com/helm/helm/issues/4375

mateusoliveira43:
Closing without merge to do not let it hanging

PR #823, opened 2023-01-27
Deserialization cleanup, test simplification, coverage expansion

KevinMGranger:
related to #757


KevinMGranger:
The new mockoon tests did catch an issue here, but that made me see two things:

- it's hard to debug with the current way we run the mockoon container
- the deserialization framework still doesn't give enough context for certain errors.

I'll address both of those separately.

KevinMGranger:
This is good enough as-is, and more tests can be added later.

KevinMGranger:
@mateusoliveira43 I'm not familiar with the workflow for updating the chart patch versions. But isn't that only supposed to happen on a release anyway?

mateusoliveira43:
@KevinMGranger we changed to apply the change to each PR, ref https://pelorus.readthedocs.io/en/latest/Development/#versioning-process (I think there was a discussion on this, but do not have the link)

on release, we bump `rc` versions to non `rc`

KevinMGranger:
I'm trying to do that, but it can't find the `operator-sdk` binary.

I try to re-run `make cli_dev_tools`, but I get:
```
Found system python version: 3.11
. .venv/bin/activate && \
                ./scripts/install_dev_tools.sh -v .venv
No matching download URL found for noobaa on Darwin arm64
Cleaning up temporary files
```
But it still exits with a 0 status 🤔

mateusoliveira43:
yeah, that is a known issue https://github.com/dora-metrics/pelorus/issues/518

try running script directly with for example `scripts/install_dev_tools.sh -c operator-sdk -v .venv` (I think you need helm and opeartor-sdk for the version script)

KevinMGranger:
```console
$ ./scripts/update_projects_version.py -f -r
INFO:root:Removing destination folder /Users/kevin/workcode/pelorus/pelorus-operator
ERROR:root:You must be logged in to a OpenShift cluster as a user with cluster-admin permissions to run this script. This avoids RBAC problems. More info: https://sdk.operatorframework.io/docs/building-operators/helm/tutorial/#prerequisites
```

I don't see why this is necessary, perhaps we should open an issue for this. I'll get my CRC cluster up and running again I suppose.

mateusoliveira43:
I think there is missing explanation in Operator SDK site, but it is in there prerequisites 

maybe with the PR I am working on https://github.com/dora-metrics/pelorus/pull/1045 we can remove this constrain, but today I believe we need it

PR #824, opened 2023-01-28
fix: Tests config to run doctests and coverage right folders

mateusoliveira43:

Run doctests and coverage only Pelorus python source code (excluding its tests) when running pytest.

Example with `make unit-tests` on master branch
Screenshot from 2023-01-28 13-19-34

Tests are considered part of the coverage, not all folders of exporters are present (like extra) in the coverage and doctests are not run.

Example with `make unit-tests` on this PR
Screenshot from 2023-01-28 13-18-42

Tests are NOT considered part of the coverage, ALL folders of exporters are present in the coverage and doctest are run.

## Linked Issues

No.

## Testing Instructions

Check CI logs and run `make unit-tests`, `make mockoon-tests` and `make integration-tests` to check coverage information. All of them should have the same number of TOTAL Stmts (2183).

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+omit = exporters/tests/*
--- a/Makefile
+++ b/Makefile
@@ -148,12 +148,11 @@ e2e-tests-scenario-2: e2e-tests-dev-env
 	./scripts/run-pelorus-e2e-tests -f "periodic/different_deployment_methods.yaml"
 
 # Integration tests
-## integration-tests: pytest -rap -m integration
+## integration-tests: pytest everything marked as integration
 .PHONY: integration-tests
 integration-tests: $(PELORUS_VENV)
 	. ${PELORUS_VENV}/bin/activate && \
-	coverage run -m pytest -rap -m "integration" && \
-	coverage report
+	pytest -rap -m "integration"
 
 # Unit tests
 ## unit-tests: pytest everything minus integration and mockoon
@@ -163,8 +162,7 @@ unit-tests: $(PELORUS_VENV)
   # because using (A)ll includes stdout
   # -m filters out integration tests
 	. ${PELORUS_VENV}/bin/activate && \
-	coverage run -m pytest -rap -m "not integration and not mockoon" && \
-	coverage report
+	pytest -rap -m "not integration and not mockoon"
 
 # Prometheus ruels
 ## test-prometheusrules: test prometheus with data in _test/test_promethusrules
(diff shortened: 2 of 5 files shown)


.coveragerc
```diff
@@ -0,0 +1,2 @@
+[run]
+omit = exporters/tests/*
```
mateusoliveira43, line 2: had to create this to exclude tests, that are inside exporters folder. I would create an issue to refactor and move tests to outside it, and remove this file

Makefile
```diff
@@ -258,16 +256,6 @@ shellcheck-optional:
 	$(warning 🐚 ⏭ Shellcheck not found, skipping)
 endif
 
-
```
mateusoliveira43, line 261: deleted this because was not being used

exporters/requirements-dev.txt
```diff
@@ -1,9 +1,6 @@
 # Needed by exporters/tests
 pytest
-
-# Mentioned in the docs/Development.md
-# Also used by the github unittests.yml workflow
-coverage
+pytest-cov
```
mateusoliveira43, line 3: to avoid duplication of writing `coverage run -m ... && coverage report` everytime

pyproject.toml
```diff
@@ -4,14 +4,17 @@ src_paths = ["exporters" ,"scripts"]
...
+testpaths = ["exporters"]
 markers = [
     "integration: marks tests as integration tests (deselect with '-m \"not integration\"')",
     "mockoon: tests that should be run with mockoon running"
 ]
 log_cli = true
 addopts = [
     "--doctest-modules",
+    "--cov-report=term-missing",
+    "--cov-report=html",
```
mateusoliveira43, line 16: added html report as well

mpryc: nice

mpryc (approved):

KevinMGranger:
This was confusing 😬 

> Note If you have the pytest-cov coverage module installed, VS Code doesn't stop at breakpoints while debugging because pytest-cov is using the same technique to access the source code being run. To prevent this behavior, include --no-cov in pytestArgs when debugging tests, for example by adding "env": {"PYTEST_ADDOPTS": "--no-cov"} to your debug configuration. (See [Debug Tests](https://code.visualstudio.com/docs/python/testing#_debug-tests) above about how to set up that launch configuration.) (For more information, see [Debuggers and PyCharm](https://pytest-cov.readthedocs.io/en/latest/debuggers.html) in the pytest-cov documentation.)

https://code.visualstudio.com/docs/python/testing#_pytest-configuration-settings

Perhaps we should go back to the old way of doing coverage, and just ensure that's used everywhere?

PR #826, opened 2023-01-30
Fix the create release pr script to include proper regex

mpryc:
Regex for the Install.md should be replaced with regex(es) to the Development.md, because that part of docs was moved.

@redhat-cop/mdt

### DIFF
--- a/scripts/create_release_pr
+++ b/scripts/create_release_pr
@@ -37,7 +37,7 @@ PELORUS_CHART="charts/pelorus/Chart.yaml"
 PELORUS_EXPORTERS_CHART="charts/pelorus/subcharts/exporters/Chart.yaml"
 OPERATORS_CHART="charts/operators/Chart.yaml"
 
-INSTALL_DOC="docs/Install.md"
+INSTALL_DOC="docs/Development.md"
 
 TRUE=1
 FALSE=0
@@ -111,8 +111,10 @@ else
   sed -i "s/^version:.*/version: $NEXT_CHART_VER/g" "$PELORUS_CHART"
 fi
 
-# Update branch in the Install documentation
-sed -i "s/--branch v.* /--branch v$V_MAJOR.$V_MINOR.$V_PATCH /g" "$INSTALL_DOC"
+# Update branch in the Development documentation
+sed -i "s/\(release number, for example \`\).*\(\`\.\)/\1v$V_MAJOR.$V_MINOR.$V_PATCH\2/g" "$INSTALL_DOC"
+sed -i "s/\(image_tag: \).*\( # Specific release\)/\1v$V_MAJOR.$V_MINOR.$V_PATCH\2/g" "$INSTALL_DOC"
+
 
 if [[ $major ]]; then
   printf "\nIMPORTANT:\n\t Do include \"major release\" text in the first line of your commit message, or label your PR with: \"major\"\n\n"


scripts/create_release_pr
```diff
@@ -111,8 +111,10 @@ else
   sed -i "s/^version:.*/version: $NEXT_CHART_VER/g" "$PELORUS_CHART"
 fi
 
-# Update branch in the Install documentation
-sed -i "s/--branch v.* /--branch v$V_MAJOR.$V_MINOR.$V_PATCH /g" "$INSTALL_DOC"
+# Update branch in the Development documentation
+sed -i "s/\(release number, for example \`\).*\(\`\.\)/\1v$V_MAJOR.$V_MINOR.$V_PATCH\2/g" "$INSTALL_DOC"
```
mateusoliveira43, line 115: can we add a flag or something to check if the operation was successful?

mpryc: No, sed does not have control over its exit code. We could in theory check by using additional grep or there is hack to "echo" sort of fake exit code directly from sed, however I don't think it's any use here.

We could always check by running `git diff` or `git status` and see if files were changed, but that's really over-design.

mateusoliveira43: ok

mateusoliveira43 (approved):

mateusoliveira43 (changes requested):
I have two question: when running `make release`

- charts/pelorus/subcharts/exporters/Chart.yaml should be updated?
- charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml `spec.source.git.ref` default updates to chart or pelorus version?

mpryc:
> I have two question: when running `make release`
> 
> * charts/pelorus/subcharts/exporters/Chart.yaml should be updated?
> * charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml `spec.source.git.ref` default updates to chart or pelorus version?

Not always, we have 3 make targets:
```Makefile
release:
        ./scripts/create_release_pr

minor-release:
        ./scripts/create_release_pr -i

major-release:
        ./scripts/create_release_pr -m
```

Let's go over each of them:
```shell
$ make release

# This modifies the files so a new github tag (release) can be created. 
# It goes to the GitHub API and checks what is the last released version x.y.z and bumps the z
# To do so, only main Pelorus Chart requires to be modifed, and the file that references GIT new tag:
# charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml
```

```shell
$ make release

# This modifies the files so a new github tag (release) can be created. 
# It goes to the GitHub API, checks what is the last released version x.y.z and bumps the z
# To do so, only main Pelorus Chart requires to be modifed, and the file that references GIT new tag, so it is
# NOT the Chart version but the github tag:
# charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml
# We can not bump each individual Chart version nor we want as the source code of the individual 
# chart versions may not be the same.
# We want here to always bump z+1 and to sustain this we can not calculate z versions of individual Chart versions
# because they may have been bumped manually.
```

```shell
$ make minor-release

# This goes to the GitHub API, checks what is the last released version x.y.z and bumps the y
# This bumps all Chart versions, so we start from 0 in the z number
# In this case our GitHub release version will match Chart versions and all the numbers should be the same
```

```shell
$ make major-release

# This goes to the GitHub API, checks what is the last released version x.y.z and bumps the x
# It's same as minor-release but X is bumped and y,z are 0'ed
```

mateusoliveira43 (approved):

Issue #825, opened 2023-01-30
Refactor and move tests to outside exporters folder


mpryc:
had to create this to exclude tests, that are inside exporters folder. I would create an issue to refactor and move tests to outside it, and remove this file

_Originally posted by @mateusoliveira43 in https://github.com/konveyor/pelorus/pull/824#discussion_r1089770856_

mateusoliveira43:
I believe this is easy to be done, but I would do after #720 is finished (to do not need to rebase the PR)

mateusoliveira43:
in https://github.com/konveyor/pelorus/pull/830 we added logic to shellckeck look into all `.sh` files of the project, except the 2 bats file in the `_test` folder. remove this ignore pattern

mateusoliveira43:
some tests file are getting big (like `exporters/tests/test_failure_exporter.py`). Separate files (in the previous example, separate by exporter provider)

mateusoliveira43:
there is a integration test folder (used by the mockoon tests), but integration tests are not only there. Think about a solution to better organize tests.

PR #830, opened 2023-01-30
fix: Add file extensions to all files

### LINKED ISSUE #684: File extensions
Hi everyone :wave:
Thanks for this awesome Project!

Pelorus have some files without extension. To get the list of files without extension, in the project root run 
```
find . -type f ! -name "?*.*" | grep -v "venv/\|git/"
```
Some of them should not have extension (like `.gitignore`), but I believe the Python files should, for example. This leads to the project quality checks not checking files properly.

For example, running `make black-check` does not warn of any errors today. But if we change the name of the file `scripts/bump-version` to `scripts/bump-version.py` and run `make black-check` again, black indicates format errors.

mateusoliveira43:

Add file extensions to all files. This enable the quality tools to check all of them.

We can add them to makefile or console scripts to avoid calling them with the extension.

## Linked Issues

resolves #684
depends on #944 

## Testing Instructions

I have run `find . -type f ! -name "?*.*" | grep -v "venv/\|git/\|.pytest_cache/\|htmlcov/"` to check for files without extension (there is still one that I think should have, that is `./pelorus-operator/PROJECT`, but it is generated by script, so we need to change script to change it).

To check if I have changed all needed places of invocation of the file, I have run `grep -Iinr 'name_of_the_file' . --exclude-dir={.venv,htmlcov,.pytest_cache,.mypy_cache}`.


scripts/chart-lint.py
```diff
@@ -10,3 +10,5 @@
     parser.parse_args()
     exit, _ = run_chart_lint()
     sys.exit(exit)
+
+# it is not used by CI, how to use it? what it does that make chart-lint does not?
```
mateusoliveira43, line 14: I did not see any mention to this file

scripts/create_pelorus_operator.sh
```diff
@@ -179,3 +179,6 @@ echo "  # # Deploy bundle, e.g. version v${NEW_OPERATOR_VERSION} in the pelorus
 echo "  # operator-sdk run bundle quay.io/${OPERATOR_ORG_NAME}/pelorus-operator-bundle:v${NEW_OPERATOR_VERSION} --namespace pelorus"
 echo "  # # Create pelorus instance from example in the pelorus namespace"
 echo "  # oc apply -f pelorus-operator/config/samples/charts_v1alpha1_pelorus.yaml -n pelorus"
+
+
+# Not documented!
```
mateusoliveira43, line 184: I did not see any mention to this file

Makefile
```diff
@@ -239,7 +239,7 @@ endif
 
 shellcheck:
 	@echo "🐚 📋 Linting shell scripts with shellcheck"
-	$(SHELLCHECK) $(find -name '*.sh' -type f | grep -v "venv/\|git/\|.pytest_cache/\|htmlcov/")
+	$(SHELLCHECK) $(shell find -name '*.sh' -type f | grep -v 'venv/\|git/\|.pytest_cache/\|htmlcov/')
```
mpryc, line 242: How about searching for all shell files (we can add exclude list similarly to .venv example):
```shell
$ find . -type f ! -path "*.venv/*" | xargs file -i - | grep "text/x-shellscript" | awk -F":" '{print $1}'
```

mateusoliveira43: I tested and this command find less files than the one I am using (it did not find `_test/conftest.sh`)

mpryc: That is correct. The `_test/conftest.sh` is not an shell file. Look at the first line:
```
#!/usr/bin/env bats
```

So it's bats interpreted file not shell and as such shellcheck should not be ran against it.

mateusoliveira43: I think that is not a problem https://github.com/koalaman/shellcheck/issues/417

or do you want to only check bash/sh files?

_test/bats-support-clone.sh
```diff
@@ -1,9 +1,10 @@
+#!/bin/bash
```
mateusoliveira43, line 1: conftest failed with this change, study how to do it

Makefile
```diff
@@ -262,3 +254,8 @@ endif
 clean-dev-env:
 	rm -rf ${PELORUS_VENV}
 	find . -iname "*.pyc" -delete
+
+PYMA=$(shell which python)
```
mateusoliveira43, line 258: remove, this was a test

Makefile
mpryc: I would be more careful and check only bash/sh files. There can be some list of *additional* files to be checked if they are not proper  `"text/x-shellscript"` type.

_test/bats-support-clone.sh
mpryc: See above. I would really ad this file to some exception list to be checked, or forget to shellscript this one all together.

scripts/chart-lint.py
mateusoliveira43: deleted

Makefile
mateusoliveira43: done

_test/bats-support-clone.sh
mateusoliveira43: alright, remove this one from shellcheck

Makefile
mateusoliveira43: I remove the bats file from the shellcheck with grep (to have a simpler command), but added a note here https://github.com/konveyor/pelorus/issues/825 to fix it later

scripts/create_pelorus_operator.sh
mateusoliveira43: opened a issue to fix this https://github.com/konveyor/pelorus/issues/841

mateusoliveira43:
can we delete `demo/deprecrated/demo.sh` file?

.github/workflows/shellcheck.yml
```diff
@@ -1,26 +1,11 @@
...
     paths:
-      - 'demo/demo-tekton.sh'
-      - 'scripts/create_release_pr'
-      - 'scripts/install_dev_tools'
-      - 'scripts/pre-commit'
-      - 'scripts/run-mockoon-tests'
-      - 'scripts/run-pelorus-e2e-tests'
-      - 'scripts/setup-pre-commit-hook'
-
+      - '**.sh'
```
mateusoliveira43, line 5: check the other workflows files

mateusoliveira43: done

.github/workflows/shellcheck.yml
```diff
@@ -1,26 +1,15 @@
...
-      - 'scripts/create_release_pr'
-      - 'scripts/install_dev_tools'
-      - 'scripts/pre-commit'
-      - 'scripts/run-mockoon-tests'
-      - 'scripts/run-pelorus-e2e-tests'
-      - 'scripts/setup-pre-commit-hook'
-
+      - '.github/workflows/shellcheck.yml'
+      - 'Makefile'
+      - '**.sh'
```
mpryc, line 7: Should this be?:

      - '**/*.sh'

.github/workflows/python-linting.yml
```diff
@@ -2,11 +2,10 @@ name: Pylava
 on:
   pull_request:
     paths:
-      - 'exporters/**'
-      - 'scripts/**'
       - 'pyproject.toml'
       - '.github/workflows/python-linting.yml'
       - 'Makefile'
+      - '**.py'
```
mpryc, line 8: Similar to .sh, should it be?

      - '**/*.py'

.github/workflows/python-formatting.yml
```diff
@@ -2,10 +2,10 @@ name: Check Python Formatting
 on:
   pull_request:
     paths:
-      - 'exporters/**'
-      - 'pyproject.toml'
       - '.github/workflows/python-formatting.yml'
+      - 'pyproject.toml'
       - 'Makefile'
+      - '**.py'
```
mpryc, line 8: Same as others, cause it won't catch all subdirs:

      - '**/*.py'

mpryc:
> can we delete `demo/deprecrated/demo.sh` file?

Yes, but another PR, so we don't hide such removals within other changes (easier to restore or track them if needed)

mpryc (approved):

mpryc (changes requested):
Fix shell or disable broken file in this PR and fix in following PR.

mpryc (dismissed):

PR #836, opened 2023-02-06
Webhook exporter allowing Pelorus to work with push data model

mpryc:
Check README.md in the webhook folder.

# Testing

Some testing instructions:

On one terminal:
```shell
$ make dev-env
$ source .venv/bin/activate
$ cd exporters/webhook
$ uvicorn app:app
```

On another terminal

```shell

$ cd exporters/tests/data
$ curl -X POST http://localhost:8000/pelorus/webhook -d @./webhook_pelorus_committime.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: committime"
$ curl -X POST http://localhost:8000/pelorus/webhook -d @./webhook_pelorus_deploytime.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: deploytime"
$ curl -X POST http://localhost:8000/pelorus/webhook -d @./webhook_pelorus_failure_created.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: failure"
$ curl -X POST http://localhost:8000/pelorus/webhook -d @./webhook_pelorus_failure_resolved.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: failure"
```

Navigate to the endpoint: `http://localhost:8000/metrics`, you should see all the metrics collected:

```html
# HELP commit_timestamp Commit timestamp
# TYPE commit_timestamp gauge
commit_timestamp{app="mongo-todolist",commit_hash="5379bad65a3f83853a75aabec9e0e43c75fd18fc",image_sha="sha256:af4092ccbfa99a3ec1ea93058fe39b8ddfd8db1c7a18081db397c50a0b8ec77d",namespace="mongo-persistent"} 1.557933657e+09
# HELP deploy_timestamp Deployment timestamp
# TYPE deploy_timestamp gauge
deploy_timestamp{app="mongo-todolist",image_sha="sha256:af4092ccbfa99a3ec1ea93058fe39b8ddfd8db1c7a18081db397c50a0b8ec77d",namespace="mongo-persistent"} 1.557933657e+09
# HELP failure_creation_timestamp Failure Creation Timestamp
# TYPE failure_creation_timestamp gauge
failure_creation_timestamp{app="mongo-todolist",failure_id="MONGO-1"} 1.557933657e+09
# HELP failure_resolution_timestamp Failure Resolution Timestamp
# TYPE failure_resolution_timestamp gauge
failure_resolution_timestamp{app="mongo-todolist",failure_id="MONGO-1"} 1.557933657e+09
```

Also for testing a simple `make unit-tests` should do most of the job, as the coverage for the new functions is almost at 100%.

@redhat-cop/mdt

### DIFF
--- a/exporters/tests/test_in_memory_metric.py
+++ b/exporters/tests/test_in_memory_metric.py
@@ -1,0 +1,170 @@
+#
+# Copyright Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+#
+
+
+from unittest import mock
+
+import pytest
+from prometheus_client import REGISTRY
+from prometheus_client.registry import Collector
+
+from webhook.models.pelorus_webhook import CommitTimePelorusPayload, PelorusPayload
+from webhook.store.in_memory_metric import (
+    PelorusGaugeMetricFamily,
+    _pelorus_metric_to_dict,
+    pelorus_metric_to_prometheus,
+)
+
+metric_labels = list(_pelorus_metric_to_dict(CommitTimePelorusPayload).values())
+
+in_memory_test_committime_metrics = PelorusGaugeMetricFamily(
+    "test_committime_metrics",
+    "Test timestamp",
+    labels=metric_labels,
+)
+
+
+class CustomCommitCollector(Collector):
+    def __init_subclass__(cls) -> None:
+        super().__init_subclass__()
+
+        # make sure __hash__ is something prometheus' registry can handle properly.
+        cls.__hash__ = lambda self: id(self)  # type: ignore
+
+    def collect(self) -> PelorusGaugeMetricFamily:
+        yield in_memory_test_committime_metrics
+
+
+class TestInMemoryMetric:
+    def setup_method(self):
+        self.custom_collector = CustomCommitCollector()
+        REGISTRY.register(self.custom_collector)
+
+    def teardown_method(self):
+        REGISTRY.unregister(self.custom_collector)
+
+    @pytest.mark.parametrize(
+        "name,timestamp,image_hash,namespace,commit_hash",
+        [
+            (
+                "todolist",
+                "timestamp_str",
+                "sha256:af4092ccbfa99a3ec1ea93058fe39b8ddfd8db1c7a18081db397c50a0b8ec77d",
+                "mynamespace",
+                "5379bad65a3f83853a75aabec9e0e43c75fd18fc",
+            ),
+        ],
+    )
+    def test_pelorus_gauge_metric_family(
+        self, name, timestamp, image_hash, namespace, commit_hash
+    ):
+        """
+        Verifies if the metric passed to the pelorus_metric_to_prometheus method
+        and then registered in our CustomCommitCollector is properly collected
+        by Prometheus. It does it by getting sample value and comparing the
+        timestamp of that metric to the timestamp of the data received from
+        Prometheus.
+        """
+        commit_payload = CommitTimePelorusPayload(
+            app=name,
+            timestamp=timestamp,
+            image_sha=image_hash,
+            namespace=namespace,
+            commit_hash=commit_hash,
+        )
+
+        prometheus_commit_metric = pelorus_metric_to_prometheus(commit_payload)
+        in_memory_test_committime_metrics.add_metric(
+            commit_payload.commit_hash,
+            prometheus_commit_metric,
+            commit_payload.timestamp,
+        )
+
+        metric_labels = {
+            "app": name,
+            "image_sha": image_hash,
+            "commit_hash": commit_hash,
+            "namespace": namespace,
+        }
+
+        query_result = REGISTRY.get_sample_value(
+            "test_committime_metrics",
+            labels=metric_labels,
+        )
+
+        assert query_result == timestamp
+
+
+def test_all_models_have_prometheus_mappings():
+    """
+    Safeguard test to ensure new PelorusPayload models also have
+    corresponding Prometheus model.
+
+    It does this by verifying if all models which are inheriting
+    from PelorusPayload class returns non empty dictionary from the
+    _pelorus_metric_to_dict method, where the mappings are defined.
+
+    We need to test this scenario to ensure we know how to store
+    each of the Pelorus data models into Prometheus.
+    """
+    import webhook.models.pelorus_webhook as pelorus_webhook
+
+    test_models = []
+
+    for cls in pelorus_webhook.__dict__.values():
+        if isinstance(cls, type) and issubclass(cls, PelorusPayload):
+            test_models.append(cls)
+
+    for test_model in test_models:
+        metric = _pelorus_metric_to_dict(test_model)
+        assert bool(metric)  # dict should not be empty
+
+
+class NewPelorusPayloadModel(PelorusPayload):
+    pass
+
+
+def test_model_do_not_have_prometheus_mapping():
+    """
+    Negative safeguard test to ensure that proper Error is raised
+    when the new models which are inheriting from PelorusPayload
+    model do not have corresponding Prometheus model defined in
+    the _pelorus_metric_to_dict method.
+    """
+
+    with pytest.raises(TypeError) as type_error:
+        _pelorus_metric_to_dict(NewPelorusPayloadModel)
+    assert "Improper prometheus data model" in str(type_error.value)
+
+
+@mock.patch(
+    "webhook.store.in_memory_metric._pelorus_metric_to_dict",
+    return_value={"app": "nonexisting"},
+)
+def test_model_missing_value_in_model(*args):
+    """
+    Negative safeguard test to ensure that the model which is
+    inheriting from the PelorusPayload  model does match
+    expected Prometheus model defined in the _pelorus_metric_to_dict
+    method.
+
+    It does this by verifying if all the expected variables are defined
+    in the PelorusPayload model, otherwise raising TypeError.
+    """
+
+    with pytest.raises(TypeError) as type_error:
+        pelorus_metric_to_prometheus(NewPelorusPayloadModel)
+    assert "Attribute nonexisting was not found in" in str(type_error.value)
--- a/exporters/tests/test_pelorus_base_plugin.py
+++ b/exporters/tests/test_pelorus_base_plugin.py
@@ -1,0 +1,349 @@
+#
+# Copyright Red Hat
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+#
+
+import http
+import json
+import time
+from typing import Any, Awaitable
+from unittest.mock import AsyncMock, Mock, patch
+
+import pytest
+from pydantic import ValidationError, parse_obj_as
+from typing_extensions import override
+
+from webhook.models.pelorus_webhook import (
+    CommitTimePelorusPayload,
+    PelorusMetric,
+    PelorusMetricSpec,
+)
+from webhook.plugins.pelorus_handler_base import (
+    Headers,
+    HTTPException,
+    PelorusWebhookPlugin,
+    PelorusWebhookResponse,
+    Request,
+)
+
+
+@pytest.mark.parametrize(
+    "test_no,http_response,http_response_code",
+    [
+        (1, "Invalid Payload.", http.HTTPStatus.UNPROCESSABLE_ENTITY),
+        (2, "Invalid Payload.", "422"),
+    ],
+)
+def test_pelorus_webhook_valid_response(test_no, http_response, http_response_code):
+    """
+    Test for the PelorusWebhookResponse class. The response code is a valid
+    HTTPStatus code. The http_response is valid string, so the
+    PelorusWebhookResponse is also valid.
+    """
+
+    # Test for proper message and response code is used.
+    response = PelorusWebhookResponse(
+        http_response=http_response, http_response_code=http_response_code
+    )
+
+    assert response.http_response == http_response
+    if test_no == 1:
+        assert response.http_response_code == http_response_code
+    if test_no == 2:
+        assert int(response.http_response_code) == int(http_response_code)
+
+
+@pytest.mark.parametrize(
+    "test_no,http_response,http_response_code",
+    [
+        (1, None, http.HTTPStatus.UNPROCESSABLE_ENTITY),
+        (2, "Invalid Payload.", 99),
+        (3, "Invalid Payload.", 601),
+        (4, "Invalid Payload.", None),
+    ],
+)
+def test_pelorus_webhook_invalid_response(test_no, http_response, http_response_code):
+    """
+    Negative test for the PelorusWebhookResponse class.
+    The response code must be a valid and http_response must be valid string.
+    """
+    with pytest.raises(ValidationError):
+        PelorusWebhookResponse(
+            http_response=http_response, http_response_code=http_response_code
+        )
+
+
+def test_pelorus_webhook_pong_response():
+    """
+    Test for the proper "pong" response which can be used by
+    any implemented plugin.
+    """
+
+    with pytest.raises(HTTPException) as http_exception:
+        PelorusWebhookResponse.pong(None)
+
+    assert type(http_exception).__name__ == "ExceptionInfo"
+    assert http_exception.typename == "HTTPException"
+
+    assert http_exception.value.detail == "pong"
+    assert http_exception.value.status_code == http.HTTPStatus.OK
+
+
+def test_abstract_classes():
+    """
+    Test for the plugin that did not implement required abstract methods.
+    """
+
+    class MyPelorusWebhookPlugin(PelorusWebhookPlugin):
+        def __init__(self, handshake_headers: Headers, request: Request) -> None:
+            super().__init__(handshake_headers, request)
+
+    # We should get an error:
+    # TypeError: Can't instantiate abstract class MyPelorusWebhookPlugin with abstract
+    #            methods _receive_pelorus_payload, handshake
+    with pytest.raises(TypeError) as type_error:
+        MyPelorusWebhookPlugin(None, None)
+
+    assert (
+        str(type_error.value)
+        == "Can't instantiate abstract class MyPelorusWebhookPlugin "
+        + "with abstract methods _handshake, _receive_pelorus_payload"
+    )
+
+
+class SimplePelorusWebhookPlugin(PelorusWebhookPlugin):
+    @override
+    async def _handshake(self) -> Awaitable[bool]:
+        pass
+
+    @override
+    async def _receive_pelorus_payload(self, Any) -> Awaitable[PelorusMetric]:
+        pass
+
+
+@pytest.mark.parametrize(
+    "handshake_headers,request_data",
+    [
+        ("Header.", "request data"),
+    ],
+)
+def test_pelorus_webhook_plugin_abc(handshake_headers, request_data):
+    """
+    Test for the plugin that did implement required abstract methods.
+    """
+
+    plugin_instance = SimplePelorusWebhookPlugin(
+        handshake_headers=handshake_headers, request=request_data
+    )
+
+    assert plugin_instance.headers == handshake_headers
+    assert plugin_instance.request == request_data
+
+
+class WithUserAgentWebhookPlugin(PelorusWebhookPlugin):
+    user_agent_str = "Pelorus-Webhook/"
+
+
+class WithoutUserAgentWebhookPlugin(PelorusWebhookPlugin):
+    pass
+
+
+class EmptyUserAgentWebhookPlugin(PelorusWebhookPlugin):
+    user_agent_str = ""
+
+
+def test_check_can_handle_methods():
+    """
+    Tests if the plugin can handle data based on the provided
+    string information that normally is given by the
+    "User-Agent:" header information.
+
+    Pelorus Webhook loads many plugins so each plugin should
+    know what payload data can it handle.
+    """
+
+    # user_agent_str = "Pelorus-Webhook/"
+    assert WithUserAgentWebhookPlugin.can_handle("Pelorus-Webhook/")
+    assert WithUserAgentWebhookPlugin.can_handle("Pelorus-Webhook/suffix")
+    assert not WithUserAgentWebhookPlugin.can_handle("")
+    assert not WithUserAgentWebhookPlugin.can_handle(None)
+    assert not WithUserAgentWebhookPlugin.can_handle("Incompatible-Webhook")
+
+    # user_agent_str not defined
+    assert not WithoutUserAgentWebhookPlugin.can_handle(None)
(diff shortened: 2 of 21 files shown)


exporters/webhook/app.py
```diff
@@ -0,0 +1,202 @@
...
+)
+
+pelorus.setup_logging()
+
+webhook_received = Counter("webhook_received_total", "Number of received webhooks")
+webhook_processed = Counter("webhook_processed_total", "Number of processed webhooks")
+
+commit_hash_set = set()
+
+# APIRouter here... ?
```
mateusoliveira43, line 49: you can, I prefer to put the routes in a separate folder, and then you need the router

exporters/webhook/app.py
```diff
@@ -0,0 +1,202 @@
...
+
+    try:
+        logging.debug("User-agent: %s" % user_agent)
+
+        if user_agent.startswith("GitHub-Hookshot/"):
+            payload_headers = parse_obj_as(GitHubDeliveryHeaders, request.headers)
+        elif user_agent.startswith("Pelorus-Webhook/"):
+            payload_headers = parse_obj_as(PelorusDeliveryHeaders, request.headers)
+        else:
+            return PelorusWebhookResponse(
```
mateusoliveira43, line 101: when there is an error I believe you can raise http exception. [example](https://github.com/mateusoliveira43/log-api/blob/deb8ac6f8986d1f78840933c7d837db17c4c4e3d/source/database/functions/user.py#L34-L37)

exporters/webhook/app.py
```diff
@@ -0,0 +1,202 @@
...
+        yield in_memory_commit_metrics
+
+
+collector = load_and_log(WebhookCommitCollector)
+REGISTRY.register(collector)
+print("Registered")
+
+
+@app.get("/metrics", response_class=PlainTextResponse)
+def metrics():
```
mateusoliveira43, line 189: should also be async?

exporters/webhook/models/pelorus_webhook.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        The failure may be one of two events. When it occurs it's created
+        and when it is resolved it's closed. Both events are different
+        Prometheus metrics, so we need to distinquish between them.
+        """
+
+        CREATED = "created"
+        RESOLVED = "resolved"
+
+    issue_number: str  # It's an str, because issue may be mix of str and int, e.g. Issue-1
```
mateusoliveira43, line 75: failure_id would be a better name?

exporters/webhook/app.py
mpryc: changed.

exporters/webhook/app.py
mpryc: changed with appropriate http error codes. This is just PoC so thought it's nice to use pydantic to control the data structure of returned item. The pydantic doesn't allow easilly wrap the exceptions around, but raise http is also ok.

exporters/webhook/app.py
mpryc: ok will leave as it is for now as it's poc.

exporters/webhook/app.py
mateusoliveira43: I think you were right not using the  http exception, the way you wanted we can after do like this https://fastapi.tiangolo.com/advanced/additional-responses/ and add to the openapi `http://127.0.0.1:8000 /docs`

exporters/tests/test_in_memory_metric.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python3
```
mateusoliveira43, line 1: all files have this shebang, but they should only have the copyright information, right?

exporters/tests/test_pelorus_base_plugin.py
```diff
@@ -0,0 +1,342 @@
...
+    ],
+)
+def test_pelorus_webhook_response(test_no, http_response, http_response_code):
+    """
+    Test for the PelorusWebhookResponse class. The response code must be a valid
+    HTTPStatus code. The http_response must be valid string.
+    """
+
+    # Test for proper message and response code is used.
+    if test_no < 3:
```
mateusoliveira43, line 60: break test in 2 to remove the `if` statement (test_pelorus_webhook_response_success and test_pelorus_webhook_response_error, for example)

exporters/tests/test_in_memory_metric.py
```diff
@@ -0,0 +1,136 @@
...
+        super().__init_subclass__()
+
+        # make sure __hash__ is something prometheus' registry can handle properly.
+        cls.__hash__ = lambda self: id(self)  # type: ignore
+
+    def collect(self) -> PelorusGaugeMetricFamily:
+        yield in_memory_test_committime_metrics
+
+
+@pytest.mark.parametrize(
```
mateusoliveira43, line 54: If we are not adding more cases, I would not use parametrize

exporters/tests/test_in_memory_metric.py
```diff
@@ -0,0 +1,136 @@
...
+    query_result = REGISTRY.get_sample_value(
+        "test_committime_metrics",
+        labels=metric_labels,
+    )
+
+    assert query_result == timestamp
+
+
+def test_all_models_have_prometheus_mappings():
+    import webhook.models.pelorus_webhook as pelorus_webhook
```
mateusoliveira43, line 104: I would have all imports top level and use parametrize here, to do not have a for of asserts

exporters/tests/test_in_memory_metric.py
```diff
@@ -0,0 +1,136 @@
...
+        _pelorus_metric_to_dict(MyPlugin)
+    assert "Improper prometheus data model" in str(type_error.value)
+
+
+@mock.patch(
+    "webhook.store.in_memory_metric._pelorus_metric_to_dict",
+    return_value={"app": "nonexisting"},
+)
+def test_model_missing_value_in_model(*args):
+    class MyPlugin(PelorusPayload):
```
mateusoliveira43, line 131: duplicate from line 118. Make it higher level and use in both tests

exporters/tests/test_pelorus_base_plugin.py
```diff
@@ -0,0 +1,342 @@
...
+    Test if the plugin is not a proper Pelorus handler.
+    """
+
+    class MyPelorusWebhookPlugin(PelorusWebhookPlugin):
+        pass
+
+    assert not MyPelorusWebhookPlugin.is_pelorus_webhook_handler()
+
+
+class WithUserAgentWebhookPlugin(PelorusWebhookPlugin):
```
mateusoliveira43, line 166: These classes are duplication of the ones in lines 149 and 160, no?

exporters/tests/test_pelorus_base_plugin.py
```diff
@@ -0,0 +1,342 @@
...
+    """
+    Tests if the plugin can handle data based on the provided
+    string information that normally is given by the
+    "User-Agent:" header information.
+
+    Pelorus Webhook loads many plugins so each plugin should
+    know what payload data can it handle.
+    """
+
+    assert WithUserAgentWebhookPlugin.can_handle("Pelorus-Webhook/")
```
mateusoliveira43, line 188: I think it is hard to read like this, I would have a test for each class

exporters/tests/test_pelorus_webhook_handler.py
```diff
@@ -0,0 +1,220 @@
...
+    methods.
+
+    The data payload is associated with the even type that comes
+    from the Header of the POST request.
+
+    The 'ping' event is a special case where the HTTPException
+    should be raised with the "pong" response. No payload data
+    is required in such case.
+    """
+    with patch(
```
mateusoliveira43, line 94: I would refactor this test, it is too complex

exporters/tests/test_pelorus_webhook_handler.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    Verifies all currently supported X-Pelorus-Event types and ensures
+    the handshake returns True for those events.
+
+    For missing "X-Pelorus-Event" value in the header or other then
+    supported event type an HTTPException exception is tested.
+    """
+    headers = Headers(header)
+    handler = PelorusWebhookHandler(None, request=None)
+    if test_no in [1, 2, 3, 4]:
```
mateusoliveira43, line 147: I would break in 2 tests

exporters/webhook/plugins/pelorus_handler_base.py
```diff
@@ -0,0 +1,208 @@
...
+
+from fastapi import HTTPException as FastapiHTTPException
+from pydantic import BaseModel
+from starlette.datastructures import Headers as StarletteHeaders
+from starlette.requests import Request as StarletteRequest
+
+from webhook.models.pelorus_webhook import PelorusMetric
+
+
+class HTTPException(FastapiHTTPException):
```
mateusoliveira43, line 31: did not understand these classes

exporters/webhook/plugins/pelorus_handler_base.py
```diff
@@ -0,0 +1,208 @@
...
+    async def _handshake(self, headers: Headers) -> Awaitable[bool]:
+        raise NotImplementedError  # pragma no cover
+
+    @abstractmethod
+    async def _receive_pelorus_payload(
+        self, json_payload_data: Any
+    ) -> Awaitable[PelorusMetric]:
+        raise NotImplementedError  # pragma no cover
+
+    async def handshake(self) -> Awaitable[Optional[bool]]:
```
mateusoliveira43, line 113: did not understand this method

exporters/tests/test_in_memory_metric.py
mpryc: yeah, copy paste. All files except app.py should have this line removed.

exporters/webhook/testdata/pelorus_committime.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
mpryc: Yes!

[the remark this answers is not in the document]
mpryc: Moving under:
```
exporters/tests/data/webhook_*
```

exporters/tests/test_in_memory_metric.py
mpryc: I actually prefer to always use parametrize if the test may be extended in the future to have more data. Here I was thinking about calling it multiple times and ensuring if there is only one metric in the prometheus stored.

exporters/tests/test_in_memory_metric.py
mpryc: For this test I actually deliberately wanted to have this under the test, so we can see that we are testing *all* subclasses of PelorusPayload which are found in the `webhook.models.pelorus_webhook` module. This test is checking if all of the models have prometheus corresponding mappings.

However I see I forgot to update this test with comments, which I need to do before creating another revision of this PR.

exporters/tests/test_in_memory_metric.py
mpryc: Fixed.

exporters/tests/test_pelorus_base_plugin.py
mpryc: changed.

exporters/tests/test_pelorus_base_plugin.py
```diff
@@ -0,0 +1,342 @@
...
+    """
+
+    with pytest.raises(HTTPException) as http_exception:
+        PelorusWebhookResponse.PONG(None)
+
+    assert type(http_exception).__name__ == "ExceptionInfo"
+    assert http_exception.typename == "HTTPException"
+
+    assert http_exception.value.detail == "pong"
+    assert http_exception.value.status_code == http.HTTPStatus.OK
```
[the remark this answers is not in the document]
mpryc: The PONG method is a classmethod of the PelorusWebhookResponse.

This is special case which informs webhook user-agent the ping was received and the webhook can be registered.

We do not want here to log anything, parse objects, receive any payload data, and so on..., Just respond with a very quick message saying the ping was received, that is why this classmethod raises HTTPException with proper OK code (saying, yes the ping is valid) and the "pong" as detail of this HTTPException.

exporters/tests/test_pelorus_base_plugin.py
mpryc: created line between With and Without and added comment:

```python

    # user_agent_str = "Pelorus-Webhook/"
    assert WithUserAgentWebhookPlugin.can_handle("Pelorus-Webhook/")
    assert WithUserAgentWebhookPlugin.can_handle("Pelorus-Webhook/suffix")
    assert not WithUserAgentWebhookPlugin.can_handle("")
    assert not WithUserAgentWebhookPlugin.can_handle(None)
    assert not WithUserAgentWebhookPlugin.can_handle("Incompatible-Webhook")

    # user_agent_str not defined
    assert not WithoutUserAgentWebhookPlugin.can_handle(None)
    assert not WithoutUserAgentWebhookPlugin.can_handle("")
```

exporters/webhook/plugins/pelorus_handler_base.py
mateusoliveira43: I think its unnecessary

exporters/webhook/plugins/pelorus_handler_base.py
```diff
@@ -0,0 +1,208 @@
...
+class PelorusWebhookResponse(BaseModel):
+    """
+    Class that represents the response to the user-agent making request.
+    """
+
+    http_response: str
+    http_response_code: http.HTTPStatus
+
+    @classmethod
+    def PONG(cls, payload_headers: Any):
```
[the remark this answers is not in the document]
mateusoliveira43: I see this only in with enum (for example, the http example `class HTTPStatus(IntEnum):`)

exporters/webhook/store/in_memory_metric.py
```diff
@@ -0,0 +1,174 @@
...
+    # This is to use model name, which equals to the
+    # class name. The __class_ can't be used here as
+    # it's inherited from pydantic.main.ModelMetaclass
+    if hasattr(pelorus_model, "__qualname__"):
+        model_name = pelorus_model.__qualname__
+    else:
+        # It's an instance
+        model_name = pelorus_model.__class__.__qualname__
+
+    if model_name == "PelorusPayload":
```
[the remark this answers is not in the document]
mpryc: I am not convinced it will be easier to read, could you give an example here?

exporters/tests/test_pelorus_base_plugin.py
```diff
@@ -0,0 +1,342 @@
...
+def test_plugin_register_methods():
+    """
+    Test if the register() method of the plugin
+    properly returns the identifier of the plugin.
+
+    The plugin identifier is a lower case string
+    from the user_agent_str value of the plugin.
+    """
+
+    assert (
```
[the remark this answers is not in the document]
mateusoliveira43: break in 2 (success and error) and parametrize the error one with the two similar classes

exporters/webhook/store/in_memory_metric.py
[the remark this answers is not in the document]
mateusoliveira43:
```python
variable_name = {
    "PelorusPayload": pelorus_payload,
    "FailurePelorusPayload": failure_payload,
    ...
}

payload = variable_name.get(model_name)

if payload:
    return payload
raise TypeError(f"Improper prometheus data model: {model_name}")
```

exporters/webhook/plugins/pelorus_handler_base.py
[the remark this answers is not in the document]
mpryc: ok changed to small letters.

exporters/webhook/store/in_memory_metric.py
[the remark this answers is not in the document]
mpryc: done

exporters/tests/test_pelorus_webhook_handler.py
mpryc: done.

exporters/tests/test_pelorus_webhook.py
```diff
@@ -0,0 +1,116 @@
...
+
+import pytest
+from fastapi.testclient import TestClient
+
+from webhook import app as pelorus_webhook
+from webhook.plugins.pelorus_handler_base import PelorusWebhookPlugin
+
+client = TestClient(pelorus_webhook.app)
+
+test_directory = os.path.dirname(os.path.realpath(__file__))
```
mateusoliveira43, line 30: Lets use `pathlib` (also one of Python standard libraries), for handling paths

exporters/tests/test_in_memory_metric.py
```diff
@@ -0,0 +1,175 @@
...
+    def collect(self) -> PelorusGaugeMetricFamily:
+        yield in_memory_test_committime_metrics
+
+
+class TestInMemoryMetric:
+    def setup_method(self):
+        self.custom_collector = CustomCommitCollector()
+        REGISTRY.register(self.custom_collector)
+
+    def teardown_method(self):
```
mateusoliveira43, line 58: This works? doesn't have to inherit from unittest.TestCase?

But I would remove this and use the solution I created in the pagerduty PR

exporters/webhook/plugins/pelorus_handler.py
```diff
@@ -0,0 +1,168 @@
...
+        if self.payload_headers and self.payload_headers.event_type:
+            try:
+                data = self.handler_functions.get(
+                    self.payload_headers.event_type, lambda payload: None
+                )(json_payload_data)
+                return PelorusMetric(
+                    metric_spec=self.payload_headers.event_type, metric_data=data
+                )
+            except ValidationError as ex:
+                logging.error(self.payload_headers)
```
mateusoliveira43, line 162: uvicorn has its own logger, should we use the same?
Screenshot from 2023-03-01 17-44-57

exporters/webhook/plugins/pelorus_handler_base.py
mpryc: Having those classes in our base class of the plugin:
 - We simplify imports in the plugins. The plugin does not need to import the classes from other libraries, which can simplify the import statements and make the code more readable.
 - We have abstracted implementation: By creating those classes that inherit from starlette or fastapi library classes, the implementation details of those classes are abstracted away, making it easier to swap out the implementation or make changes without affecting the plugin code. So we can decide to use other frameworks and keep the plugins compatible.
 - By creating those classes within the plugin, the plugin can better encapsulate the functionality it provides.
 
From the cons I can think of just a bit more code, but we don't implement body of those classes so we don't have visible performance impact nor maintenance overhead.

exporters/tests/test_pelorus_webhook.py
mpryc: Done. `os` vs `pathlib` is really not that big of a deal here. We don't use any functions that would be dependent on the operating system, so I don't see any advantage of pathlib in this situation.  Even more we need to convert pathlib objects to the strings, but if that's your preference I am also happy with it and changed to the pathlib module.

`os` is faster and have more native (unix like) feeling while using.

mpryc: see diff and decide:

```diff
diff --git a/exporters/tests/test_pelorus_webhook.py b/exporters/tests/test_pelorus_webhook.py
index 87fd4cd..63530e6 100644
--- a/exporters/tests/test_pelorus_webhook.py
+++ b/exporters/tests/test_pelorus_webhook.py
@@ -16,7 +16,7 @@
 
 
 import json
-import os
+from pathlib import Path
 from http import HTTPStatus
 
 import pytest
@@ -27,7 +27,7 @@ from webhook.plugins.pelorus_handler_base import PelorusWebhookPlugin
 
 client = TestClient(pelorus_webhook.app)
 
-test_directory = os.path.dirname(os.path.realpath(__file__))
+test_directory = Path(__file__).resolve().parent
 
 TEST_DATA_DIR = f"{test_directory}/data/"
 
diff --git a/exporters/webhook/app.py b/exporters/webhook/app.py
index 460ba32..d4fc1dc 100644
--- a/exporters/webhook/app.py
+++ b/exporters/webhook/app.py
@@ -19,7 +19,7 @@ import asyncio
 import http
 import importlib
 import logging
-import os
+from pathlib import Path
 import sys
 from typing import Optional, TypeVar
 
@@ -48,7 +48,7 @@ from webhook.store.in_memory_metric import (
 
 pelorus.setup_logging()
 
-script_directory = os.path.dirname(os.path.realpath(__file__))
+script_directory = Path(__file__).resolve().parent
 
 DEFAULT_PLUGIN_FOLDER = "plugins"
 
@@ -77,13 +77,14 @@ def register_plugin(webhook_plugin: PelorusWebhookPlugin):
 
 def load_plugins(plugins_dir_name: Optional[str] = DEFAULT_PLUGIN_FOLDER):
     plugin_path = f"{script_directory}/{plugins_dir_name}/"
-    sys.path.append(script_directory)
+    sys.path.append(f"{script_directory}")
     logging.info(f"Loading plugins from directory {plugin_path}")
-    if os.path.isdir(plugin_path):
-        for filename in os.listdir(plugin_path):
-            if filename.endswith("_handler.py"):
+    plugin_dir_path = Path(plugin_path)
+    if plugin_dir_path.is_dir():
+        for filename in plugin_dir_path.iterdir():
+            if filename.is_file() and filename.name.endswith("_handler.py"):
                 module = importlib.import_module(
-                    f".{filename[:-3]}", package=plugins_dir_name
+                    f".{filename.stem}", package=f"{plugins_dir_name}"
                 )
                 for name in dir(module):
                     obj = getattr(module, name)
```

exporters/tests/test_in_memory_metric.py
mpryc: Yes, because we use **pytest** and this is `setup_method` from **pytest** and not `setUp` from **unittest**, same for the `teardown_method` which is not `tearDown` from **unittest**, you can invoke the test with some verbose info and redirect stdout from the test with such simple commands for the one file:

```shell
$ source .venv/bin/activate
$ cd exporters/tests
$ pytest -s -vvv test_in_memory_metric.py
```

exporters/webhook/store/in_memory_metric.py
```diff
@@ -0,0 +1,176 @@
...
+        labels: Optional[Sequence[str]] = None,
+        unit: str = "",
+    ):
+        super().__init__(name, documentation, value, labels, unit)
+        self.lock = threading.Lock()
+        self.added_metrics = set()
+
+    def add_metric(self, metric_id, *args, **kwargs):
+        with self.lock:
+            if metric_id and metric_id not in self.added_metrics:
```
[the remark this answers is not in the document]
mpryc: I don't think so, user do not need to know how we handle duplicates and it's ok to receive duplicate. Same happens on providers like Github webhook which allows to resent same payload.

The sender should just be informed "yes webhook received data" what it does with the data should be hidden from the sender.

exporters/webhook/plugins/pelorus_handler.py
mpryc: We could if we *really* wanted, but that comes at a cost of handling many `uvicorn`, `gunicorn` loggers and passing our streams to those. I don't think it's worth here.

See this thread: https://github.com/tiangolo/uvicorn-gunicorn-fastapi-docker/issues/19

Also I want to use setup_logging, so the logging from other exporters are same, this isn't really expected fromat by the uvicorn handler. I think this isn't important enough to be resolved now, we can always address this issue later if needed, by reworking logging mechanism for all the exporters and providing different handlers for the terminal and uvicorn.

mpryc: Another thing is separation between http events and internal application validations.

Here the log should be:

INFO for http event (we have OK message saying 422 error occured), so we sent OK back to the requestor with the "{"detail":"Invalid payload."}" message, and we log this as INFO, because the response was OK.

ERROR from validations, to know in logs *why* the data was invalid, but this is only information for user to see why the payload was invalid, we should not tell the requester *why* we should just say *bye bye invalid payload*. So the ERROR is not really uvicorn error, uvicorn was OK.

exporters/tests/test_failure_exporter.py
```diff
@@ -189,6 +189,7 @@ def test_jira_search_with_wrong_jql():
     assert len(issues) == 0
 
 
+@pytest.mark.skip(reason="Issue-848")
```
mateusoliveira43, line 192: remove this skip, as it was fixed

mpryc: Removed. Will push in few once tests passes.

mateusoliveira43 (approved):
As discussed, the missing features will be implemented in following PRs, guided by the created issues

Issue #751, opened 2022-12-16
Helm values in the openAPIV3Schema


mpryc:
We have problem when we want to use properties section in here. The OCP evaluates every entry in properties and it's not matching openAPIV3Schema because it's Helm values.yaml which is in fact json.

I have tried with:
```
              properties:
                exporters:
                  description: >-
                    References configuration for the Pelorus exporters. More
                    info about exporters configuration:
                    https://pelorus.readthedocs.io/en/latest/Configuration/#configuring-exporters-details
                  type: object
                  properties:
                    instances:
                      description: >-
                        Exporter instances configuration. More information:
                        https://pelorus.readthedocs.io/en/latest/Configuration/
                      type: array
                      items:
                        type: object
                      properties:
                        app_name:
                          description: 'Exporter name, must start with "- ".'
                          type: string
                        enabled:
                          description: 'If set to false, exporter won''t be provisioned.'
                          type: boolean
                        env_from_configmaps:
                          description: >-
                            ConfigMap name(s) with configuration used by the
                            exporter instance.
                          type: object
                        env_from_secrets:
                          description: >-
                            Secret name(s) with secrets used by the exporter
                            instance.
                          type: object
                        exporter_type:
                          description: >-
                            Exporter type as in the:
                            https://pelorus.readthedocs.io/en/latest/Configuration/#configuring-exporters-overview
                          type: string
```

But OCP was creating malformed values, because there is no such type that can fit our list (problematic - at the beginning of the line):
```
- app_name: first
- app_name: second
```

_Originally posted by @mpryc in https://github.com/konveyor/pelorus/pull/750#discussion_r1050746324_

KevinMGranger:
I don't understand the issue-- the form can't represent this spec, or we can't represent what we want in the spec?

mpryc:
> I don't understand the issue-- the form can't represent this spec, or we can't represent what we want in the spec?

The form can not represent this spec (with "-")

KevinMGranger:
I still don't understand. Where are dashes coming from, and why does `app_name` say they're required? Are you saying the form doesn't understand how to do values nested in an array like we need in `instances`?

mpryc:
> 

The example OpenShift instance spec is as follows. You can see the "-" before app_name. This is exactly the same as we have in our documentation (many of the code-snippets in our documentation have such examples): https://pelorus.readthedocs.io/en/latest/Configuration/#instance-config

[...]

PR #840, opened 2023-02-08
Add external exporter to Pelorus

mateusoliveira43:

Add a Proof of Concept (PoC) for adding external exporters to Pelorus.

## Linked Issues

No.

## Testing Instructions

Follow the instructions in `poc_external_exporter/poc_external_prometheus.md`.

### DIFF
--- a/poc_external_exporter/prometheus-additional.yaml
+++ b/poc_external_exporter/prometheus-additional.yaml
@@ -1,0 +1,3 @@
+- job_name: "test"
+  static_configs:
+  - targets: ["node.demo.do.prometheus.io"]
--- a/poc_external_exporter/additional-scrape-configs.yaml
+++ b/poc_external_exporter/additional-scrape-configs.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+data:
+  prometheus-additional.yaml: LSBqb2JfbmFtZTogInRlc3QiCiAgc3RhdGljX2NvbmZpZ3M6CiAgLSB0YXJnZXRzOiBbIm5vZGUuZGVtby5kby5wcm9tZXRoZXVzLmlvIl0K
+kind: Secret
+metadata:
+  creationTimestamp: null
+  name: additional-scrape-configs
--- a/poc_external_exporter/poc_external_prometheus.md
+++ b/poc_external_exporter/poc_external_prometheus.md
@@ -1,0 +1,61 @@
+# Example of using Pelorus with external exporter
+
+For more information, check [here](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/additional-scrape-config.md#creating-an-additional-configuration).
+
+Create a file called `prometheus-additional.yaml`, with the following content
+```yaml
+- job_name: "test"
+  static_configs:
+  - targets: ["node.demo.do.prometheus.io"]
+```
+We could have added more than one target here and used any of this [configurations](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
+
+Create Secret file from previous file
+```
+oc create secret generic additional-scrape-configs \
+--from-file=prometheus-additional.yaml \
+--dry-run=client -oyaml > additional-scrape-configs.yaml
+```
+
+Create Secret in the cluster
+```
+oc apply -f additional-scrape-configs.yaml -n pelorus
+```
+
+Create Pelorus instance with the following content
+```yaml
+apiVersion: charts.pelorus.konveyor.io/v1alpha1
+kind: Pelorus
+metadata:
+  name: github-example-configuration
+spec:
+  exporters:
+    instances:
+      - app_name: deploytime-exporter
+        exporter_type: deploytime
+        extraEnv:
+          - name: NAMESPACES
+            value: example_namespace
+      - app_name: committime-exporter
+        exporter_type: committime
+        extraEnv:
+          - name: NAMESPACES
+            value: example_namespace
+```
+
+Navigate to Prometheus operator route and check the number of targets (in **Status** -> **Targets**). Tt should have 2 targets (1 deploytime and 1 committime).
+
+Add the following content
+```yaml
+  additionalScrapeConfigs:
+    name: additional-scrape-configs
+    key: prometheus-additional.yaml
+```
+to the `spec` section of prometheus-pelorus yaml.
+
+A **warning** message like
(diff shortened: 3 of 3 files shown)


poc_external_exporter/prometheus-additional.yaml
```diff
@@ -0,0 +1,3 @@
+- job_name: "test"
+  static_configs:
+  - targets: ["node.demo.do.prometheus.io"]
```
mpryc, line 3: If there is more then one target it's standard list, so can be?:

```yaml
- job_name: "test"
  static_configs:
  - targets:
    -  "node.demo.do.prometheus.io"
    -  "another.target.io"
```

mateusoliveira43: I think so https://prometheus.io/docs/prometheus/latest/configuration/configuration/#static_config

mpryc (changes requested):
Overall I would use the already existing helm chart to create secrets and other resources instead of manually creating them:

https://github.com/konveyor/pelorus/blob/master/charts/pelorus/templates/prometheus-scrape-config-secret.yaml

Also some instructions on how this was done:

https://github.com/konveyor/pelorus/commit/f03a03677627349e3ccd38318d942398f78edf10#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R47

mateusoliveira43:
As talked with @mpryc, using `extra_prometheus_hosts` should get the job done. But I tested out and I believe `extra_prometheus_hosts` is not working (the configuration of charts seems correct, `charts/pelorus/templates/prometheus-scrape-config-secret.yaml` would create the config and secret and `charts/pelorus/templates/prometheus-cr.yaml` would add the config to prometheus operator, just like in this PoC, but I think not even the secret is being created, because `additional-scrape-configs` does not appear in `oc get secret -n pelorus`)

I have created an Pelorus instance with
```yaml
apiVersion: charts.pelorus.konveyor.io/v1alpha1
kind: Pelorus
metadata:
  name: external-example-configuration
spec:
  exporters:
    instances:
      - app_name: deploytime-exporter
        exporter_type: deploytime
        extraEnv:
          - name: NAMESPACES
            value: example_namespace
      - app_name: committime-exporter
        exporter_type: committime
        extraEnv:
          - name: NAMESPACES
            value: example_namespace
  extra_prometheus_hosts:
    - id: "mateus"
      hostname: "node.demo.do.prometheus.io"
      password: ""
```
to test it out.

I tried it manually, by creating `prometheus-additional.yaml` with (this is the content that helm would have created using `extra_prometheus_hosts`)
```yaml
- job_name: "federated-prometheus-mateus"
  scrape_interval: 15s
  honor_labels: true
  metrics_path: '/federate'
  params:
    'match[]':
      - '{job="openshift-state-metrics"}'
  scheme: "https"
  basic_auth:
    username: 'internal'
    password: ""
  tls_config:
    insecure_skip_verify: true
  static_configs:
    - targets:
        - "node.demo.do.prometheus.io"
      labels:
        federated_job: "federated-prometheus-mateus"
```
and following the subsequent steps of `poc_external_exporter/poc_external_prometheus.md`, and it works.
Screenshot from 2023-03-04 12-23-29

I suspect is something related to checking if the variable is being used, the other logic seems correct.

mateusoliveira43:
Tested more, and I think the problem is the operator, not helm.

I have run `helm template --debug pelorus-test charts/pelorus > debug.yaml` changing the `charts/pelorus/values.yaml`, and the configs are added to `prometheus-cr.yaml`.

I tried to used `custom_ca` in operator, and it did not change anything in `prometheus-cr.yaml` (to test this, create an instance, can be like the example in OpenShift, but adding `custom_ca: true`. It should apply the logic of `charts/pelorus/templates/prometheus-cr.yaml`, but if you check the object, it does not have any of the options `custom_ca` would add).

I believe the problem is in `pelorus-operator/bundle/manifests/charts.pelorus.konveyor.io_pelorus.yaml` (related to https://github.com/konveyor/pelorus/issues/751). Because the values of `spec.properties` that are listed (only `exporters`, `prometheus_retention`, `prometheus_retention_size`, `prometheus_storage`, `prometheus_storage_pvc_capacity` and  `prometheus_storage_pvc_storageclass` from [full list](https://pelorus.readthedocs.io/en/v2.0.7/GettingStarted/configuration/PelorusCore/#list-of-all-configuration-options)), seem to work, but `extra_prometheus_hosts` and `custom_ca`, for example, that are not listed, do not work.

mpryc:
@mateusoliveira43 tested and you are right the values are not consumed by the operator.

The Prometheus operator expected values [additionalScrapeConfigs](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.PrometheusSpec) which are in our templates are [correct](https://github.com/konveyor/pelorus/blob/master/charts/pelorus/templates/prometheus-cr.yaml#L39)

The pelorus operator templates fails to create secret `additional-scrape-configs` from the `charts/pelorus/templates/prometheus-scrape-config-secret.yaml`, which then is used by the prometheus-operator.

mpryc:
@mateusoliveira43 looks like it's not only related to the external exporter configs, also defining things such as thanos won't be used:

```
thanos_bucket_name: <...>
bucket_access_point: <...>
bucket_access_key: <...>
bucket_secret_access_key: <...>
```

This did not created proper thanos configs and configured grafana to point to the prometheus rather then thanos.

mpryc:
The top level configs such as, does work:

```
  prometheus_retention: 1y
  prometheus_retention_size: 1GB
  prometheus_storage: true
  prometheus_storage_pvc_capacity: 2Gi
  prometheus_storage_pvc_storageclass: gp2
```

mpryc:
So it looks like anything outside of the schama is ignored while creating instance. This was not the case when we created 0.0.1 of this operator.

mpryc:
Depends On: https://github.com/konveyor/pelorus/pull/861

Note it's possible to test without creating new operator by modyfing `Pelorus` CRD in the cluster.
Left side menu (Administration->CustomResourceDefinition), search for Pelorus, select Yaml and add missing option (number of spaces are important).

Screenshot from 2023-03-06 13-07-10

mateusoliveira43:
@mpryc this workaround did not work for me, the Pelorus manager gets error

mateusoliveira43:
Closing this PR without merging as it was a PoC. The implementation with the given requirements from https://github.com/orgs/konveyor/projects/51/views/2?pane=issue&itemId=20613515 will be opened in follow up

PR #834, opened 2023-02-04
Add ADR proposal and a skeleton for architecture docs.

KevinMGranger:
## Describe the behavior changes introduced in this PR

Adds the ADR proposal document.

Also adds the smallest README for the architecture docs directory.

This was separated from #753 to keep it small and not depend on the committime changes.

Let's talk about the ADR format, but let's not bikeshed too much, especially about where the architecture docs should live.

## Linked Issues

addresses parts 2 and 3 of #727

PR #844, opened 2023-02-20
ADR for the Pelorus Webhook Exporter

mpryc:
This Architecture Decision Record outlines the Pelorus Webhook Exporter.

Depends on: https://github.com/konveyor/pelorus/pull/834

@redhat-cop/mdt


mpryc:
Partially resolves #846

mateusoliveira43:
Just having PRs with good descriptions would not do same job as the ADRs?

Today we have a lot of PRs without description (or a good description). We would need to enforce this (and we already have a PR template)

What you believe are the benefits of ADRs over we (and our contributors) writing good and meaningful PRs' descriptions?

mpryc:
@mateusoliveira43 
The ADR may be updated and it's living document where PR is sometimes hard to find.

Also the ADR may and should be submitted before PR with implementation.

weshayutin:
@mateusoliveira43 In the context of large complex features having an ADR or design PR allows for discussion and review of a feature prior to any implementation.  It's important to have a blueprint prior to building a ship, you wouldn't want to start building a ship at the same time you started building it.

We have been discussion the designs informally in our daily standups, so IMHO we are just recording those discussion in github by creating an ADR or Design PR.

mateusoliveira43:
@mpryc @weshayutin alright, the information prior to implementation can be important. I think we could also try to use issues for that, but lets give ADRs a shot then :smile:

architecture_docs/ADRs/0001-webhook_architecture.md
```diff
@@ -0,0 +1,66 @@
...
+5. Prometheus scrapes the webhook's Prometheus endpoint and stores the received data in its time-series database
+
+```
+                +-----------------------+            +-----------------------+
+                |  External System (1)  |            |  External System (N)  |
+                +-----------+-----------+            +-----------+-----------+
+                            | (send)                             | (send)
+            +---------------+              +---------------------+----------+
+            |                              |                                |
++-----------v-----------+      +-----------v-----------+        +-----------v-----------+
```
mateusoliveira43, line 46: I like the image, but I thought you can use one instance of webhook to receive data from countless external systems. The image passes me the message you need one for each external system, like our exporters today. 

So, what is the intended way to use the webhook?

mpryc: The (N) in the "External System (N)" represents the "countless external systems".

The use case to have multiple webhooks is still valid for the scenario where the webhook is using secrets to configure the "Secret". For example here is [GitHub secret doc](https://docs.github.com/en/webhooks-and-events/webhooks/securing-your-webhooks#setting-your-secret-token).

For the scenario with "secret" if user would like to create many endpoints with different secrets then the user would need to create multiple instances of webhook to apply such secret to it.

mateusoliveira43 (changes requested):
As we discussed, missing template for closing this PR

architecture_docs/ADRs/0001-webhook_architecture.md
mateusoliveira43: we can address this concern of mine in the webhook documentation

docs/adr/0002-Webhook-Pelorus-Exporter.md
```diff
@@ -0,0 +1,87 @@
...
+
+Accepted
+
+## Context
+
+Currently all Pelorus exporters like failure, committime and deploytime are designed and implemented to use `pull` model to gather the data from external sources such as GitHub, GitLab, Jira, ServiceNow, etc...
+
+The proposal is to allow Pelorus to receive data from the external data sources using, which will use `push` method to send data to the Pelorus.
+
+The benefits of the `push` method:
```
mateusoliveira43, line 15: I would group this in the **Consequences** section

mateusoliveira43 (approved):

Issue #856, opened 2023-03-03
CI job to create pelorus-exporter quay.io/pelorus/pelorus-webhook-exporter images.


mpryc:
Create CI job to build our webhook-exporter and store it in the quay.io similarly to our exporters.

A new quay.io/pelorus/pelorus-webhook-exporter registry is already set up for this.

For more information please see the relevant PRs:
  #844 
  #836

Issue #857, opened 2023-03-03
Create documentation for the webhook-exporter


mpryc:
The webhook exporter is a new type of exporter and should be documented on how to communicate with it. 

Also a well defined schema for different metrics should be part of that documentation, so the users should have clear instructions what data is expected to be delivered.

For more information please see the relevant PRs:
https://github.com/konveyor/pelorus/pull/844
https://github.com/konveyor/pelorus/pull/836

mateusoliveira43:
Important: add how to create custom plugins in the documentation

PR #859, opened 2023-03-03
Add GitHub Actions to build the webhook exporter

mpryc:
Covers building webhook exporter and pushing it to the quay.io repository for the commit, pre-release and release events.

Documentation will be covered as part of the #857, however here one link was added as it's directly pointing to the quay image which will be built by those actions.

Depends on: #836

@redhat-cop/mdt

### DIFF
--- a/.github/workflows/build-pelorus.yml
+++ b/.github/workflows/build-pelorus.yml
@@ -131,3 +131,32 @@ jobs:
           registry: ${{ env.imageregistry }}/${{ env.imagenamespace }}
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
+
+
+  build-webhook:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      # Setup S2i and Build container image
+      - name: Setup and Build
+        id: build-exporter
+        uses: redhat-actions/s2i-build@v2
+        with:
+          image: 'pelorus-webhook-exporter'
+          path_context: 'exporters'
+          builder_image: ${{ env.builder_image }}
+          tags: ${{ env.latesttag }} ${{ github.sha }}
+          log_level: ${{ env.s2i_log_level }}
+
+
+      # Push Image to Quay registry
+      - name: Push To Quay Action
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-exporter.outputs.image }}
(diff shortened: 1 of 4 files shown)


mpryc:
Fixes #856

.github/workflows/pre-release.yml
```diff
@@ -57,3 +57,13 @@ jobs:
           registry: ${{ env.imageregistry }}
           registry_username: ${{ secrets.QUAY_USERNAME }}
           registry_password: ${{ secrets.QUAY_PASSWORD }}
+
+      - name: Tag failure with (pre)released version
```
mateusoliveira43, line 61: webhook instead of failure

.github/workflows/build-pelorus.yml
```diff
@@ -131,3 +131,32 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      # Setup S2i and Build container image
+      - name: Setup and Build
+        id: build-exporter
+        uses: redhat-actions/s2i-build@v2
+        with:
+          image: 'pelorus-webhook-exporter'
+          path_context: 'exporters'
```
mateusoliveira43, line 148: I think this ok, but I want to understand how this works. 

It builds the whole exporters package on it container? It build only webhook on it (and each exporter on the other ones)?

mateusoliveira43: Alright, checked it out, it builds everything. And then my question is: why have 4 equal images in quay? (the `pelorus-{{type}}-exporter` ones, basically what changes them is the `APP_FILE` environment variable)

I did the steps of the action, got `2si` executable from https://github.com/openshift/source-to-image and run `./s2i build exporters registry.access.redhat.com/ubi8/python-39 pelorus-webhook-exporter:latest --loglevel 2`. 

Then run `docker container run -ti --rm -e APP_FILE=webhook/app.py pelorus-webhook-exporter` to test out, and worked. Should we add a test job for checking the image entrypoint? (for the webhook looks easier, the other exporters need kube-config file, I think)

.github/workflows/pre-release.yml
mpryc: fixed

.github/workflows/build-pelorus.yml
mateusoliveira43: Alright, we can think about it later. But I would like that we have a issue related to that, so we do not lose track.

mateusoliveira43: created it https://github.com/konveyor/pelorus/issues/862

Issue #860, opened 2023-03-06
Config options other then specified in the open api schema are not used


mpryc:
Issue with additional config options for the pelorus instance, which are not used if they are not specified in the openapi schema.

PR #861, opened 2023-03-06
Ensure additional config options are used by the Pelorus

mpryc:
This adds an option to the CRD which preserves the config options which are additional to the ones we define in the openapi schema. This option is already existing for the exporters but was missing for the top level configs.

@redhat-cop/mdt


mpryc:
Fixes #860

mateusoliveira43:
I would close this PR without merging, as another solution is proposed here https://github.com/konveyor/pelorus/pull/869

Or do you believe this option of letting user add any config can be useful?

mpryc:
Closing as we agree to change the API schema to confirm with only known and supported options.

PR #865, opened 2023-03-06
Update webhook port to 8080 to match other exporters

mpryc:
We use one server to serve both /pelorus/webhook and the /metrics for prometheus. They should run on the same port as other exporters.

@redhat-cop/mdt


mateusoliveira43 (approved):

mpryc:
Merging as the prow jobs has noting to do with that PR and need it for the webhook follow work.

Issue #867, opened 2023-03-07
Webhook exporter does not validate timestamp properly


mpryc:
It's possible to cause webhook exporter to fail if the wrong timestamp format is provided, e.g. string and not the EPOCH int timestamp "1s57933657".

This causes pod to fail with the following traceback:

```
File "/opt/app-root/lib64/python3.9/site-packages/prometheus_client/exposition.py", line 195, in sample_line
return f'{line.name}{labelstr} {floatToGoString(line.value)}{timestamp}\n'
File "/opt/app-root/lib64/python3.9/site-packages/prometheus_client/utils.py", line 9, in floatToGoString
d = float(d)
ValueError: ("could not convert string to float: '1s57933657'", Metric(failure_creation_timestamp, Failure Creation Timestamp, gauge, , [Sample(name='failure_creation_timestamp', labels={'app': 'mongo-todolist', 'failure_id': 'MONGO-1'}, value='1557933657', timestamp=None, exemplar=None), Sample(name='failure_creation_timestamp', labels={'app': 'mongo-todolist', 'failure_id': 'MONGO-1'}, value='1s57933657', timestamp=None, exemplar=None)]))
```

mpryc:
To reproduce the issue:
Start webhook exporter in one terminal
```shell
make dev-env
source .venv/bin/activate
python exporters/webhook/app.py
```

From another teminal send some fake data:
```shell
curl -X POST http://localhost:8080/pelorus/webhook -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: failure" -d '{"app": "mongo-todolist","failure_id": "MONGO-1","failure_event": "created", "timestamp": "15579ff33657"}'
```

Open web-page with `metrics`:
http://localhost:8080/metrics

Expected result:
  metrics for prometheus visible

Actual result:
 * "Internal Server Error"
 * Traceback  in the terminal where python app.py was running
 * http://localhost:8080/metrics no longer accessible

mateusoliveira43:
Sending timestamp as `None` still breaks code. I believe the problem is setting None as the default in here https://github.com/konveyor/pelorus/blob/master/exporters/webhook/models/pelorus_webhook.py#L54

Writing as `timestamp: int = Field(ge=1262307661, le=2840144461)` prevents this.

Write test to catch this use case and test a little more.

PR #871, opened 2023-03-07
Fix the webhook timestamp validator

mpryc:
We simplify the timestamp of the webhook pelorus plugin to accept only EPOCH timestamp.

It's up to the service sending POST data to ensure timestamp is a valid format and we do not have troubles converting them from unknown data format.

Fixes: #867 

## Testing Instructions

Added test to capture wrong data and modify other tests to cover positive test case.
$ make unit-tests

@redhat-cop/mdt

### DIFF
--- a/exporters/webhook/models/pelorus_webhook.py
+++ b/exporters/webhook/models/pelorus_webhook.py
@@ -42,14 +42,16 @@ class PelorusPayload(BaseModel):
 
     Attributes:
         app (str): Application name.
-        timestamp (str): Timestamp of the event. This is different from the
-                         time when the webhook could have been received.
+        timestamp (int): 10 digit EPOCH timestamp of the event. This
+                         is different from the time when the webhook
+                         could have been received. The date value must
+                         be between 1.1.2010 and 1.1.2060
     """
 
     # Even if we consider git project name as app, it still should be below 100
     app: str = Field(max_length=200)
-    # ISO 8601 is 19 chars long, so don't expect any format can be longer then 50 chars
-    timestamp: str = Field(max_length=50)
+
+    timestamp: int = Field(None, ge=1262307661, le=2840144461)
 
     def get_metric_model_name(self) -> str:
         return type(self).__name__
--- a/exporters/tests/test_in_memory_metric.py
+++ b/exporters/tests/test_in_memory_metric.py
@@ -61,7 +61,7 @@ class TestInMemoryMetric:
         [
             (
                 "todolist",
-                "timestamp_str",
+                "1678269658",
                 "sha256:af4092ccbfa99a3ec1ea93058fe39b8ddfd8db1c7a18081db397c50a0b8ec77d",
                 "mynamespace",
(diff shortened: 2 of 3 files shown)


exporters/webhook/models/pelorus_webhook.py
```diff
@@ -42,14 +42,15 @@ class PelorusPayload(BaseModel):
...
+                         is different from the time when the webhook
+                         could have been received.
     """
 
     # Even if we consider git project name as app, it still should be below 100
     app: str = Field(max_length=200)
-    # ISO 8601 is 19 chars long, so don't expect any format can be longer then 50 chars
-    timestamp: str = Field(max_length=50)
+
+    timestamp: str = Field(regex=r"^\d{10}$")  # 10 digit number
```
mateusoliveira43, line 53: using `int` with `gt` and `lt` wouldn't do the trick as well? Just so we do not use regex

mpryc: Yes, but we allow this to be a string of 10 digits as well. What's wrong with regex?

mateusoliveira43: I think if is int, it accepts both 3 and "3"

We need to interpret it (in this case that is small, no big trouble)

mpryc: It's same vice-versa. I know it's not correct to think about possible future implementation, but we may want to change this to consume other types of timestamp formats adding other validators and then this will be required to be str.

mpryc: Tested with:
```python
    timestamp: int = Field(ge=1577840461, le=2524611661) # between 2020 and 2050.
```

Got the error on one of the tests:
```shell

        query_result = REGISTRY.get_sample_value(
            "test_committime_metrics",
            labels=metric_labels,
        )

>       assert query_result == timestamp
E       AssertionError: assert 1234567890 == '1234567890'

exporters/tests/test_in_memory_metric.py:108: AssertionError
```

The prometheus is returning str, while we pass int. We could have this, but still prefer str over int as we do str(int) at some point when storing data to prometheus anyway, so why not to accept this from payload.

mpryc: Actually it should be:

    timestamp: int = Field(None, ge=1577840461, le=2524611661) # between 2020 and 2050.

Otherwise the ge and le is not taken into account that's we have not allowed '1234567890', but still str over int in my opinion.

mpryc:
/hold

Need to check int / str

exporters/webhook/models/pelorus_webhook.py
mpryc: Ok agree to change to int, if needed we can always switch back to str. Having int added some date constrains to get some reasonable date validation. Think the 2060 as the gt limit year is fine at this point (thinking of  Y2K bug... but I am pretty sure this data won't have lifespan of 35+ years from now).

mateusoliveira43 (approved):

PR #883, opened 2023-03-09
fix: Worflows permissions to run integration tests

### LINKED ISSUE #838: Integration tests credentials
The integration-tests.yml job is being run without credentials during PRs (which causes the tests to be skipped). This should not occur

mateusoliveira43:

Allow forks to run integration tests (use repository secrets).

## Linked Issues

resolves #838
depends on #948

## Testing Instructions

Check integration tests CI job logs.


mateusoliveira43:
/hold
trying to understand why CI job was not triggered

mateusoliveira43:
I do not know why, but `pull_request_target` did not work.

`workflow_run` did not work as well, but as I read, once it is on the default branch (`master` in our case), it will work (and saw an example of usage). But the way I did, it takes more time to run it (upload and download zip file).

I would let the way it is (only running on `push`) if we can not find a way that does not downgrade the CI performance.

References:
- https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
- https://github.com/tiangolo/fastapi/blob/master/.github/workflows/preview-docs.yml

PR #885, opened 2023-03-10
Ensure chart versions are in sync between each other

mpryc:
The Chart versions should be the same across different files in the base helm charts directory and the
pelorus-operator.

@redhat-cop/mdt


mpryc:
This PR will pass, because we do not modify any charts/* folder, however next one after this merge will need to sync versions between different charts properly.

This follows discussion #811 around keeping the same versions within charts.

scripts/chart-test.sh
```diff
@@ -21,3 +21,19 @@ git fetch "$REMOTE" --unshallow &> /dev/null
...
+FILES_VERSIONS=$(find charts/ pelorus-operator/helm-charts/ -type f -name 'Chart.yaml' -exec grep -H '^version: ' {} \;)
+VERSIONS=$(echo "$FILES_VERSIONS" | cut -d ' ' -f 2 | sort | uniq)
+NO_VERSIONS=$(echo "$VERSIONS" | wc -l)
+if [[ "$NO_VERSIONS" -eq 1 ]]; then
+    echo "All chart versions are in-sync. Chart version: $VERSIONS"
+else
+  echo "ERROR: Found different versions in Chart.yaml files:"
+  echo "$FILES_VERSIONS"
+  exit 1
+fi
```
mateusoliveira43, line 39: Questions:
- Now we are also checking  `charts/operators` folder, but they are not used by Pelorus Operator, right?
  I would simply delete the `charts/operators` folder (did not test if this breaks anything!)
- We are also checking `pelorus-operator` folder, this means that each time we make changes in `charts` folder, we need to run create operator script. I still want to delete that folder from our source code (because of the duplication of `charts` folder). What are the pros of running this new check?

mpryc: It's not really relevant. There should be one version across all of the Chart.yaml files.

Yes, each time we make the change or version bump in the charts/* folder a new rc.X version of the pelorus-operator rc.X should also be created. Later when we will have automation to create pelorus-operator:x.y.z-rc.X and relevant operator-bundle it will be easier to test merged dev changes with that process. Of course we won't have same major.minor.patch for the pelorus-operator as we are at 0.0.5 vs 2.0.8

mateusoliveira43: As we talked, we are not deleting `pelorus-operator` folder at the moment. I will create an issue about the first bullet, so we can discuss it better.

But I would change the approach of the command: instead of doing grep in `^version:`, I would run grep on `2\.0\.x`, x being the last version (this would need to be dynamic, but we can use something like `git describe --tag --abbrev=0` to achieve this), and if there are matches, break.

Examples:

- When I create PR and change `2.0.8` to `2.0.9-rc.1`, if there are still `2.0.8` matches in `charts/` or `pelorus-operator/` folders, it should break.

- When I create PR and change `2.0.9-rc.1` to `2.0.9-rc.2`, if there are still `2.0.9-rc.1` matches in `charts/` or `pelorus-operator/` folders, it should break.

what you think?

mpryc: Storing a part of the version as an input to the grep command is asking for trouble in the future.

This is simple verification if the version does match and nothing more. Another function or mechanism (which in fact I am implementing at the moment) is to ensure `make pre-release` `make release` `make major release` does properly create versions and updates them on top of what's released / pre-released.

mpryc:
/hold

I don't like implementing multiple features in one PR, but will do this time. Following update on this PR will:
 - ensure prometheus and grafana are in sync with pelorus-operator
 - enable this action on the changes of the both pelorus-operator/ and helm/ subfolders

charts/operators/values.yaml
```diff
@@ -1,2 +1,3 @@
-# empty
 ---
+grafana_subscription_version: 4.8.0
```
mateusoliveira43, line 2: how do we test these changes in `charts/operators`?

pelorus-operator/Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.5
+VERSION ?= 0.0.6-rc.1
```
mateusoliveira43, line 6: how this was generated?

For the next release we need to change how we invoke the script?

scripts/chart-test.sh
mateusoliveira43: Yeah, I agree about the first part.

But I think it would be interesting to add to the docs that developer must run `curl https://raw.githubusercontent.com/konveyor/pelorus/master/charts/pelorus/Chart.yaml 2> /dev/null | grep "^version:" | cut -c 10-` (this gives the current version of charts in master) and then run `grep -Iinr "VERSION" ./charts/` to check if all previous versions were updated (I do not want to scriptize this, because I think this can break in the future).

I have run `grep -Iinr "[0-9]\{1,\}\.[0-9]\{1,\}\." ./charts/` to check if there are any missing updates in this PR, and changed this (did not create operator image to check if these changes break anything)
```diff
diff --git a/charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml b/charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml
index 06e05a1..96d5bbc 100644
--- a/charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml
+++ b/charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml
@@ -17,7 +17,7 @@ spec:
   source:
     contextDir: {{ .source_context_dir | default "exporters/" }}
     git:
-      ref: {{ .source_ref | default "v2.0.8" }}
+      ref: {{ .source_ref | default "v2.0.9-rc.1" }}
       uri: {{ .source_url | default "https://github.com/konveyor/pelorus.git"}}
     type: Git
   strategy:
diff --git a/charts/pelorus/subcharts/exporters/templates/_deploymentconfig.yaml b/charts/pelorus/subcharts/exporters/templates/_deploymentconfig.yaml
index 1f07b01..c88b56d 100644
--- a/charts/pelorus/subcharts/exporters/templates/_deploymentconfig.yaml
+++ b/charts/pelorus/subcharts/exporters/templates/_deploymentconfig.yaml
@@ -60,7 +60,7 @@ spec:
 
             {{- if and (not .source_ref) (not .source_url) }}
             - name: PELORUS_IMAGE_TAG
-              value: {{ .app_name }}:{{ .image_tag | default "v2.0.7" }}
+              value: {{ .app_name }}:{{ .image_tag | default "v2.0.9-rc.1" }}
             {{- end }}
 
             {{- if .extraEnv }}
diff --git a/charts/pelorus/subcharts/exporters/templates/_imagestream_from_image.yaml b/charts/pelorus/subcharts/exporters/templates/_imagestream_from_image.yaml
index c0a533c..5f30c4d 100644
--- a/charts/pelorus/subcharts/exporters/templates/_imagestream_from_image.yaml
+++ b/charts/pelorus/subcharts/exporters/templates/_imagestream_from_image.yaml
@@ -20,10 +20,10 @@ spec:
       name: {{ .image_name }}
     # .image_name is provided without tag
     {{- else }}
-      name: {{ .image_name }}:{{ .image_tag | default "v2.0.8" }}
+      name: {{ .image_name }}:{{ .image_tag | default "v2.0.9-rc.1" }}
     {{- end }}
 {{- else }}
-      name: quay.io/pelorus/pelorus-{{ .exporter_type }}-exporter:{{ .image_tag | default "v2.0.8" }}
+      name: quay.io/pelorus/pelorus-{{ .exporter_type }}-exporter:{{ .image_tag | default "v2.0.9-rc.1" }}
 # .image_name
 {{- end }}
     name: {{ .image_tag | default "stable" }}
```

If we apply this change, then I believe the command I suggest to add in the docs will work

pelorus-operator/Makefile
mateusoliveira43: understood

I would remove the changes made to `pelorus-operator` folder then, apply then in the next PR

mpryc: We can't as there is already check to ensure charts are in sync, so had to manually update those.

mpryc: https://github.com/dora-metrics/pelorus/pull/885/files#diff-63c82fa5dd1fbb3eef9b134e7f94495ad1b7e961e61efd5e27004c55b5abf70b

will have very soon follow up and the clean process, but don't want to make that huge PR...

charts/operators/values.yaml
mpryc: Not this PR.

mateusoliveira43 (approved):

charts/pelorus/Chart.yaml
```diff
@@ -14,9 +14,9 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.0.8
+version: 2.0.9-rc.1
 
 dependencies:
   - name: exporters
-    version: 2.0.8
+    version: "v2.0.9-rc.1"
```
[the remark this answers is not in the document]
mpryc: We are releasing github versions with v, but here it's really not necessary. Will update the script to not include the v here.

mateusoliveira43 (approved):

mpryc:
merging as we know CI is flake at the moment with integration tests.

PR #920, opened 2023-03-31
Build exporter images with the tag from updated _imagestream_from_image

mpryc:
This allows to set proper tags whenever our chart points to a new tag.

The change moves some common action code to a template making less code duplication.

### DIFF
--- a/.github/workflow_templates/build_push_exporters/action.yml
+++ b/.github/workflow_templates/build_push_exporters/action.yml
@@ -1,0 +1,53 @@
+name: build_push_exporters
+
+inputs:
+  image: 
+    required: true
+  builder_image:
+    required: true
+  s2i_log_level:
+    required: true
+  quay_imageregistry:
+    required: true
+  quay_imagenamespace:
+    required: true
+  quay_username:
+    required: true
+  quay_password:
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Set Image version
+      shell: bash
+      run: |
+        IMG_VER=$(grep quay.io charts/pelorus/subcharts/exporters/templates/_imagestream_from_image.yaml | grep -oP '(?<=default ")[^"]+')
+        echo "Image version: $IMG_VER"
+        echo "VERSION=$IMG_VER" >> $GITHUB_ENV
+
+    # Setup S2i and Build container image
+    - name: Setup and Build
+      id: build-exporter
+      uses: redhat-actions/s2i-build@v2
+      with:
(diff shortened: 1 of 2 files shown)


mpryc:
/hold

testing

.github/workflow_templates/build_push_exporters/action.yml
```diff
@@ -0,0 +1,53 @@
...
+  quay_password:
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Set Image version
+      shell: bash
+      run: |
+        IMG_VER=$(grep quay.io charts/pelorus/subcharts/exporters/templates/_imagestream_from_image.yaml | grep -oP '(?<=default ")[^"]+')
```
mateusoliveira43, line 25: why get the version from this file and not `charts/pelorus/Chart.yaml`?

mateusoliveira43: and this version isn't the same for every exporter? why get it every time instead of just getting it before the jobs?

.github/workflows/build-pelorus.yml
```diff
@@ -7,156 +7,103 @@ on:
     paths:
       - 'exporters/**'
       - 'charts/pelorus/subcharts/exporters/templates/_buildconfig.yaml'
+      - 'charts/pelorus/subcharts/exporters/templates/_imagestream_from_image.yaml'
```
mateusoliveira43, line 10: couldn't be `'charts/**'`? what special about these 2 files?

.github/workflow_templates/build_push_exporters/action.yml
mpryc: The versions should be the same, but we reference the quay image tag in the `_imagestream_from_image.yaml` so I want to make this file source of truth not the Chart.yaml. Imagine one will bump the Chart.yaml without `_imagestream_from_image.yaml` which should not happen but it's possible, this way the tag will be wrong.

.github/workflows/build-pelorus.yml
mpryc: I am not sure if we want to re-build images on every charts change. E.g. what if we modify only some text file?

Those two files are entry points to the images, one for image from quay and another to create build from the source.

Even if the _buildconfig.yaml references master I still believe we should recreate images as one may want to create new image with same tag.

Other files are really not relevant for the exporter image build process and we should not create them on every change.

mateusoliveira43 (approved):

PR #926, opened 2023-04-05
Add secret to secure payload for the webhook exporter

### LINKED ISSUE #922: Add secret for the webhook exporter
based on: https://docs.github.com/en/webhooks-and-events/webhooks/securing-your-webhooks

We should add secret to our webhook exporter.

HMAC-SHA256 represented by 'sha256=' prefix followed by hexadecimal 64 characters (32 bytes x 2 hex digits per byte)

Above is used in various Webhooks including GitHub and our webhook should use similar.

_Note the "HTTP Message Signatures" specification, however it's draft:
https://datatracker.ietf.org/doc/draft-ietf-httpbis-message-signatures/_

mpryc:
Adds option to configure webhook exporter to share SECRET_TOKEN to validate payload integrity.

With the webhook configured to use SECRET_TOKEN the sender is obligated to calculate the SHA256 hash of the payload and include that hash as a header during POST method, otherwise the payload will not be received and recorded by the webhook.

resolves #922 

## Testing Instructions

1. Deploy Pelorus with the webhook exporter that does have SECRET_TOKEN configured
2. Send sample data using curl to the webhook exporter endpoint as described in the examples section of the documentation

Alternatively use python to start the webhook exporter previously exporting SECRET_TOKEN:

```shell
$ make dev-env
$ source .venv/bin/activate
$ export SECRET_TOKEN="My Secret Token"
$ python exporters/webhook/app.py
```

And sent from another shell sharing the same SECRET_TOKEN the payload as described in the documentation to the `http://localhost:8080/pelorus/webhook` endpoint. 

```shell
$ cd exporters/tests/data
$ export SECRET_TOKEN="My Secret Token"
$ SHA256_HASH_SIGNATURE=$(echo -n $(cat "./webhook_pelorus_committime.json") | openssl dgst -sha256 -hmac "${SECRET_TOKEN}"| cut -d ' ' -f 2)

# Validate if hash was correctly calculated
$ echo $SHA256_HASH_SIGNATURE
789c5630940ee8ede816709f6535cf576457e7b625b7b1c64ec01b8db358c6cc


$ curl -X POST http://localhost:8080/pelorus/webhook -d @./webhook_pelorus_committime.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: committime"  -H "X-Hub-Signature-256: sha256=${SHA256_HASH_SIGNATURE}"
{"http_response":"Webhook Received","http_response_code":200}

# Try to send data with no 'X-Hub-Signature-256'
$ curl -X POST http://localhost:8080/pelorus/webhook -d @./webhook_pelorus_committime.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: committime" 
{"detail":"Non existing signature."}

# Try to send data with wrong format signature
$ curl -X POST http://localhost:8080/pelorus/webhook -d @./webhook_pelorus_committime.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: committime"  -H "X-Hub-Signature-256: sha256=fake_signature"
{"detail":"Improper headers."}

# Try to send data with mismatch signature
$ export SHA256_HASH_SIGNATURE=111111111111e8ede816709f6535cf576457e7b625b7b1c64ec01b8db358c6cc
$ curl -X POST http://localhost:8080/pelorus/webhook -d @./webhook_pelorus_committime.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: committime"  -H "X-Hub-Signature-256: sha256=${SHA256_HASH_SIGNATURE}"
{"detail":"Invalid signature."}

# Try to calculate signature using jq:
$ SHA256_HASH_SIGNATURE=$(jq -c "" "./webhook_pelorus_committime.json" | openssl dgst -sha256 -hmac "${SECRET_TOKEN}"| cut -d ' ' -f 2)

$ echo $SHA256_HASH_SIGNATURE
3ce994968588a4d69f8732b1cde8c1d3581b38ce7bd97275ed9b0f61a1f2fc78

$ curl -X POST http://localhost:8080/pelorus/webhook -d @./webhook_pelorus_committime.json -H "Content-Type: application/json" -H "User-Agent: Pelorus-Webhook/test" -H "X-Pelorus-Event: committime"  -H "X-Hub-Signature-256: sha256=${SHA256_HASH_SIGNATURE}"
{"http_response":"Webhook Received","http_response_code":200}
```

Review if the data is received by opening web page `http://localhost:8080/metrics`

### DIFF
--- a/docs/GettingStarted/configuration/ExporterWebhook.md
+++ b/docs/GettingStarted/configuration/ExporterWebhook.md
@@ -49,7 +49,15 @@ This is the list of options that can be applied to `env_from_secrets`, `env_from
 
 | Variable | Required | Default Value |
 |----------|----------|---------------|
+| [SECRET_TOKEN](#secret_token) | no | - |
 | [LOG_LEVEL](#log_level) | no | `INFO` |
+
+###### SECRET_TOKEN
+
+- **Required:** no
+- **Type:** string
+
+: Set the secret token to ensure that the webhook receives only the intended payload. This secret token is used by the sender of the payload to calclate hash signature, which then is included with the headers of each request as `X-Hub-Signature-256`. Please refer to the [Securing Webhook](#securing-webhook) section for examples.
 
 ###### LOG_LEVEL
 
@@ -107,6 +115,8 @@ For the Header X-Pelorus-Event: `failure`
 
 
 ## Example usage
+
+### Sending non-secure payload
 
 You can easily send a POST request using [Curl](https://curl.se) directly from the shell. You can store the payload data in a file or pass it as an argument to [Curl](https://curl.se). Below is an example of sending several requests to cover the lifecycle of an application:
 
@@ -200,4 +210,53 @@ $ curl -X POST <Webhook route URI>/pelorus/webhook \
        -H "X-Pelorus-Event: failure" \
        -H "Content-Type: application/json" \
        -d @./mongo_production_failure_resolved.json
+```
+
+### Securing payload
+
+In order to ensure the security of your webhook payload, it is important to configure your webhook exporter to use a [SECRET_TOKEN](#secret_token).
+
+This token value is necessary to calculate the SHA256 hash signature and include it in the headers of each request with the 'X-Hub-Signature-256' POST header.
+
+You may utilize your own tools to calculate the signature, the following example uses two **shell** methods to do so.
+
+ - the first method uses `echo -n` together with `openssl` CLI
+
+```shell
+$ export PELORUS_METRIC_FILE="./mongo_committime.json"
+$ export SECRET_TOKEN="My Secret Token"
+$ SHA256_HASH_SIGNATURE=$(echo -n $(cat "${PELORUS_METRIC_FILE}") | openssl dgst -sha256 -hmac "${SECRET_TOKEN}"| cut -d ' ' -f 2)
+
+# Check calculated token:
+$ echo "${SHA256_HASH_SIGNATURE}"
+29a4606fccb6976cd43cb8a6cf210d402294e08fffda86f87c96fdb39d9fbc3c
+```
+
+ - the second method uses `jq` together with `openssl` CLI
+
+```shell
+$ export PELORUS_METRIC_FILE="./mongo_committime.json"
+$ export SECRET_TOKEN="My Secret Token"
+$ SHA256_HASH_SIGNATURE=$(jq -c "" "${PELORUS_METRIC_FILE}" | openssl dgst -sha256 -hmac "${SECRET_TOKEN}"| cut -d ' ' -f 2)
+
+# Check calculated token:
+$ echo "${SHA256_HASH_SIGNATURE}"
+9f6140a622159a4e1d379603afe6a0d288d45bc9ad99943c29ec77e3f2cf1cee
+
+```
+
+   > **NOTE:** It is worth noting that the SHA256 signatures are different in the above example, but the Pelorus webhook accepts both. This is because the Pelorus uses best effort to match the payload hash based on its content, ignoring new lines and json separators, that may vary depending on the tooling used to calculate payload signatures.
+
+
+Sending the payload from the directory where `*.json` files are with inclusion of the calculated SHA256 value:
+
+```shell
+$ curl -X POST <Webhook route URI>/pelorus/webhook \
+       -H "User-Agent: Pelorus-Webhook/test" \
+       -H "X-Pelorus-Event: committime" \
+       -H "Content-Type: application/json" \
+       -d @./mongo_committime.json
+       -H "X-Hub-Signature-256: sha256=${SHA256_HASH_SIGNATURE}"
+
+{"http_response":"Webhook Received","http_response_code":200}
 ```
--- a/exporters/tests/test_pelorus_webhook_handler.py
+++ b/exporters/tests/test_pelorus_webhook_handler.py
@@ -15,6 +15,8 @@
 #
 
 
+import hashlib
+import hmac
 import http
 import json
 from unittest.mock import AsyncMock, patch
@@ -31,7 +33,10 @@ from webhook.models.pelorus_webhook import (
     PelorusMetricSpec,
     PelorusPayload,
 )
-from webhook.plugins.pelorus_handler import PelorusWebhookHandler
+from webhook.plugins.pelorus_handler import (
+    PelorusWebhookHandler,
+    _verify_payload_signature,
+)
 from webhook.plugins.pelorus_handler_base import Headers, HTTPException
 
 
@@ -244,3 +249,64 @@ async def test_pelorus_receive_pelorus_payload_error(headers, json_payload):
         await handler._receive_pelorus_payload(json_payload_data)
     assert http_exception.value.detail == "Invalid payload."
     assert http_exception.value.status_code == http.HTTPStatus.UNPROCESSABLE_ENTITY
+
+
+@pytest.mark.parametrize(
+    "json_payload_data_bytes",
+    [
+        b'{"data": "value", "data2": "value2"}',
+        b'{"data":"value","data2":"value2"}',
+        b'{"data":"value", "data2":"value2"}',
+        b'{"data" :"value","data2" :"value2"}',
+        b'{"data" :"value", "data2" :"value2"}',
+        b'{ "data": "value", "data2": "value2" }',
+        b'{ "data":"value","data2":"value2" }',
+        b'{ "data":"value", "data2":"value2" }',
+        b'{ "data" :"value","data2" :"value2" }',
+        b'{ "data" :"value", "data2" :"value2" }',
+        b'{"data": "value", "data2": "value2"}\n',
+        b'{"data":"value","data2":"value2"}\n',
+        b'{"data":"value", "data2":"value2"}\n',
+        b'{"data" :"value","data2" :"value2"}\n',
+        b'{"data" :"value", "data2" :"value2"}\n',
+        b'{ "data": "value", "data2": "value2" }\n',
+        b'{ "data":"value","data2":"value2" }\n',
+        b'{ "data":"value", "data2":"value2" }\n',
+        b'{ "data" :"value","data2" :"value2" }\n',
+        b'{ "data" :"value", "data2" :"value2" }\n',
+    ],
+)
+def test_verify_payload_signature_different_json(json_payload_data_bytes):
+    """
+    Verifies if the json payload was properly verified based on provided json string.
+
+    Even if the string of the payload format is different we should still find a match as the
+    format does not really affect it's content.
+    """
+    json_payload_data = {"data": "value", "data2": "value2"}
+    secret = b"My Secret"
+    calculated_hash = (
+        "sha256="
+        + hmac.new(secret, json_payload_data_bytes, hashlib.sha256).hexdigest()
+    )
+
+    assert _verify_payload_signature(secret, calculated_hash, json_payload_data) is True
+
+
+@pytest.mark.parametrize(
+    "secret,expected_signature,json_payload_data",
+    [
+        (
+            b"MySecret",
+            "sha256=f1dbf8a5d2aa74fb479c6bab52d80e947c66c98c131bb2fcfe97a6912623b05d",
+            {"data": "value", "data2": "value2"},
+        ),
+    ],
+)
+def test_verify_payload_not_matching_hash(
+    secret, expected_signature, json_payload_data
+):
+    assert (
+        _verify_payload_signature(secret, expected_signature, json_payload_data)
+        is False
+    )
--- a/exporters/webhook/plugins/pelorus_handler.py
+++ b/exporters/webhook/plugins/pelorus_handler.py
@@ -14,9 +14,12 @@
 #    under the License.
 #
 
+import hashlib
+import hmac
 import http
+import json
 import logging
-from typing import Any, Awaitable
+from typing import Any, Awaitable, Dict
 
 from pydantic import ValidationError, parse_obj_as
 from typing_extensions import override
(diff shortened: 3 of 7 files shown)


exporters/webhook/plugins/pelorus_handler.py
```diff
@@ -71,16 +164,12 @@ class PelorusWebhookHandler(PelorusWebhookPlugin):
...
-    Attributes:
-        handshake_headers: (Headers): Headers that are received by the webhook.
-        request: (Request): The request object associated with the webhook.
     """
 
     user_agent_str = "Pelorus-Webhook/"
 
-    def __init__(self, handshake_headers: Headers, request: Request) -> None:
-        super().__init__(handshake_headers, request)
+    def __init__(self, *args, **kwargs) -> None:
```
mateusoliveira43, line 171: I prefer to be explicit about each parameter

docs/GettingStarted/configuration/ExporterWebhook.md
```diff
@@ -49,8 +49,16 @@ This is the list of options that can be applied to `env_from_secrets`, `env_from
 
 | Variable | Required | Default Value |
 |----------|----------|---------------|
+| [SECRET_TOKEN](#secret_token) | no | - |
 | [LOG_LEVEL](#log_level) | no | `INFO` |
 
+###### SECRET_TOKEN
+
+- **Required:** no
+- **Type:** string
```
mateusoliveira43, line 58: Could not be `bool`? I only saw it being used in code as `if secret`. I feel `bool` is easier to type than `None | str`

mpryc: No, the secret is string value and it needs to be specified by the user.

exporters/webhook/plugins/pelorus_handler.py
mpryc: This is plugin implementation the explicit values are in the upper class, so I actually prefer to accept anything that is implemented on the upper class.

docs/GettingStarted/configuration/ExporterWebhook.md
mateusoliveira43: where this is used in the code?

exporters/webhook/plugins/pelorus_handler.py
```diff
@@ -35,10 +38,100 @@
...
+    { "data":"value","data2":"value2" }\n
+    { "data":"value", "data2":"value2" }\n
+    { "data" :"value","data2" :"value2" }\n
+    { "data" :"value", "data2" :"value2" }\n
+
+    Returns:
+        bool: True when the matching hash was found, False otherwise
+    """
+
+    # Formats:
```
mateusoliveira43, line 86: is not this comment in the docstrings? lines 61-80

exporters/webhook/plugins/pelorus_handler.py
```diff
@@ -35,10 +38,100 @@
...
+    ]
+    additional_transformation = [
+        "extra_spaces_no_newline",
+        "extra_spaces_newline",
+        "no_extra_spaces_no_newline",
+        "no_extra_spaces_newline",
+    ]
+
+    for separator in separator_formats:
+        for idx, transformation in enumerate(additional_transformation):
```
mateusoliveira43, line 107: remove enumerate and index, they are not used

docs/GettingStarted/configuration/ExporterWebhook.md
mpryc: When you start app it's visible as REDACTED:

```
04-05-2023 14:13:58 INFO     Registered webhook plugin for user-agent: 'pelorus-webhook/'
04-05-2023 14:13:58 INFO     Loading WebhookCollector, inputs below:
04-05-2023 14:13:58 INFO     app_label='app.kubernetes.io/name', default value; APP_LABEL was not set
04-05-2023 14:13:58 INFO     secret_token=REDACTED, from env var SECRET_TOKEN
```

When unset:
```
04-05-2023 15:04:56 INFO     Registered webhook plugin for user-agent: 'pelorus-webhook/'
04-05-2023 15:04:56 INFO     Loading WebhookCollector, inputs below:
04-05-2023 15:04:56 INFO     app_label='app.kubernetes.io/name', default value; APP_LABEL was not set
04-05-2023 15:04:56 INFO     secret_token=REDACTED, default value; SECRET_TOKEN was not set
```

exporters/webhook/plugins/pelorus_handler.py
```diff
@@ -35,10 +38,100 @@
...
+            # no_extra_spaces_no_newline - default
+            payload_json_string = json.dumps(
+                json_payload_data, separators=separator, indent=None
+            )
+            payload_json_string.rstrip("\n")
+
+            if transformation == "extra_spaces_no_newline":
```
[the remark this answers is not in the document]
mpryc: I think that would make things less visible/clear by reading the code.

exporters/webhook/plugins/pelorus_handler.py
mpryc: Sort of, the comment contains all the variations, here for visibility only those that are within separator_formats, so maybe a bit of duplication, but makes the code readable.

docs/GettingStarted/configuration/ExporterWebhook.md
mpryc: This has nothing to do with the authentication TOKEN.

This is a value used to sign the payload. See similarly how github webhook is securing payload, we do quite same thing, except our payload must be signed by the sender and in the github example it's the github service that is signing it before sending.

https://docs.github.com/en/webhooks-and-events/webhooks/securing-your-webhooks#setting-your-secret-token

It is being used in the code see a new method in the app.py:
```
def _get_hash_token() -> str:
    return collector.secret_token
```

This method is used in the line
```
    handler = webhook_handler(request.headers, request, secret=_get_hash_token())
```

It needs to be a method that takes a value from a collector object, otherwise it's really hard to write test cases that mocks entire collector which is an async func.

exporters/webhook/plugins/pelorus_handler.py
mateusoliveira43: `for transformation in additional_transformation:` does the same thing without storing an additional variable

mpryc: I could simply remove enumeration and do:
```
    for separator in separator_formats:
        for transformation in additional_transformation:
```

mpryc: correct, will modify

exporters/webhook/plugins/pelorus_handler.py
```diff
@@ -35,10 +38,100 @@
...
+        "no_extra_spaces_newline",
+    ]
+
+    for separator in separator_formats:
+        for idx, transformation in enumerate(additional_transformation):
+            # no_extra_spaces_no_newline - default
+            payload_json_string = json.dumps(
+                json_payload_data, separators=separator, indent=None
+            )
+            payload_json_string.rstrip("\n")
```
mateusoliveira43, line 112: I think this line does not do anything, can you check it out?

I believe you would have to do `payload_json_string = payload_json_string.rstrip("\n")` to give the wanted result

mpryc: Good catch ! You are correct. Fixed.

docs/GettingStarted/configuration/ExporterWebhook.md
mateusoliveira43: As we discussed, it's value is being used, so everything is ok

exporters/tests/test_pelorus_webhook_handler.py
```diff
@@ -244,3 +247,162 @@ async def test_pelorus_receive_pelorus_payload_error(headers, json_payload):
         await handler._receive_pelorus_payload(json_payload_data)
     assert http_exception.value.detail == "Invalid payload."
     assert http_exception.value.status_code == http.HTTPStatus.UNPROCESSABLE_ENTITY
+
+
+# fmt: off
```
mateusoliveira43, line 252: what are these comments?

exporters/tests/test_pelorus_webhook_handler.py
```diff
@@ -244,3 +247,162 @@ async def test_pelorus_receive_pelorus_payload_error(headers, json_payload):
...
+    [
+        (
+            b"My Secret",
+            "sha256=8b1e67669068057d751c49ca128fcf3b112dad8f26d2d00ece451c51b18351c1",
+            {"data": "value", "data2": "value2"},
+        ),
+        (
+            b"My Secret",
+            "sha256=8b1e67669068057d751c49ca128fcf3b112dad8f26d2d00ece451c51b18351c1",
+            {"data":"value","data2":"value2"}, # noqa
```
mateusoliveira43, line 264: and these comments?

exporters/tests/test_pelorus_webhook_handler.py
```diff
@@ -244,3 +247,162 @@ async def test_pelorus_receive_pelorus_payload_error(headers, json_payload):
...
+        (
+            b"My Secret",
+            "sha256=8b1e67669068057d751c49ca128fcf3b112dad8f26d2d00ece451c51b18351c1",
+            { "data" :"value", "data2" :"value2" }, # noqa
+        ),
+    ],
+)
+# fmt: on
+def test_verify_payload_signature_different_dict(
+    secret, expected_signature, json_payload_data
```
mateusoliveira43, line 310: parametrize here only `json_payload_data`, the other values are the same

exporters/tests/test_pelorus_webhook_handler.py
```diff
@@ -244,3 +247,162 @@ async def test_pelorus_receive_pelorus_payload_error(headers, json_payload):
...
+        ),
+        (
+            b"My Secret",
+            "sha256=2f6e3dde9a092d1fbce03bfbb0cb5d0b1db4419ab1e525a99684fd75b9ec8abb",
+            { "data":"value","data2":"value2" }, # noqa
+        ),
+    ],
+)
+# fmt: on
+def test_verify_payload_signature_different_hash(
```
mateusoliveira43, line 361: what this test does that the previous does not?

exporters/tests/test_pelorus_webhook_handler.py
mpryc: pylama will complain that the dict is improperly formatted, but that's what we want in this test.

exporters/tests/test_pelorus_webhook_handler.py
mpryc: The tests are very similar, however this one exercises different hashes instead of one. Different hashes mapping to different dictionaries formats. If you look closely at the comment of this test you can see that the hashes were calculated using 2 different methods on 2 different formats. This matches our documentation and will catch eventual issues.

In theory the previous test should be sufficient but we have 1-n mapping there, where here we have a better matrix.

mpryc: Also additional note here. We catch additional cases where the hash was calculated using shell method that produces the newline at the end of the file `echo {}` without `-n` flag. This is not covered by previous test.

exporters/tests/test_pelorus_webhook_handler.py
mpryc: From the black documentation:
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

"Black reformats entire files in place. It doesn’t reformat  [...]  blocks that start with # fmt: off and end with # fmt: on.

exporters/webhook/plugins/pelorus_handler.py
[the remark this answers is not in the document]
mateusoliveira43: line 106 is unnecessary? I commented it and no test broke

How it would like with the change I commented before
```python
    separator_formats = [
        (", ", ": "),
        (",", ":"),
        (", ", ":"),
        (",", " :"),
        (", ", " :"),
    ]
    transformations = [
        lambda text: text,  # default
        lambda text: text + "\n",
        lambda text: f"{{ {text[1:-1]} }}",
        lambda text: f"{{ {text[1:-1]} }}\n",
    ]

    for separator in separator_formats:
        for transformation in transformations:
            payload_json_string = transformation(
                json.dumps(json_payload_data, separators=separator, indent=None)
            )

            encoded_payload = payload_json_string.encode("utf-8")

            sha256_signature = (
                "sha256="
                + hmac.new(secret, encoded_payload, hashlib.sha256).hexdigest()
            )

            if hmac.compare_digest(sha256_signature, signature_secret):
                return True
    return False
```

mateusoliveira43 (approved):

exporters/webhook/plugins/pelorus_handler.py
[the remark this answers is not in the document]
mpryc: Line 106 is necessary for  the first case 'extra_spaces_no_newline' so we remove newline if was there.

I prefer to not not have such nested function, as it may be harder to debug imo.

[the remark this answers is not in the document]
mateusoliveira43: then why no tests broke when we comment it?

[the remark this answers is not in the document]
mateusoliveira43: I think `json.dumps` (with the parameters we use) do not create string with line break, and that's why the tests are not breaking (and line 106 is not necessary)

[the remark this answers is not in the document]
mpryc: Maybe, but it doesn't hurt to add one to not be dependent on the json.dumps() return format. There is nothing in the documentation of dump() nor dumps() which says this is the case.

PR #929, opened 2023-04-06
Pelorus-Operator 0.0.6 with fresh patches and updated release scripts

mpryc:
This change includes Pelorus-Operator 0.0.6 which refreshes patches, updated linting of the charts so they now must match the pelorus-operator charts.

Also it consists of change where failing patches were allowing to proceed with pelorus-operator creation.

To create release it's enough to type: make release

The RC releases are still two step as the create_pelorus_operator must have explicit rc version to be passed for the operator rc version.


mateusoliveira43:
Can you provide a operator test image for this PR?

mateusoliveira43 (approved):

Issue #934, opened 2023-04-14
Webhook exporter metric should match format of other exporters


mpryc:
Currently webhook deploytime metric is different from deploytime exporter deploytime metric (the difference is storing exactly timestamp and storing the received timestamp by Prometheus). 

There should be unique way of representing different events.

Related: #933

Issue #933, opened 2023-04-14
Consider dropping / not accepting events that are above threashold


mpryc:
Problem:
1. Currently some of the metrics are registered with and some without timestamp.
2. Some of the metrics are not registered by Prometheus, because the event happened above accepted threashold
3. For a new Pelorus exporter deployment the time of the event without timestamp is the time when the exporter was created and not actual time of the event, so if we have a failure that occurred a year ago it will be visible as current failure. This may be problematic to calculate our **Mean Time to Restore** or **Change Failure Rate** properly for a given time range.

For some of the metrics a timestamp is a value, here is the list:
 
Failure (Create / Resolution events) - timestamp as a value
```
    failure_creation_timestamp (app, issue_number, timestamp)
    failure_resolution_timestamp (app, issue_number, timestamp)
```

Commit Time
```
    commit_timestamp (namespace, app, commit_hash, image_sha) timestamp
```

Deploy Time
```
    deploy_timestamp (namespace, app, image_sha) timestamp
    deployment_active (namespace, app, image_sha)
```

Prometheus accepts metrics which are without timestamps or the timestamp is within it's threshold, with it's error:
```
"Error on ingesting samples that are too old or are too far into the future"
```

mpryc:
Corresponding PR for the webhook deploytime:
#943

PR #941, opened 2023-04-17
Add the timestamp to the webhook exporter.

mpryc:
Currently a valid event is the one that happened within 30min from the time of sending that event to our webhook exporter. This is the limitation of Prometheus that is not accepting events that are too old or are too far into the future.

Fixes also webhook exporter to contain timestamps for the events that occured.

Fixes: #934 
Fixes: #933

### DIFF
--- a/docs/GettingStarted/configuration/ExporterWebhook.md
+++ b/docs/GettingStarted/configuration/ExporterWebhook.md
@@ -90,7 +90,7 @@ For the Header X-Pelorus-Event: `deploytime`
 | `app`       | `string` | Monitored application name |
 | `image_sha` | `string` | Image SHA used for deployment. Must be prefixed with the `sha256:` followed by 64 characters of small letters and numbers |
 | `namespace` | `string` | OpenShift namespace to which application was deployed |
-| `timestamp` | `int`    | EPOCH timestamp representing event occurrence. Allowed format: `10 digit int`|
+| `timestamp` | `int`    | EPOCH timestamp representing event occurrence. Must be not older then 30min from now. Allowed format: `10 digit int`|
 
 ##### committime
 For the Header X-Pelorus-Event: `committime`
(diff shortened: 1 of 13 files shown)


exporters/webhook/models/pelorus_webhook.py
```diff
@@ -44,17 +45,14 @@ class PelorusDeliveryHeaders(BaseModel):
...
-            except (TypeError, ValueError):
-                raise
+            algorithm, signature = value.split("=", 1)
+            if algorithm != "sha256":
+                raise ValueError("Signature should use sha256 algorithm")
+            if not signature or len(signature) != 64:
+                raise ValueError(
+                    "Signature should be in format 'sha256=' followed by 64 characters"
+                )
+            int(signature, 16)
```
mateusoliveira43, line 55: this line does anything?

mpryc: Yes it ensures the string is proper hex.

docs/GettingStarted/configuration/ExporterWebhook.md
```diff
@@ -90,7 +90,7 @@ For the Header X-Pelorus-Event: `deploytime`
 | `app`       | `string` | Monitored application name |
 | `image_sha` | `string` | Image SHA used for deployment. Must be prefixed with the `sha256:` followed by 64 characters of small letters and numbers |
 | `namespace` | `string` | OpenShift namespace to which application was deployed |
-| `timestamp` | `int`    | EPOCH timestamp representing event occurrence. Allowed format: `10 digit int`|
+| `timestamp` | `int`    | EPOCH timestamp representing event occurrence. Must be not older then 30min from now. Allowed format: `10 digit int`|
```
mateusoliveira43, line 93: "then" here should be "than"?

I think we can rewrite this (I am afraid user can miss the "not" and understand it wrong)

what about `Maximum of 30 minutes old data from now`

PR #943, opened 2023-04-18
Ensure deploytime exporter accepts only valid timestamps

mpryc:
Make deploytime exporter consistent between webhook and deploy type exporter.

## Problem

Current exposed [metric](https://github.com/dora-metrics/pelorus/blob/616ff8f4ebbd807038e93de15e5f8c14da7c4b2e/exporters/webhook/app.py#L124-L126) format of the webhook's deploytime is different from the original deploytime exporter [metric](https://github.com/dora-metrics/pelorus/blob/616ff8f4ebbd807038e93de15e5f8c14da7c4b2e/exporters/deploytime/app.py#L55-L59).

The difference is visible below and shows that the webhook exporter implementation does not add timestamp() to the metric. 

There are few reasons why *not* to include the timestamp within the metric as discussed under https://github.com/dora-metrics/pelorus/discussions/956, however with current queries and Prometheus limitations we are required to use it, so we record deployments accurately.

This is especially visible when using webhook exporter that allows to manipulate the timestamps. On top of that the differences in recording metrics caused additional problems in the Grafana dashboards presenting incorrect data, because original deploytime was considered an event that have happened at particular point of time and webhook caused this to be a series of metrics with different timestamps and particular value. In short webhook converted an event to a time series.

```
# Webhook exporter:
deploy_timestamp{app="mongo-todolist",image_sha="sha256:bf4092ccbfa99a3ec1ea93058fe39b8ddfd8db1c7a18081db397c50a0b8ec77d",namespace="mongo-persistent"} 1.682585478e+09

# The original deploytime exporter:
deploy_timestamp{app="mongo-todolist",image_sha="sha256:bf4092ccbfa99a3ec1ea93058fe39b8ddfd8db1c7a18081db397c50a0b8ec77d",namespace="mongo-persistent"} 1.682585478e+09 1682585478000
```

In this PR the webhook and deploy time exporter behaves exactly the same way making webhook consistent rather than changing the current deploy time and the queries associated with the current metrics.

The change does modify the webhook commit time and failure time code, but the time verification applies only to the deploy time.

The accepted samples for the deploy time exporter and deploy events received by webhook must be within 30min prior to sending that payload. This is due to Prometheus limitation. Please consider discussing https://github.com/dora-metrics/pelorus/discussions/956 as another options which would remove such limitation.

### DIFF
--- a/docs/GettingStarted/configuration/ExporterDeploytime.md
+++ b/docs/GettingStarted/configuration/ExporterDeploytime.md
@@ -2,6 +2,8 @@
 
 The job of the deploy time exporter is to capture the timestamp at which a deployment event happen in a production environment.
 
+Deploytime exporter only collects deployment events that are less than 30 minutes old. Older deployments won't be included unless they have been already collected.
+
 > **NOTE:** In order for proper collection, we require that all objects (Pods, Deployments, replicaSets, etc) associated with a particular application be labelled with the same [APP_LABEL](#app_label), which by default is `app.kubernetes.io/name=<app_name>`, where `<app_name>` is the name of the application being monitored.
 
 ## Example
--- a/docs/GettingStarted/configuration/ExporterWebhook.md
+++ b/docs/GettingStarted/configuration/ExporterWebhook.md
@@ -85,15 +85,19 @@ When sending an HTTP POST request to the webhook's configured URL endpoint, the 
 ##### deploytime
 For the Header X-Pelorus-Event: `deploytime`
 
+> **NOTE:** To align with the nature of Prometheus, it is required that the deploytime event being sent to the webhook exporter should not have occurred more than 30 minutes prior to the time of sending.
+
 | Key         | Type     | Description |
 |-------------|----------|---------------|
 | `app`       | `string` | Monitored application name |
 | `image_sha` | `string` | Image SHA used for deployment. Must be prefixed with the `sha256:` followed by 64 characters of small letters and numbers |
 | `namespace` | `string` | OpenShift namespace to which application was deployed |
-| `timestamp` | `int`    | EPOCH timestamp representing event occurrence. Allowed format: `10 digit int`|
+| `timestamp` | `int`    | EPOCH timestamp must be within the last 30 minutes. Allowed format: `10 digit int`|
 
 ##### committime
 For the Header X-Pelorus-Event: `committime`
+
+> **NOTE:** Committime event must be sent to the webhook exporter prior to it's related deploytime event.
 
 | Key           | Type     | Description |
 | ------------- |----------|-------------|
@@ -123,7 +127,7 @@ You can easily send a POST request using [Curl](https://curl.se) directly from t
 * Our application:
     * is named **`mongo-todolist`** in OpenShift.
     * is deployed to the **`mongo-persistent`** namespace.
-    * deployment happened at (EPOCH): **`1678106205`**
+    * deployment happened 10 minutes ago (EPOCH): **`$ date -d '10 min ago' +%s`**
 
     * the container image used for the deployment has SHA **`af4092ccbfa99a3ec1ea93058fe39b8ddfd8db1c7a18081db397c50a0b8ec77d`**
 
--- a/exporters/deploytime/app.py
+++ b/exporters/deploytime/app.py
@@ -13,6 +13,7 @@ import pelorus
 from deploytime import DeployTimeMetric
 from pelorus.config import load_and_log, no_env_vars
 from pelorus.config.converters import comma_separated
+from pelorus.timeutil import METRIC_TIMESTAMP_THRESHOLD_MINUTES, is_out_of_date
 
 supported_replica_objects = {"ReplicaSet", "ReplicationController"}
 
@@ -37,39 +38,41 @@ class DeployTimeCollector(pelorus.AbstractPelorusExporter):
             "Deployment timestamp",
             labels=["namespace", "app", "image_sha"],
         )
-        deployment_active_metric = GaugeMetricFamily(
-            "deployment_active",
-            "Active deployments in cluster",
-            labels=["namespace", "app", "image_sha"],
-        )
+
+        number_of_dropped = 0
 
         for m in metrics:
-            logging.info(
-                "Collected deploy_timestamp{namespace=%s, app=%s, image=%s} %s (%s)",
-                m.namespace,
-                m.name,
-                m.image_sha,
-                m.deploy_time_timestamp,
-                m.deploy_time,
+            if not is_out_of_date(str(m.deploy_time_timestamp)):
+                logging.info(
+                    "Collected deploy_timestamp{namespace=%s, app=%s, image=%s} %s (%s)",
+                    m.namespace,
+                    m.name,
+                    m.image_sha,
+                    m.deploy_time_timestamp,
+                    m.deploy_time,
+                )
+                deploy_timestamp_metric.add_metric(
+                    [m.namespace, m.name, m.image_sha],
+                    m.deploy_time_timestamp,
+                    timestamp=m.deploy_time_timestamp,
+                )
+            else:
+                number_of_dropped += 1
+                logging.debug(
+                    "Deployment too old to be collected: deploy_timestamp{namespace=%s, app=%s, image=%s} %s (%s)",
+                    m.namespace,
+                    m.name,
+                    m.image_sha,
+                    m.deploy_time_timestamp,
+                    m.deploy_time_timestamp,
+                )
+        if number_of_dropped:
+            logging.debug(
+                "Number of deployments that are older then %smin and won't be collected: %s",
+                METRIC_TIMESTAMP_THRESHOLD_MINUTES,
+                number_of_dropped,
             )
-            deploy_timestamp_metric.add_metric(
-                [m.namespace, m.name, m.image_sha],
-                m.deploy_time_timestamp,
-                timestamp=m.deploy_time_timestamp,
-            )
-            logging.info(
-                "Collected deployment_active{namespace=%s, app=%s, image=%s} %s (%s)",
-                m.namespace,
-                m.name,
-                m.image_sha,
-                m.deploy_time_timestamp,
-                m.deploy_time,
-            )
-            deployment_active_metric.add_metric(
-                [m.namespace, m.name, m.image_sha],
-                m.deploy_time_timestamp,
-            )
-        return (deploy_timestamp_metric, deployment_active_metric)
+        yield deploy_timestamp_metric
 
     def get_and_log_namespaces(self) -> set[str]:
         """
--- a/exporters/pelorus/timeutil.py
+++ b/exporters/pelorus/timeutil.py
@@ -5,9 +5,12 @@ Note: `parse_assuming_utc`, `parse_tz_aware`, and `parse_guessing_timezone_DYNAM
 will _always_ produce timezone-aware objects,
 which are necessary for correctness with `astimezone(tz)`, `timestamp()`, and other methods.
 """
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 _ISO_ZULU_FMT = "%Y-%m-%dT%H:%M:%SZ"
+
+# Time after which metrics for the deploytime exporter will not be accepted
+METRIC_TIMESTAMP_THRESHOLD_MINUTES = 30
 
 
 def is_zone_aware(d: datetime) -> bool:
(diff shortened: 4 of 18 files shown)


exporters/pelorus/timeutil.py
```diff
@@ -101,3 +104,16 @@ def to_iso(dt: datetime) -> str:
...
     return dt.astimezone(timezone.utc).strftime(_ISO_ZULU_FMT)
+
+
+def is_outofdate(timestring: str) -> bool:
+    """
+    Helper function, which allows to filter out metrics which are older then
+    the accepted time. This is to ensure Prometheus will not try to scrape too
+    old metrics.
+    """
+    if to_epoch_from_string(timestring) < datetime.now() - timedelta(
```
mateusoliveira43, line 115: I think this is wrong

should not be
```python
datetime.now() - to_epoch_from_string(timestring) < timedelta(minutes=METRIC_TIMESTAMP_THRESHOLD_MINUTES)
```

and this if already returns a boolean, just do
```python
return condition
```

mateusoliveira43: Add a unit test to this function

exporters/webhook/models/pelorus_webhook.py
```diff
@@ -119,6 +122,17 @@ class DeployTimePelorusPayload(PelorusPayload):
     # rfc1035/rfc1123: An alphanumeric string, with a maximum length of 63 characters
     namespace: str = Field(max_length=63)
 
+    @validator("timestamp")
+    def accepted_timestamp_therashold(cls, v):
+        timestamp_datetime = datetime.fromtimestamp(v)
+        if timestamp_datetime < datetime.now() - timedelta(
```
mateusoliveira43, line 128: use timeutil function here

mateusoliveira43:
I felt it hard to review the way it is (with some changes from #941)

exporters/pelorus/timeutil.py
mpryc: No the:

```
datetime.now() - to_epoch_from_string(timestring)
```

Represents the time difference between now and the time of measurement

If the above is greater then timedelta() then it should be outofdate (return True).

So we could use datetime.now() on the left side, but then the logic would be:


```
datetime.now() - to_epoch_from_string(timestring) > timedelta(minutes=METRIC_TIMESTAMP_THRESHOLD_MINUTES)
```

exporters/pelorus/timeutil.py
```diff
@@ -101,3 +104,18 @@ def to_iso(dt: datetime) -> str:
         )
 
     return dt.astimezone(timezone.utc).strftime(_ISO_ZULU_FMT)
+
+
+def is_outofdate(timestring: str) -> bool:
```
mateusoliveira43, line 109: here is not snake_case, should be `is_out_of_date`

exporters/deploytime/app.py
```diff
@@ -37,39 +38,41 @@ def collect(self) -> Iterable[GaugeMetricFamily]:
             "Deployment timestamp",
             labels=["namespace", "app", "image_sha"],
         )
-        deployment_active_metric = GaugeMetricFamily(
-            "deployment_active",
-            "Active deployments in cluster",
-            labels=["namespace", "app", "image_sha"],
-        )
+
+        no_dropped = 0
```
mateusoliveira43, line 42: number_of_dropped?

exporters/pelorus/timeutil.py
```diff
@@ -101,3 +104,18 @@ def to_iso(dt: datetime) -> str:
...
+    """
+    Helper function, which allows to filter out metrics which are older then
+    the accepted time. This is to ensure Prometheus will not try to scrape too
+    old metrics.
+    """
+    print(datetime.now())
```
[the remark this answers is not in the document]
mpryc: Ah right, forgot about that.

exporters/pelorus/timeutil.py
```diff
@@ -5,10 +5,13 @@
 will _always_ produce timezone-aware objects,
 which are necessary for correctness with `astimezone(tz)`, `timestamp()`, and other methods.
 """
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 _ISO_ZULU_FMT = "%Y-%m-%dT%H:%M:%SZ"
 
+# Time after which metrics for the deploytime exporter will not be accepted
```
mateusoliveira43, line 12: I believe it is not necessary this comment

but if you want to keep it, this will apply for everything, right? Not just deploytime exporter

mpryc:
/hold

I need to test the queries which will require some updates (e.g. use of timestamp() function)

docs/GettingStarted/configuration/ExporterWebhook.md
```diff
@@ -120,6 +120,8 @@ For the Header X-Pelorus-Event: `failure`
 
 You can easily send a POST request using [Curl](https://curl.se) directly from the shell. You can store the payload data in a file or pass it as an argument to [Curl](https://curl.se). Below is an example of sending several requests to cover the lifecycle of an application:
 
+  > **NOTE:** To align with the nature of Prometheus, it is required that the event being sent to the webhook exporter should not have occurred more than 30 minutes prior to the time of sending.
```
mateusoliveira43, line 123: I think this note is important, but maybe it is in the wrong place, reading the docs it feels like it applies to all payloads of webhook, and it is just `deploytime` payload, right?

maybe right below deploytime payload specification would be better

exporters/deploytime/app.py
```diff
@@ -37,39 +38,41 @@ def collect(self) -> Iterable[GaugeMetricFamily]:
...
-        deployment_active_metric = GaugeMetricFamily(
-            "deployment_active",
-            "Active deployments in cluster",
-            labels=["namespace", "app", "image_sha"],
-        )
+
+        number_of_dropped = 0
 
         for m in metrics:
+            if not is_out_of_date(str(m.deploy_time_timestamp)):
```
mateusoliveira43, line 45: adding this to code without adding also to docs doesn't feel good for me

exporters/deploytime/app.py
```diff
@@ -37,39 +38,41 @@ def collect(self) -> Iterable[GaugeMetricFamily]:
...
+                logging.debug(
+                    "Deployment too old to be collected: deploy_timestamp{namespace=%s, app=%s, image=%s} %s (%s)",
+                    m.namespace,
+                    m.name,
+                    m.image_sha,
+                    m.deploy_time_timestamp,
+                    m.deploy_time_timestamp,
+                )
+        if number_of_dropped:
             logging.info(
```
mateusoliveira43, line 70: I do not about this log, can it be too often called?

exporters/tests/data/webhook_pelorus_deploytime.json
```diff
@@ -1,6 +1,5 @@
 {
   "app": "mongo-todolist",
   "image_sha": "sha256:af4092ccbfa99a3ec1ea93058fe39b8ddfd8db1c7a18081db397c50a0b8ec77d",
-  "namespace": "mongo-persistent",
-  "timestamp": 1557933657
+  "namespace": "mongo-persistent"
```
mateusoliveira43, line 4: ok to change for deploytime, but others test data should not be touched

Also this breaks `exporters/webhook/README.md` steps, fix it or remove it from there

exporters/tests/test_datetime.py
```diff
@@ -135,3 +137,39 @@ def test_to_epoch_from_string_bad_value(timestamps):
...
+        (
+            "1681828643",
+            1681828643,
+            False,
+        ),  # current_timestamp - metric_timestamp = 0 min
+        ("1681828645", 1681828643, False),  # metric_timestamp in the future
+    ],
+)
+@mock.patch("pelorus.timeutil.datetime")
+def test_sample_too_old(mock_datetime, metric_timestamp, current_timestamp, expected):
```
mateusoliveira43, line 169: I would refactor here (to improve readability and avoid duplication) to something like
```python
@pytest.mark.parametrize(
    "metric_timestamp, expected",
    [
        ("1681826783", True),  # 31 min before mocked_current_datetime
        ("1681826843", False),  # 30 min before mocked_current_datetime
        ("1681826903", False),  # 29 min before mocked_current_datetime
        ("1681828643", False),  # 0 min before mocked_current_datetime
        ("1681828645", False),  # 2 sec after mocked_current_datetime
    ],
)
@mock.patch("pelorus.timeutil.datetime")
def test_sample_too_old(mock_datetime, metric_timestamp, expected):
    mocked_current_datetime = datetime.fromtimestamp(1681828643)

    mock_datetime.now.return_value = mocked_current_datetime
    mock_datetime.fromtimestamp = datetime.fromtimestamp

    assert is_out_of_date(metric_timestamp) == expected
```

exporters/tests/test_in_memory_metric.py
```diff
@@ -61,7 +64,7 @@ def teardown_method(self):
         [
             (
                 "todolist",
-                "1678269658",
+                str(CURRENT_TIMESTAMP),
```
mateusoliveira43, line 67: No need to change here, its commit metric, not deploy

exporters/tests/test_pelorus_base_plugin.py
```diff
@@ -316,7 +315,9 @@ async def test_proper_receive_metric(json_payload):
         "webhook.plugins.pelorus_handler_base.PelorusWebhookPlugin._receive",
         new_callable=AsyncMock,
     ) as mock_receive:
-        mock_receive.return_value = json.loads(json_payload)
```
mateusoliveira43, line 319: No need to change here, its commit metric, not deploy

(create test specific for deploy?)

exporters/webhook/app.py
```diff
@@ -117,24 +117,30 @@ async def prometheus_metric(received_metric: PelorusMetric):
 
     if received_metric_type == PelorusMetricSpec.COMMIT_TIME:
         in_memory_commit_metrics.add_metric(
-            metric.commit_hash, prometheus_metric, metric.timestamp
+            metric.commit_hash,
+            prometheus_metric,
+            metric.timestamp,
```
mateusoliveira43, line 122: remove trailing comma so only the line of deploytime metric is changed

mateusoliveira43 (changes requested):
fix the README commands

exporters/webhook/app.py
mpryc: done

exporters/tests/test_webhook_models.py
```diff
@@ -32,9 +33,11 @@
 
 # TODO no tests for PelorusDeliveryHeaders
 
+CURRENT_TIMESTAMP = int(time.time())
+
 test_payload = {
     "app": "todolist",
-    "timestamp": "1678269658",
+    "timestamp": CURRENT_TIMESTAMP,
```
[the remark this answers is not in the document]
mpryc: ack

exporters/tests/test_pelorus_base_plugin.py
mpryc: actually it will be valid to only accept timestamps from the time of starting the application. I know that at the moment it's not really valid, but really it does not change the test logic, just modifying the time to be current and not in the past, so I think it's not harmful change here.

exporters/tests/test_in_memory_metric.py
mpryc: Similar comment to the one below:

```
actually it will be valid to only accept timestamps from the time of starting the application. I know that at the moment it's not really valid, but really it does not change the test logic, just modifying the time to be current and not in the past, so I think it's not harmful change here.
```

exporters/tests/test_datetime.py
mpryc: changed.

exporters/tests/data/webhook_pelorus_deploytime.json
mpryc: Ok updated the README and added back timestamp as it's a valid to have it there to present how it should be used.

exporters/deploytime/app.py
mpryc: added doc to cover it.

exporters/deploytime/app.py
mpryc: moved to debug. you are right.

docs/GettingStarted/configuration/ExporterWebhook.md
mpryc: Moved that note to the deploytime, also added one note to commit time that it should be sent prior to sending deploytime.

exporters/webhook/app.py
mateusoliveira43: failures still have trailing comma

docs/GettingStarted/configuration/ExporterDeploytime.md
```diff
@@ -2,6 +2,8 @@
 
 The job of the deploy time exporter is to capture the timestamp at which a deployment event happen in a production environment.
 
+The deploy time exporter only takes into account application deployments that occur after it has been set up. Therefore, any deployment events that occurred more than 30 minutes before the deploy time exporter's installation will not be included in its monitoring.
```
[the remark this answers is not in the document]
mpryc: Reworded.

exporters/webhook/app.py
mpryc: fixed in failure as well.

mateusoliveira43 (approved):

PR #945, opened 2023-04-18
docs: ADR for e2e tests

### LINKED ISSUE #937: ADR for e2e tests

mateusoliveira43:

ADR for end to end (e2e) tests.

Specifications for rework/refactor of e2e tests scripts.

## Linked Issues

resolves #937

## Testing Instructions

Read through ADR and check if every case was covered and you agree with every specification.


docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Decision
+
+e2e tests will be more user friendly.
+
+Prow CI will run e2e tests that cover more scenarios and are more reliable.
+
+### How
+
+- We will rewrite the e2e tests script in Python.
```
mpryc, line 21: I do not like putting implementation specifics as the first bullet point. This is implementation detail to achieve `user friendly` but not achievement on it's own.

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+    - easier to write and read it
```
mpryc, line 22: That is statement which I disagree. In some cases it is easier to write and read in other it's not. We need to be careful adding such statement in ADR. I really don't like locking ourselves in one particular implementation language. Parts of the e2e may be considered better in python, parts in shell and parts in a testing framework which is available as community driven e2e test framework for dashboard that is currently used by Pelorus.

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+    - they will run against Pelorus operator (using test image) and not helm
```
mpryc, line 23: They will run against any Pelorus deployment. The deployment method is not part of this ADR. To make it nice and robust we should not combine Pelorus deployment method with e2e tests.

Deployment may be part of the e2e test scripts, but a separate from executing e2e tests itself. The e2e test cases should be smart enough to enable/disable test cases based on the current Pelorus deployment.

Let's say we want to run tests against webhook, deploytime and committime, where committime is has two backends: github, bitbucket.

1. Deploy Pelorus (we don't care if it's Helm or Operator):
    Helm uses: `values.yaml`
    Operator uses: `charts.pelorus.dora-metrics.io/v1alpha1`

2. Verify if the deployment was success.
3. Gather information about deployment (e.g. we have one deploytime, 2 committime with github and bitbucket).
4. Run pre-condition workload based on the data from point 3., such as sending data to the Pelorus via webhook, creating bugs if necessary, deploying application to be monitored by deploytime and so on.
5. Run tests against workload based on the data from point 3. The test cases available may be much higher, but we could dynamically decide if the test is suitable for the particular deployment. 
6. Gather logs from the deployment and test results.

The exit() should happen only after point where the logs are gathered.

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+
+- We will write better documentation on what e2e tests do (script help and read the docs).
+    - more accessible to new developers of the Pelorus project
+
+- We will log only relevant information on tests run.
+    - Too much log information is not good to identify a simple error
```
mpryc, line 29: Yes, but too little is also sometimes wrong, so we could have two logs. Summary and full log?

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+
+- We will not use `mig-demo-apps` anymore.
```
mpryc, line 31: Could you have more information what would be faster to deploy and how this affects our runs? E.g. is it worth replacing it for the 30sec gain?

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+    - make development faster (only one PR and everything in one repository)
```
mpryc, line 32: This is true.

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+    - do not use yaml files for different tests scenarios
```
mpryc, line 33: I do not understand this point.

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+        - use `image_name` option instead of `source_url` for faster runs (using test image)
+    - use the simplest application possible for tests (faster tests runs)
```
mpryc, line 35: Smallest container footprint app?

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+
+- New commits/deployments/issues will be created for each test scenario.
```
mpryc, line 37: I do not understand this point.

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+    - To address new approach of Pelorus metrics
+
+- We will have more tests scenarios:
+    - check each exporter `/metrics` endpoint
```
mpryc, line 41: This is what we do currently.

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+    - check Grafana dashboards for each application
```
mpryc, line 42: So we will create N applications with different timestamps and so on? How would you do this with the time constrains of the timestamp() ?

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+    - webhook exporter tests
```
mpryc, line 43: I think this should not be a separate test point here. It's as any other exporter, just enable appropriate test scenarios for this exporter `type`

docs/adr/0003-End-to-end-tests.md
```diff
@@ -0,0 +1,57 @@
...
+
+- Faster and more reliable tests (some tests were manual)
+- Faster runs (using a simpler application)
+- Individual runs (`todolist` application for GitHub was always run)
+- Faster development
+- Easier to identify errors
+
+### Cons
+
+- Create new commits/issues to providers in each run, which should be cleaned afterwards (more complexity)
```
mpryc, line 57: Another cons I see here is the effort put to the implementation may be grater then gain at this stage.

mpryc (changes requested):

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: we can have log levels, like in the exporters, and the default not being verbose (debug)

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: Example: in the mig-demo-apps PRs ([like this one](https://github.com/konveyor/mig-demo-apps/pull/101/files)), we write the almost the same thing 3 times in 3 different files. We should use just programming logic to change the behavior instead of files

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: yeah, that as well, going to add it

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: I think this is already done for commit/deploy, but issues are always used the old ones

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: yeah, I think we should continue having so added it here. Do you want to remove this scenario?

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: I think that is not a problem, e2e takes minutes to run, it should not hit the 30 minutes window problem (in Prow it takes sometimes more than 1 hour, but it is the setup of it, the script run takes less)

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: alright

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: what you suggest as a change?

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: changing it to 

- We will improve the development (writing) and understanding (readability) of e2e tests script
  - use frameworks to help us rewrite it (like Python's argparse)
  
would be right for you?

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: agreed

changing here to

- it will run against any Pelorus deployment (like Helm, Operator SDK, etc)

is ok for you? or do you want more info?

docs/adr/0003-End-to-end-tests.md
mpryc: `use frameworks where applicable to improve code (...)`

docs/adr/0003-End-to-end-tests.md
mpryc: yes

docs/adr/0003-End-to-end-tests.md
mpryc: or multiple log streams with different log level. I rather do not produce different format of e2e tests depending on some switch. If test breaks many times we don't have opportunity to check why afterwords, so all info necessary should be associated with every run. It's not user application, but development tool.

docs/adr/0003-End-to-end-tests.md
mpryc: no, but reword:

- We will have more tests scenarios:

- The test scenarios will cover:

Current expression can be read that we don't have /metrics now.

docs/adr/0003-End-to-end-tests.md
mpryc: I think this will be a problem. Some of the exporters are running against bugs or issues created in the past. Some deployments for the commits that happened in the past, so if we will add 30min window, we *will* hit this problem here.

docs/adr/0003-End-to-end-tests.md
mpryc: We really have a lot other things to cover, so rewriting a tool just because it will be more developer friendly isn't going to pay for itself if other priorities aren't addressed first.

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: the failure exporter tests (maybe committime/deploytime as well) use issues that already exist to run tests ([like this one](https://github.com/konveyor/mig-demo-apps/issues/82)). we should create them during the test run

docs/adr/0003-End-to-end-tests.md
mateusoliveira43: alright, forgot the most important point: grafana validation that is manual today

that would be enough for paying itself, right?

mpryc: yes

docs/adr/0003-End-to-end-tests.md
mpryc: If we create issues during the test run we won't be able to finish test quickly. Our queries and numbers are based on time series, so I think we will be missing the long enough period of time to correctly capture numbers and graphs represented in grafana.

Think of the situation, where the failure is opened and then closed right away by e2e test (remember we don't want our e2e tests run for a long time, so we don't want to wait 5-10minutes between those events). Will this give us enough time window to do any tests against mean time to restore? The minimum Grafana time window is 5 minutes and I think open/close events should happen outside of minimum time window.

mateusoliveira43:
Closing without merge to do not let it hanging

PR #948, opened 2023-04-19
Feat: Create test images for PRs

mateusoliveira43:
## Describe the behavior changes introduced in this PR

Create test images for each PR, making testing of operator easier for the developers.

## Linked Issues

resolves #881
depends on #969

## Testing Instructions

Test images available! 🧪🚀 To test operator with them, run
```
operator-sdk run bundle \
quay.io/pelorus/rc-pelorus-operator-bundle:vpr948-abc555 \
--namespace test-pelorus-operator
```
To clean up environment afterwards, run
```
operator-sdk cleanup pelorus-operator --namespace test-pelorus-operator
```

PR #963, opened 2023-05-02
Fix: Pin Python dependencies

### LINKED ISSUE #951: Pin Python dependencies
### Fix

It is not the first time a Python dependency updates and breaks our CI. This leads to investigation and waste of time. We should have all Python dependencies versions pinned to avoid this.

We can use GitHub dependabot to check for updates and run CI on them before accepting the update, [like is done here](https://github.com/mateusoliveira43/python-project-template).

### Requirements

- [ ] Pin all Python dependencies versions
- [ ] OPTIONAL Use Poetry 
  - [ ] check if this does not break any step of project (like creating exporters images) 
- [ ] OPTIONAL Use dependabot to check for updates

mateusoliveira43:

## Linked Issues

resolves #951
depends on #948

## Description

Avoid code breaking with unexpected updates from Python dependencies.

Now, poetry will manage the Python dependencies for us. To add (or update) a dependency, run a poetry command (it will check if there are no conflicts with the current dependencies, before adding it). This will update both `pyproject.toml` and `poetry.lock`, but not requirements files. To update them, run the new make command (we will not update requirements file anymore, this is poetry job). Requirements files now have more things in there:
- the previous dependencies, but now all pinned (no more unexpected updates to break the project CI)
- the dependencies of the dependencies (that were already installed previously, but were under the rug, and we had no control over them, which poetry also monitors)
- the hash, which is a security constrain  

Dependabot will open PRs when there are updates available, and the new workflow will warn if the requirements files where not updated.

## Testing Instructions

Run documentation commands and check if nothing broke with the changes (for example `make dev-env`).

To test the built exporters images with pinned versions, run the command github-action[bot] commented on the PR.


.github/workflows/check-python-dependencies.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  python-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+
```
mateusoliveira43, line 16: install poetry before running command

pyproject.toml
```diff
@@ -1,3 +1,64 @@
+[tool.poetry]
+name = "pelorus"
+version = "2.0.9"
```
mateusoliveira43, line 3: exporters are the same version as charts?

pyproject.toml
```diff
@@ -1,3 +1,64 @@
...
+azure-devops = "6.0.0b4"
+exceptiongroup = "^1.1.1"
+fastapi = "^0.95.1"
+git-url-parse = "^1.2.2"
+jira = "3.4.1"
+jsonpath-ng = "^1.5.3"
+openshift = "0.13.1"
+prometheus-client = "^0.16.0"
+pygithub = "^1.58.1"
+python-gitlab = ">=2.4.0" # change to ^?
```
mateusoliveira43, line 27: I think here is the discussion about how to pin the dependencies

For example, we are using version 3 of `python-gitlab` (check `poetry.lock` or `exporters/requirements.txt` files), but we tell here that we accept `2.4` or above, but we do not test that (our CI and we will install version pined in `poetry.lock` or `exporters/requirements.txt` files). I think is best to us use `^` (indicates that we only allow higher minor or patch versions, no major, that could break the code)

More about pining versions in Poetry https://python-poetry.org/docs/dependency-specification/#version-constraints

.github/workflows/check-python-dependencies.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  check-python-dependencies:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
```
mpryc, line 16: Our exporter images are based on the python 3.9 base ubi8 image, do we really want to have this against 3.11 or should we stick with python version that is shipped in quay ?:

```
ubi8/python-39
```

docs/Development.md
```diff
@@ -73,6 +73,14 @@ Within the Pelorus GitHub fork directory:
 
        To ensure the implementation's pull request is merged **after** the ADR's PR it is recommended to add `Depends on: #ADR_PULL_REQUEST_NUMBER` as the first comment on the implementation PRs. Please follow the [dpulls](https://www.dpulls.com/docs#/dependencies) for more information.
 
+## Updating Python dependencies
+
+To update Python dependencies of the project, after updating `pyproject.toml`, run
+```
+make update-requirements
+```
+It is required to have [Poetry](https://python-poetry.org/) installed.
```
mpryc, line 82: Is the `make dev-env` installing poetry?

mpryc (commented):
Could you please explain how this works in the PR comment?

I see now many different depelndencies, some in poetry.lock, some in pyproject.toml and some generated in requirements.*. This at first glance isn't making things any simpler imo. Also is the `make dev-env` still relevant to make all the required development tools across different operating systems?

.github/workflows/check-python-dependencies.yml
mateusoliveira43: since we have the restriction in pyproject.toml that python needs to be 3.9 or higher, I think the python version we run this job is not a problem

But I checked and we have jobs that run only on 3.9, only in 3.11, in 3.10 and 3.11 and in 3.9, 3.10 and 3.11, we should standardize this

mateusoliveira43:
TODO

add documentation how to add, remove and update dependencies with poetry

add poetry to install tools script

scripts/install_dev_tools.sh
```diff
@@ -295,3 +295,9 @@ if should_cli_be_installed "shellcheck" "${cli_tools_arr[@]}" && \
       tar -xJvf "${SHELLCHECK_PATH}" -C "${DWN_DIR}" "$shellcheck_executable"
       cp "${DWN_DIR}/$shellcheck_executable" "${VENV}/bin/"
 fi
+
+if should_cli_be_installed "poetry" "${cli_tools_arr[@]}" && \
+    ! [ -x "$(command -v "${DEFAULT_VENV}/bin/poetry")" ]; then
+        echo "é isso aqui ó $DEFAULT_VENV"
```
mateusoliveira43, line 301: TODO delete

docs/Development.md
```diff
@@ -73,6 +73,47 @@ Within the Pelorus GitHub fork directory:
...
 
+## Python dependencies
+
+The project uses [Poetry](https://python-poetry.org/) to manage the Python dependencies
+
+### Add a Python dependency
+
+To add a Python dependency, run
+```
+poetry add package_name
```
mpryc, line 84: Include here and the following, so it's clear the update requirements is necessary, otherwise one may miss that step. Remove `### Update requirements files` after such change as it won't be relevant anymore.

```shell
$ poetry add package_name
$ make update-requirements
```

pyproject.toml
```diff
@@ -1,3 +1,64 @@
+[tool.poetry]
+name = "pelorus"
+version = "0.1.0" # we are not using it yet
+description = "Automate the measurement of organizational behavior"
+authors = [
```
mpryc, line 5: Is this generated from `OWNERS` file? If not we have two places where we need to update authors, one for OpenShift PROW and second here. How we ensure those are in-sync?

Can we simply say:

authors = "OWNERS" ?

scripts/install_dev_tools.sh
```diff
@@ -295,3 +295,9 @@ if should_cli_be_installed "shellcheck" "${cli_tools_arr[@]}" && \
...
+        curl -sSL https://install.python-poetry.org | POETRY_HOME=$DEFAULT_VENV python3 -
```
mpryc, line 302: use "" around POETRY_HOME.

mpryc (changes requested):

pyproject.toml
mateusoliveira43: No, it is not automatically

Yes, we can specify a string

Going to put `authors = ["Pelorus community"]`

scripts/install_dev_tools.sh
mateusoliveira43: It did not work with the quotes and in the docs is they also do not use quotes https://python-poetry.org/docs/#installing-with-the-official-installer

PR #968, opened 2023-05-03
fix: Version validation for exporters

### LINKED ISSUE #966: Updating exporters images instead of creating new tag
### OpenShift version

Not related to OpenShift

### Problem description

The last 2 runs of [Pelorus Exporter Build GitHub action job](https://github.com/dora-metrics/pelorus/actions/workflows/build-pelorus.yml) have updated v2.0.9 tag instead of creating new tag.

### Steps to reproduce

Merge PR touching `exporters` folder, but without updating charts version.

### Current behavior

Image is updated.

### Expected behavior

- [ ] A new tag should be created and previous version left untouched

### Code of Conduct

mateusoliveira43:

## Linked Issues

resolves #966

## Description

Exporters version (quay tag) should be updated every time exporters folder is modified.

## Testing Instructions

Test different scenarios, like just updating exporters folder, just updating charts folder, just updating pelorus-operator folder, and checking if the script catches all wrong usages.

### DIFF
--- a/exporters/setup.py
+++ b/exporters/setup.py
@@ -1,8 +1,3 @@
 from setuptools import find_packages, setup
 
-setup(
-    name="pelorus",
-    version="2.0.10-rc.1",
-    packages=find_packages(where="."),
-    python_requires=">=3.9",
-)
+setup(name="pelorus", packages=find_packages(where="."), python_requires=">=3.9")
--- a/exporters/setup.py
+++ b/exporters/setup.py
@@ -1,3 +1,8 @@
 from setuptools import find_packages, setup
 
-setup(name="pelorus", packages=find_packages(where="."), python_requires=">=3.9")
+setup(
+    name="pelorus",
+    version="2.0.10-rc.1",
+    packages=find_packages(where="."),
+    python_requires=">=3.9",
+)
--- a/scripts/chart-test.sh
+++ b/scripts/chart-test.sh
@@ -21,7 +21,7 @@ git remote update upstream --prune &> /dev/null
 
 # Enforce chart version bump when exporters folder are touched
 
-if git status exporters | grep exporters &> /dev/null || git --no-pager diff $REMOTE/master --name-status exporters/ | grep exporters &> /dev/null; then
+if ! git --no-pager diff --quiet $REMOTE/master --name-status exporters/; then
   CURRENT_CHART_VERSION="$(grep '^version: ' charts/pelorus/Chart.yaml | cut -c 10-)"
   CHART_FILE_IN_MASTER="$(curl https://raw.githubusercontent.com/dora-metrics/pelorus/master/charts/pelorus/Chart.yaml 2> /dev/null)"
   CHART_VERSION_IN_MASTER=$(echo "$CHART_FILE_IN_MASTER" | grep '^version: ' | cut -c 10-)
--- a/scripts/chart-test.sh
+++ b/scripts/chart-test.sh
@@ -1,5 +1,28 @@
 #!/usr/bin/env bash
 
+# Enforce version bump on exporters
+
+# TODO DEBUG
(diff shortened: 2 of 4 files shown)


exporters/setup.py
```diff
@@ -1,3 +1,8 @@
 from setuptools import find_packages, setup
 
-setup(name="pelorus", packages=find_packages(where="."), python_requires=">=3.9")
```
mateusoliveira43, line 3: did not test if this can break source to image yet

scripts/chart-test.sh
```diff
@@ -1,5 +1,28 @@
 #!/usr/bin/env bash
 
+# Enforce version bump on exporters
+
+# TODO DEBUG
```
mateusoliveira43, line 5: remove before merging

exporters/setup.py
mpryc: Why this change in the PR for validation check?

scripts/chart-test.sh
```diff
@@ -1,5 +1,28 @@
...
+git rev-parse --abbrev-ref HEAD
+git --no-pager diff master... --name-status exporters/
+# TODO DEBUG
+
+if git status exporters | grep exporters &> /dev/null || git --no-pager diff master... --name-status exporters/ | grep exporters &> /dev/null; then
+  EXPORT_VERSION_IN_MASTER="$(curl https://raw.githubusercontent.com/dora-metrics/pelorus/master/exporters/setup.py &> /dev/null)"
+  CURRENT_EXPORT_VERSION="$(grep '    version="' exporters/setup.py  | cut -c 14- | rev | cut -c 3- | rev)"
+  if $EXPORT_VERSION_IN_MASTER | grep "$CURRENT_EXPORT_VERSION" &> /dev/null; then
+    # TODO breaks when updating from 2.0.10-rc.1 to 2.0.10, for example
```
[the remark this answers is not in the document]
mpryc: I think it can be so much easier.

We do have a script:
```
$ make rc-release (runs $./scripts/create_release_pr.sh -n)
```

Which is doing all the version checking and updating logic, so why duplicate it here. It also checks if the PR is rebased on master, this way we don't need to do the curl against master/exporters/setup.py just check if one of the chart file versions is updated, when the exporters/* files were modified.

```
charts/pelorus/subcharts/exporters/Chart.yaml
```

The rest will happen as there is part which checks if the versions are the same across other Chart.yaml files, operator files etc...

exporters/setup.py
mateusoliveira43: I think adding version to exporters can help us do the check

scripts/chart-test.sh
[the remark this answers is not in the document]
mateusoliveira43: how about this?

```sh
if git status exporters | grep exporters &> /dev/null || git --no-pager diff $REMOTE/master --name-status exporters/ | grep exporters &> /dev/null; then
  CURRENT_CHART_VERSION="$(grep '^version: ' charts/pelorus/Chart.yaml  | cut -c 10-)"
  CHART_FILE_IN_MASTER="$(curl https://raw.githubusercontent.com/dora-metrics/pelorus/master/charts/pelorus/Chart.yaml 2> /dev/null)"
  CHART_VERSION_IN_MASTER=$(echo "$CHART_FILE_IN_MASTER" | grep '^version: ' | cut -c 10-)
  if [ "$CHART_VERSION_IN_MASTER" == "$CURRENT_CHART_VERSION" ]; then
    echo "ERROR: Exporters were modified, Charts need version bumping!"
    exit 1
  fi
fi
```

with this we do not need to add version in exporters

mpryc (approved):

mpryc:
Merging as it does not need to wait for the full e2e tests and other tests were ok.

PR #981, opened 2023-05-10
Rework of deploytime codebase

mpryc:
Rework of deploytime codebase

Move functions to the common openshift library.

Optimize querying for the parent objects such as ReplicaSet and ReplicationController. With this implementation we relay on known api versions and api objects directly from the pod.

Added cache to optimize queries to the OpenShift API for the parent objects. For the running pods we do want to always have the latest state, which means we can't do caching there.

Reworked extracting image SHAs from the running pod instances, which is a base for the skopeo as we also get the full URI of the image.

### DIFF
--- a/exporters/deploytime/app.py
+++ b/exporters/deploytime/app.py
@@ -1,11 +1,9 @@
 import logging
-import re
 import time
-from typing import Iterable, Optional
+from typing import Iterable
 
 from attrs import field, frozen
-from openshift.dynamic import DynamicClient, ResourceInstance
-from openshift.dynamic.exceptions import ResourceNotFoundError
+from openshift.dynamic import DynamicClient
 from prometheus_client import start_http_server
 from prometheus_client.core import REGISTRY, GaugeMetricFamily
 
@@ -14,8 +12,13 @@ from deploytime import DeployTimeMetric
 from pelorus.config import load_and_log, no_env_vars
 from pelorus.config.converters import comma_separated
 from pelorus.timeutil import METRIC_TIMESTAMP_THRESHOLD_MINUTES, is_out_of_date
-
-supported_replica_objects = {"ReplicaSet", "ReplicationController"}
+from provider_common.openshift import (
+    filter_pods_by_replica_uid,
+    get_and_log_namespaces,
+    get_images_from_pod,
+    get_owner_object_from_child,
+    get_running_pods,
+)
 
 
 @frozen
@@ -30,7 +33,6 @@ class DeployTimeCollector(pelorus.AbstractPelorusExporter):
 
     def collect(self) -> Iterable[GaugeMetricFamily]:
         logging.info("collect: start")
-
         metrics = self.generate_metrics()
 
         deploy_timestamp_metric = GaugeMetricFamily(
@@ -43,7 +45,7 @@ class DeployTimeCollector(pelorus.AbstractPelorusExporter):
 
         for m in metrics:
             if not is_out_of_date(str(m.deploy_time_timestamp)):
-                logging.info(
+                logging.debug(
                     "Collected deploy_timestamp{namespace=%s, app=%s, image=%s} %s (%s)",
                     m.namespace,
                     m.name,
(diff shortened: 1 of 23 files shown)


exporters/deploytime/app.py
```diff
@@ -67,168 +69,45 @@ def collect(self) -> Iterable[GaugeMetricFamily]:
                     m.deploy_time_timestamp,
                 )
         if number_of_dropped:
-            logging.debug(
+            logging.info(
```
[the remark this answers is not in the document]
mpryc: No. I actually reduced a lot number of logging.info by moving each found metric to logging.debug from logging.info.

We may still possibly rework logging to include only main start loop across *all* exporters in the logs, but that's for another PR. Currently we have for each http request:

```
05-11-2023 15:20:24 INFO     collect: start
05-11-2023 15:20:24 INFO     No namespaces specified and no PROD_LABEL given, watching all namespaces.
05-11-2023 15:20:25 INFO     Watching namespaces {'openshift-network-diagnostics', 'openshift-node', 'openshift-cloud-network-config-controller', 'openshift-machine-api', 'openshift-kube-storage-version-migrator', 'openshift-config-managed', 'kube-public', 'openshift-authentication', 'openshift-service-ca-operator', 'openshift-console-operator', 'openshift-console-user-settings', 'openshift-ingress-canary', 'openshift-marketplace', 'openshift-ovirt-infra', 'openshift-infra', 'openshift-service-ca', 'openshift-cluster-machine-approver', 'openshift-network-operator', 'openshift-authentication-operator', 'openshift-kube-storage-version-migrator-operator', 'openshift-cluster-storage-operator', 'openshift-dns', 'openshift-image-registry', 'openshift-kube-scheduler', 'openshift-kni-infra', 'openshift-cloud-controller-manager-operator', 'openshift-etcd', 'openshift-kube-apiserver-operator', 'openshift-multus', 'openshift-cluster-samples-operator', 'openshift-insights', 'openshift-dns-operator', 'openshift-controller-manager-operator', 'openshift-oauth-apiserver', 'openshift-openstack-infra', 'openshift-etcd-operator', 'openshift-operators', 'testapp', 'openshift-kube-controller-manager-operator', 'openshift-controller-manager', 'kube-node-lease', 'openshift-apiserver-operator', 'openshift-machine-config-operator', 'openshift-cluster-node-tuning-operator', 'openshift-monitoring', 'openshift-operator-lifecycle-manager', 'openshift-cloud-controller-manager', 'openshift-config', 'openshift-kube-apiserver', 'openshift-ingress', 'openshift-kube-scheduler-operator', 'openshift-apiserver', 'openshift-config-operator', 'openshift', 'openshift-vsphere-infra', 'openshift-cloud-credential-operator', 'openshift-sdn', 'openshift-cluster-csi-drivers', 'openshift-ingress-operator', 'openshift-kube-controller-manager', 'default', 'openshift-console', 'kube-system', 'openshift-cluster-version', 'openshift-host-network', 'openshift-user-workload-monitoring'}
05-11-2023 15:20:25 INFO     generate_metrics: start
05-11-2023 15:20:29 INFO     Number of deployments that are older then 30min and won't be collected: 47

```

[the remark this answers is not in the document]
mpryc: currently we log way more, but I really don't want to make such optimization in this PR as it needs to be across other exporters to be consistent.

Dockerfile_skopeo
mateusoliveira43, line 1: For now this file is not used, right?

exporters/deploytime/app.py
```diff
@@ -67,168 +69,45 @@ def collect(self) -> Iterable[GaugeMetricFamily]:
...
 
-            logging.debug(
-                "Getting Replicas for pod: %s in namespace: %s",
-                pod.metadata.name,
-                pod.metadata.namespace,
-            )
+            # Since a commit will be built into a particular image and there could be multiple
+            # containers (images) per pod, we will push one metric per image/container in the
+            # pod template
+            images = get_images_from_pod(pod)
```
mateusoliveira43, line 100: the comment here I think is not clear, can you rewrite it?

also, this function should return  list to be easier to use it

but, shouldn't it check if each image sha is different? (return a set)

exporters/provider_common/openshift.py
```diff
@@ -1,11 +1,61 @@
...
 
 from pelorus.timeutil import parse_assuming_utc
 
 # https://docs.openshift.com/container-platform/4.10/rest_api/objects/index.html#io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta
 _DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
+SUPPORTED_REPLICA_OBJECTS = ["ReplicaSet", "ReplicationController"]
+
+# Cache threshold for each of the cached_parents_dict entries
+CACHE_THRESHOLD = 60 * 60 * 24
```
mateusoliveira43, line 19: add value and unit to variable name

exporters/provider_common/openshift.py
```diff
@@ -1,11 +1,61 @@
...
+cached_parents_dict: dict[str, ResourceInstance] = {}
```
mateusoliveira43, line 20: do not like having two things, let's merge them together 

we can use tuple, dict or a pydantic object as the dict value

example

```py
cached_parents = {
    '123': (<k8_obj_reference>, 123456)
}
```

exporters/provider_common/openshift.py
```diff
@@ -1,11 +1,61 @@
...
+    """
+    return cached_parents_dict.get(uid)
+
+
+def _remove_expired_objects() -> None:
+    """
+    Cleanup function to remove expired objects from the cache.
+    """
+    current_time = time.time()
+    expired_keys = []
```
mateusoliveira43, line 49: [applies for others places]
Do not like creating a variable for storing the values of a for loop, we should use list comprehension 

example
```py
    expired_keys = [
        uid
        for uid, insertion_time in cached_parents.items()
        if current_time - insertion_time > CACHE_THRESHOLD
    ]
```

exporters/provider_common/openshift.py
```diff
@@ -19,3 +69,234 @@ def convert_datetime(dt: Union[str, datetime]) -> datetime:
...
+                owner_ref.uid,
+            )
+    return {}
+
+
+def filter_pods_by_replica_uid(
+    pods_list: List[ResourceField],
+) -> Dict[str, ResourceField]:
+    """
+    Filters out the given list of Pod objects to create a dictionary with ReplicaSet or
```
mateusoliveira43, line 193: I think the explanation here is too big and not clear

I understand as "return a relation 1-1 replica object uid to pods. Not all pod may appear in this ralation", am I right?

exporters/provider_common/openshift.py
```diff
@@ -19,3 +69,234 @@ def convert_datetime(dt: Union[str, datetime]) -> datetime:
...
+    client: DynamicClient, namespaces: set[str], prod_label: str
+) -> set[str]:
+    """
+    Get the set of namespaces to watch, and log what they are.
+    They will be either:
+    1. The namespaces explicitly specified
+    2. The namespaces matched by PROD_LABEL
+    3. If neither namespaces nor the PROD_LABEL is given, then implicitly matches all namespaces.
+    """
+    if namespaces:
```
mateusoliveira43, line 234: this function was not touched, just moved here, right? But I think we can improve it

here shouldn't we check if the namespaces exist first? 

and in line 252 (log.info) should be a else from the following if

exporters/provider_common/openshift.py
```diff
@@ -19,3 +69,234 @@ def convert_datetime(dt: Union[str, datetime]) -> datetime:
...
+    namespaces = {ns.metadata.name for ns in all_namespaces.get(**query_args).items}
+    logging.info("Watching namespaces %s", namespaces)
+    if not namespaces:
+        logging.warning(
+            "No NAMESPACES given and PROD_LABEL did not return any matching namespaces."
+        )
+    return namespaces
+
+
+def _extract_image_sha(image_url_or_id: str) -> Optional[Tuple[str, str, str]]:
```
mateusoliveira43, line 260: Here I think is over complicated (and the name/description doesn't seen right for me: gets registry and image name as well)

why don't return string or None instead of Tuple?

exporters/deploytime/app.py
mpryc: Return function gives us a `Dict[str, str]` where the first item is a key that corresponds just to the image `sha` and the value is a full URI pointing to a docker image. The `keys()` from the returned dictionary gives us a unique image sha, so there is no need to use set(). Secondly list would be simpler, however we are interested in the full docker URI as well for later usage by other commands such as `skopeo`. I know this is not relevant at this stage, but do want to have function generic enough to give us the flexibility:

```
def get_images_from_pod(pod: ResourceField) -> Dict[str, str]:
    image_shas = {}

......

    for container_status in pod.status.containerStatuses:
        image_shas[
            sha256_value
        ] = f"docker://{registry}{image_name}{sha256_value}"

   return image_shas
```
For example `image_sha : URI`:
```
{
'sha256:6b88f190c7a6cd87c1f1b552c20855e606773cfa96d43f4402840a4b9d2e13a8' :
'docker://quay.io/pelorus/pelorus-committime-exporter@sha256:6b88f190c7a6cd87c1f1b552c20855e606773cfa96d43f4402840a4b9d2e13a8'
}
```

Allows us to have sha256 used in the metric, but as well will allow us to run based on the value URI:

```
skopeo inspect docker://quay.io/pelorus/pelorus-committime-exporter@sha256:6b88f190c7a6cd87c1f1b552c20855e606773cfa96d43f4402840a4b9d2e13a8
```

exporters/provider_common/openshift.py
mpryc: The relation here is a bit more complex. In some cases it's many-many.

One pod may have multiple replica objects similarly one replica object may contain multiple pods.

The relation we are taking here is that we always have homogeneous pods within one replica.

In the scenario where we have multiple replicas and the same 2 pods controlled by those replicas, so:

```
replicaA --- podX
replicaA --- podY
replicaB --- podX
replicaB --- podY
replicaC --- podX
replicaC --- podY
```

We will get actually 3 deployments as our replica UID makes a `deployment` and here we have 3 different ReplicaSets controlling the same pods.

In the other scenario where we have two replicas controlling different pods:
```
replicaA --- podA
replicaA --- podB
replicaA --- podC
replicaB --- podX
replicaB --- podY
replicaB --- podZ
```

We will get 2 deployments as we have 2 replica unique UIDs.

exporters/provider_common/openshift.py
mpryc: No, it's perfectly valid scenario to setup exporter first and then create that namespace later.

exporters/provider_common/openshift.py
mpryc: It's just to make the function more generic. We could have only (sha256, full URI), but really I want to make sure we are also able to get just the image name. In situations where we use internal registry then using `{internal-registry}image_name@sha_value` may seem like a valid scenario and will improve dramatically calls skipping external registry queries.

Dockerfile_skopeo
mpryc: Nope, added by mistake now.

exporters/provider_common/openshift.py
mpryc: fixed.

exporters/provider_common/openshift.py
mpryc: merged.

exporters/provider_common/openshift.py
mpryc: fixed. what other places do you want to use list comprehension? I am fine using them if within the list comprehension there is no call to the external method as it's harder to debug later.

mpryc (commented):
replied to comments and submitted new revision.

exporters/provider_common/openshift.py
mateusoliveira43: the typing is inverted, its a tuple of first item being openshift object and the second being a float representing a timestamp, right?

mpryc: good catch !

exporters/provider_common/openshift.py
```diff
@@ -1,11 +1,60 @@
...
+    """
+    if uid not in cached_parents_dict:
+        cached_parents_dict[uid] = (k8s_obj, time.time())
+
+
+def _get_object_from_cache(uid: str) -> ResourceInstance:
+    """
+    Gets the object from the cache by it's uid.
+    """
+    k8s_obj, insertion_time = cached_parents_dict.get(uid) or (None, None)
```
mateusoliveira43, line 39: here there is an unused variable (and in the remove function). We can use `_` in these cases

example
```py
k8s_obj, _ = cached_parents_dict.get(uid, (None, None))
return k8s_obj
```

exporters/provider_common/openshift.py
mateusoliveira43: Adding an example to the description seems like a good solution

But I would also want tests for it

for example, for the first scenario, this are both valid returns from the function, right?
```py
# option 1
{'uidA': <podX>, 'uidB': <podX>, 'uidC': <podX>}
# option 2
{'uidA': <podY>, 'uidB': <podY>, 'uidC': <podY>}
# there are more options still, right?
``` 

and for the second scenario
```py
# option 1
{'uidA': <podA>, 'uidB': <podX>}
# option 2
{'uidA': <podB>, 'uidB': <podY>}
# there are more options still, right?
``` 

since the return of the function can be not unique, this can lead us to not collect all data?

mateusoliveira43 (changes requested):

exporters/provider_common/openshift.py
mpryc: Done.

Renamed to `_parse_container_image_uri`, changed return value to be Union of values or Tuple of None's with the following description:

```
def _parse_container_image_uri(image_uri: str) -> Union[Tuple[str, str, str], Tuple[None, None, None]]:
    """
    Parses the container image URI and extracts image registry, image name and image SHA256 value.

    Args:
        image_uri[str]: Container image URI. 
            Expected is an URI with registry URI and SHA256 value.

    Returns:
        Tuple[str, str, str]
            Parsed container URI into Tuple of registry URI, image name and SHA256 value.
            If any of the expected values is not found it then returns (None, None, None).
    """
```

mateusoliveira43 (approved):

PR #992, opened 2023-05-16
Skopeo exporter

### LINKED ISSUE #967: Relying on `image.openshif.io/Image` for the `provider: image` in CommitTime doesn't make sense
### OpenShift version

Not related to OpenShift

### Problem description

It appears that the way that the Commit Time exporter discovers container images to scrape for `Label` info doesn't make sense. The exporter does this:

```
        v1_images = self.kube_client.resources.get(
            api_version="image.openshift.io/v1", kind="Image"
        )

        images = v1_images.get(label_selector=app_label)
```
https://github.com/dora-metrics/pelorus/blob/master/exporters/committime/collector_image.py#L161-L165

Which is the equivalent of:

```
oc get Images -l app.kubernetes.io/name
```

The problem with this approach is that `Image` type in OpenShift is not a resource that Developers ever interact with, or would ever have the ability to label. It is a resource that gets generated by the OpenShift internal image registry based on `docker/podman push` events. There is no way for a developer to build or push an image such that the `APP_LABEL` would get set on the respective `Image` resource.

This basically makes this feature unusable in the real world, despite it making it very easy to test & mock. 

## Alternatives

1. The exporter could just get ALL images, and not filter by kuberenetes labels, and then use python filters to look for app label in `"dockerImageMetadata.Config.Labels"`. This however would still couple the implementation to OpenShift clusters only.
2. The better way would be to have the image provider function more like the DeployTime exporter, and search for Pods with `APP_LABEL` on them, then grab it's `.spec.containers[*].image` reference, and then pull labels directly from the referenced image via the registry API. Grabbing the labels can be done either via API calls in python, or something like skopeo:
   ```
   skopeo inspect docker://registry.acme.com/myapp:stable | jq .Labels
   ```

### Steps to reproduce

1. ...
2. ...
3. ...

### Current behavior

See above

### Expected behavior

See above

### Code of Conduct

mpryc:

## Linked Issues

resolves #967

## Description

A new type of provider for the commit time exporter that works similarly to the deploytime, but additionally for the found running applications it reaches out to the external container registry and tries to get commit time from the Docker label.

## Testing Instructions

Currently run the exporter using python interpreter on the system that has skopeo installed:

```
$ export KUBECONFIG=<path to kubeconfig>
$ export LOG_LEVEL=debug
$ export PROVIDER=containerimage
$ export NAMESPACES=<optional step to list namespaces for running pods>
$ python exporters/committime/app.py
$ #Open http://localhost:8080
```

This exporter can be use to calculate the above commit time, based on the provided example application, that is already in the test quay.io repository:

Create ReplicaSet using OCP WebUI or `oc apply` command with the following ReplicaSet in the `testnamespace` namespace, that must be existing:

```yaml
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: example
  namespace: testnamespace
spec:
  replicas: 2
  selector:
    matchLabels:
      app: httpd
  template:
    metadata:
      name: httpd
      labels:
        app: httpd
        app.kubernetes.io/name: httpd
    spec:
      containers:
        - name: httpd
          image: >-
            quay.io/pelorus/httpd-sample-app:latest
          ports:
            - containerPort: 8080
```

And start the exporter as above.

You can also create instance of the committime as in the following example (note committime one):

```yaml
apiVersion: charts.pelorus.dora-metrics.io/v1alpha1
kind: Pelorus
metadata:
  name: pelorus-sample
  namespace: test-pelorus-operator
spec:
  exporters:
    instances:
      - enabled: true
        app_name: deploytime-exporter
        exporter_type: deploytime
      - enabled: true
        extraEnv:
          - name: PROVIDER
            value: containerimage
        app_name: committime-exporter
        exporter_type: committime
        image_name: 'quay.io/pelorus/rc-pelorus-exporter:skopeo_8'
    global: {}
  openshift_prometheus_basic_auth_pass: changeme
  openshift_prometheus_htpasswd_auth: 'internal:{SHA}+pvrmeQCmtWmYVOZ57uuITVghrM='
  prometheus_retention: 1y
  prometheus_retention_size: 1GB
  prometheus_storage: false
  prometheus_storage_pvc_capacity: 2Gi
  prometheus_storage_pvc_storageclass: gp2
```

### DIFF
--- a/docs/GettingStarted/configuration/ExporterCommittime.md
+++ b/docs/GettingStarted/configuration/ExporterCommittime.md
@@ -2,11 +2,13 @@
 
 The job of the commit time exporter is to find and associate time of the relevant source code commit with a container image SHA built from that source code.
 
-This information can be found in two ways:
+This information can be found in three ways:
 
 - [Git API](#using-commit-time-with-git-apis) making a request to the Git provider
 
-- [Containers' Image](#using-commit-time-with-images) querying `Annotations` or `Labels` within the Image object, without reaching to the git external services. For the image Time provider, the Commit Date must be present as a valid string. This allows to give greater flexibility around 3rd party CI systems used to create container images for the OpenShift running applications.
+- [OpenShift Image Objects](#using-commit-time-with-openshift-image-objects) querying `Annotations` or `Labels` within the OpenShift Image object, without reaching to the git external services. For the `image` Commit Time provider type, the Commit Date must be present as a valid string. This allows to give greater flexibility around 3rd party CI systems used to create container images for the OpenShift running applications.
+
+- [Containers' Image Labels](#using-commit-time-with-containers-image-labels) querying container registry for the OCI `Labels`. For the `containerimage` Commit Time provider type, the Commit Date and Commit Hash must be present as a valid string. This allows to include the Commit Time information at a container build time.
 
 
 Later the Deploy Time Exporter can associate that image SHA with a production deployment and allow to calculate [Lead Time for Change](../../philosophy/outcomes/SoftwareDeliveryPerformance.md#lead-time-for-change) metrics.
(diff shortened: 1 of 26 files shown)


exporters/committime/collector_base.py
```diff
@@ -67,6 +67,8 @@ class AbstractCommitCollector(pelorus.AbstractPelorusExporter):
 
     namespaces: set[str] = field(factory=set, converter=comma_separated(set))
 
+    prod_label: str = field(default=pelorus.DEFAULT_PROD_LABEL)
```
mateusoliveira43, line 70: this adds `PROD_LABEL` to git and image(stream) committime exporter, don't? I think it is wrong, should be a config only for new exporter

mpryc: I've added this to collector_base as I think it's important to rework the commit time exporter(s) to use:

```
get_and_log_namespaces()
```

From:
```
provider_common.openshift
```

But I did not want to make such change in this PR as it's not relevant to the skopeo work, however though the PROD_LABEL natural  place is on the base abstract class.

mateusoliveira43:
Don't forget documentation for the new exporter

exporters/committime/collector_base.py
mpryc: This one should be replaced with get_and_log_namespaces() which does support PROD_LABEL to be consistent with deploytime exporter as well:

https://github.com/dora-metrics/pelorus/blob/master/exporters/committime/collector_base.py#L135-L146

mateusoliveira43: I am fine with this only if we add NAMESPACES and PROD_LABEL options to all committime exporters

docs/GettingStarted/configuration/ExporterCommittime.md
```diff
@@ -367,4 +386,52 @@ $ oc label image "sha256:${IMAGE_SHA}" "app.kubernetes.io/name=${NAME}"
...
+
+LABEL io.openshift.build.commit.date=${LAST_COMMIT_DATE_TIME}
+LABEL io.openshift.build.commit.id=${LAST_COMMIT_SHA}
+LABEL io.openshift.build.source-location="https://github.com/sclorg/httpd-ex"
+```
+
+To build above example, use any container build method, in this example we will use `podman`.
+
+```shell
+git clone https://github.com/dora-metrics/pelorus
```
mateusoliveira43, line 414: I think is too much to clone repo because of one file, I would simply tell user to create file with the following content and run the following command

exporters/tests/httpd_docker_with_labels/Makefile
```diff
@@ -0,0 +1,22 @@
+VERSION ?= latest
```
mateusoliveira43, line 1: this file was cited anywhere?

mateusoliveira43:
please update PR description/test instructions, it is outdated

exporters/committime/collector_containerimage.py
```diff
@@ -0,0 +1,312 @@
...
+        logging.info("generate_metrics: start")
+
+        _clear_cleanup_set()
+
+        pods = get_running_pods(self.kube_client, namespaces, self.app_label)
+
+        # Build dictionary with controllers and retrieved pods
+        replica_pods_dict = filter_pods_by_replica_uid(pods)
+
+        for _, pod in replica_pods_dict.items():
```
mateusoliveira43, line 285: `.values` instead of `.items`?

exporters/committime/collector_containerimage.py
```diff
@@ -0,0 +1,312 @@
+#!/usr/bin/env python3
```
mateusoliveira43, line 1: I do not like the idea of adding new exporter without any new tests

At least some usage tests like these https://github.com/dora-metrics/pelorus/blob/master/exporters/tests/test_committime_exporter_app.py#LL172C1-L227C84

docs/GettingStarted/configuration/ExporterCommittime.md
mpryc: It is not possible, because in this example we take the last commit hash and last commit time from the git repo to create appropriate LABEL. This is to show how that workflow can be used in real life from any git repository that is building the container image.

exporters/tests/httpd_docker_with_labels/Makefile
mpryc: Yes in the updated example in `ExporterCommittitme.md`:

```
git clone https://github.com/dora-metrics/pelorus
pushd pelorus/exporters/tests/httpd_docker_with_labels/
export CONTAINERFILE_LOCATION=./Dockerfile
```

exporters/committime/collector_containerimage.py
mpryc: Added unit tests, not covering 100% but the rest of the functions are common openshift tested elsewhere.

exporters/committime/collector_containerimage.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+def _cache_container_images_labels(sha_256: str, labels: Dict) -> None:
+    with image_label_cache_lock:
+        if sha_256 not in image_label_cache:
+            logging.debug(f"Adding SHA256 to the cache: {sha_256} ")
+            image_label_cache[sha_256] = (labels, time.time())
+
+
+def _remove_sha256_from_cache(sha_256: str) -> None:
```
mateusoliveira43, line 94: This is not used

mpryc: right, added it to have a nice interfaces, but will remove as it's not really used.

exporters/tests/httpd_docker_with_labels/Makefile
mpryc: deleted

exporters/committime/collector_containerimage.py
```diff
@@ -0,0 +1,306 @@
...
+        metrics = []
+
+        namespaces = get_and_log_namespaces(
+            self.kube_client, self.namespaces, self.prod_label
+        )
+
+        if not namespaces:
+            return metrics
+
+        logging.info("generate_metrics: start")
```
mateusoliveira43, line 270: debug

mpryc: done

mateusoliveira43:
I have run the exporter locally without installing skopeo, and it does not break (I think it triggers skopeo_failures). I would create an issue about this, I think it strange to not break when skopeo is not available

mateusoliveira43 (approved):

mpryc:
> I have run the exporter locally without installing skopeo, and it does not break (I think it triggers skopeo_failures). I would create an issue about this, I think it strange to not break when skopeo is not available

If skopeo fails we create a failed list and we re-try up to 3 times and then pass for 2 days. We really don't care what was the failure as skopeo may have number of errors, but you are right that missing skopeo CLI may be something we should catch in a first place. We do control the container image content and with your change we now include skopeo, so it's not really something standard user will see imo.

PR #998, opened 2023-05-19
feat: Add fallback option when Jira exporter can not determine application

### LINKED ISSUE #994: Define static app_name for Failure exporter
### Feature description

Ability to add a static definition of the app_name for the failure exporter.

The main problem is that the Jira issues do not always have a united label that we could unite them under and some Jira projects contain multiple services. Changing the Jira template is non-feasible.

We can overcome this by defining a new instance of the exporter for each service with a custom Jira query which would get only tickets that are relatable to some service and statically defining the app_name under which the tickets should be accumulated. 

Possible config snippet.
```
- app_name: psca-failure-exporter-test
        enabled: true
        exporter_type: failure
        extraEnv:
          - name: TOKEN
            value: <token>
          - name: PROVIDER
            value: jira
          - name: SERVER
            value: https://issues.redhat.com/
          - name: JIRA_JQL_SEARCH_QUERY
            value: project = PVCORE AND issuetype = Bug and EXD-Service in (product-service-config, product-service-config-auto)
          - name: LOG_LEVEL
            value: debug
          - name: APP_NAME
            value: PSCA
```

### How this will help you

It would add versatility and allow us to define Failure exporter for each service respectively.
That would unblock us from progressing the PoC and would solve a lot of Edge cases with Jira projects.

### Feature priority

🚨 As soon as possible, it block me of using Pelorus

### Code of Conduct

mateusoliveira43:

## Linked Issues

resolves #994

## Description

Add fallback option when Pelorus can not determine `app_name` from issue.

## Testing Instructions

Create Pelorus failure exporter using
```yaml
apiVersion: charts.pelorus.dora-metrics.io/v1alpha1
kind: Pelorus
metadata:
  name: test-app-name-option
spec:
  exporters:
    instances:
      - app_name: failure-exporter
        exporter_type: failure
        extraEnv:
          - name: API_USER
            value: jira_user
          - name: TOKEN
            value: jira_token
          - name: SERVER
            value: https://pelorustest.atlassian.net
          - name: APP_NAME
            value: robotron
          - name: PROJECTS
            value: 'Test,pelorus_tekton_test,todolist-mongo'
          - name: JIRA_RESOLVED_STATUS
            value: DONE
```
and check that no issues are related to app "unknown" and app "robotron" appears in Grafana, using Operator commented by CI.

### DIFF
--- a/docs/GettingStarted/configuration/ExporterFailure.md
+++ b/docs/GettingStarted/configuration/ExporterFailure.md
@@ -56,6 +56,7 @@ This is the list of options that can be applied to `env_from_secrets`, `env_from
 | [API_USER](#api_user) | no | - |
 | [TOKEN](#token) | yes | - |
 | [APP_LABEL](#app_label) | no | `app.kubernetes.io/name` |
+| [APP_NAME](#app_name) | no | - |
 | [APP_FIELD](#app_field) | no | `u_application` |
 | [PROJECTS](#projects) | no | - |
 | [PELORUS_DEFAULT_KEYWORD](#pelorus_default_keyword) | no | `default` |
@@ -116,6 +117,13 @@ This is the list of options that can be applied to `env_from_secrets`, `env_from
 - **Type:** string
 
 : Changes the label used to identify applications.
+###### APP_NAME
+
+- **Required:** no
+    - Only applicable for [PROVIDER](#provider) set to `jira`
+- **Type:** string
+
+: Set `unknown` application to desired value.
 
 ###### APP_FIELD
 
--- a/docs/GettingStarted/configuration/ExporterFailure.md
+++ b/docs/GettingStarted/configuration/ExporterFailure.md
@@ -117,13 +117,14 @@ This is the list of options that can be applied to `env_from_secrets`, `env_from
 - **Type:** string
 
 : Changes the label used to identify applications.
+
 ###### APP_NAME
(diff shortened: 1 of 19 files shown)


docs/GettingStarted/configuration/ExporterFailure.md
```diff
@@ -116,6 +117,13 @@ This is the list of options that can be applied to `env_from_secrets`, `env_from
 - **Type:** string
 
 : Changes the label used to identify applications.
+###### APP_NAME
```
mateusoliveira43, line 120: This is enough for the documentation update?

exporters/failure/collector_jira.py
```diff
@@ -279,4 +281,7 @@ def get_app_name(self, issue: Issue) -> str:
             matcher = f"{self.app_label}="
             if label.startswith(matcher):
                 return label.replace(matcher, "")
-        return "unknown"
+        if self.app_name is None:
+            # TODO set this behavior to all failure exporters?
```
mateusoliveira43, line 285: Should we add this to all failure exporters?

mpryc: I don't think so, maybe when there is a request. Also the sdp queries may require some modification as some of them do have "unknown" filter.

mpryc (approved):

PR #1011, opened 2023-06-01
Add Timestamp to the Failure exporter

### LINKED ISSUE #1010: Improper Change Failure Rate metric
### OpenShift version

Not related to OpenShift

### Problem description

Broken UI for the Change Failure Rate metric when sending some failure create/resolve events associated with the deployments.

### Steps to reproduce

1. Create webhook exporter

2. Sent some data using the following script:
```python
import subprocess
import requests
import json
from datetime import datetime
import time
import hashlib
import secrets
import random

PELORUS_NAMESPACE = "test-pelorus-operator"

TIME_IN_S_BETWEEN_EVENTS = 60

def get_webhook_route(exporter_name="webhook-exporter", namespace=None):
    command = ["oc", "get", "route", exporter_name]
    if namespace:
        command += ["-n", namespace]
    command += ["--template='{{ .spec.host }}'"]

    webhook_route = subprocess.check_output(command).decode("utf-8").strip("\n\r")
    return webhook_route


def generate_sha256_hash_signature(payload_json, secret_value):
    command = [
        "echo", "-n", payload_json, "|", "openssl", "dgst", "-sha256", "-hmac", secret_value, "|", "cut", "-d", " ", "-f", "2"
    ]
    sha256_hash_signature = subprocess.check_output(command).decode("utf-8").strip("\n\r")
    return sha256_hash_signature


def send_request(webhook_route, payload_data, header_pelorus_event, secret_value=""):
    header_user_agent = "Pelorus-Webhook/e2e"
    header_content = "application/json"
    headers = {
        "User-Agent": header_user_agent,
        "X-Pelorus-Event": header_pelorus_event,
        "Content-Type": header_content
    }

    if secret_value:
        payload_json = json.dumps(payload_data)
        sha256_hash_signature = generate_sha256_hash_signature(payload_json, secret_value)
        headers["X-Hub-Signature-256"] = f"X-Hub-Signature-256: sha256={sha256_hash_signature}"

    payload_json = json.dumps(payload_data)
    webhook_route_url = "http://" + webhook_route.strip("'") + "/pelorus/webhook"

    print(f"Webhook: {webhook_route_url}, Headers: {headers}, Payload: {payload_json}")
    response = requests.post(
        webhook_route_url,
        headers=headers,
        data=payload_json,
    )
    print(response.text)
    return response

def send_payload(webhook_route, payload_data=None, header_pelorus_event=None, secret_value="", namespace=None):
    if payload_data is None or header_pelorus_event is None:
        return
    response = send_request(webhook_route, payload_data, header_pelorus_event, secret_value)
    return response

def get_commit_hash_value(length: int):
    num_bytes = length // 2
    random_string = secrets.token_hex(num_bytes)
    sha256_hash = hashlib.sha256()
    sha256_hash.update(random_string.encode('utf-8'))
    return sha256_hash.hexdigest()[:length]

def send_events(x: int, y: int, z: int):
    """
    x, y and z represents every N-th time to sent
         commit, deployment and failure payload
    """

    webhook_route=get_webhook_route()

    counter = 1
    while True:
        image_hash=get_commit_hash_value(64)
        commit_hash=get_commit_hash_value(40)
        failure_no = (failure_no + 1) if 'failure_no' in locals() else 1
        
        payload_data_commit = {
            "app": "mongo-todolist",
            "commit_hash": commit_hash,
            "image_sha": f"sha256:{image_hash}",
            "namespace": "mongo-persistent",
            "timestamp": int(datetime.now().timestamp()),
        }

        if counter % x == 0:
            send_payload(webhook_route, payload_data_commit, header_pelorus_event="committime", namespace="test-pelorus-operator")
        time.sleep(TIME_IN_S_BETWEEN_EVENTS)

        payload_data_deploy = {
            "app": "mongo-todolist",
            "image_sha": f"sha256:{image_hash}",
            "namespace": "mongo-persistent",
            "timestamp": int(datetime.now().timestamp()),
        }
        
        if counter % y == 0:
            send_payload(webhook_route, payload_data_deploy, header_pelorus_event="deploytime", namespace="test-pelorus-operator")
        time.sleep(TIME_IN_S_BETWEEN_EVENTS)

        payload_failure_create = {
          "app": "mongo-todolist",
          "failure_id": f"MONGO-{failure_no}",
          "failure_event": "created",
          "timestamp": int(datetime.now().timestamp()),
        }

        if counter % z == 0:
            send_payload(webhook_route, payload_failure_create, header_pelorus_event="failure", namespace="test-pelorus-operator")
        time.sleep(TIME_IN_S_BETWEEN_EVENTS)

        payload_failure_resolve = {
          "app": "mongo-todolist",
          "failure_id": f"MONGO-{failure_no}",
          "failure_event": "resolved",
          "timestamp": int(datetime.now().timestamp()),
        }

#       if counter % z == 0:
#            send_payload(webhook_route, payload_failure_resolve, header_pelorus_event="failure", namespace="test-pelorus-operator")
#            time.sleep(TIME_IN_S_BETWEEN_EVENTS)
        counter += 1

send_events(1, 1, 1)
```

3. Wait around 1h, and run the script again, however this time modify the above script and this time uncomment commented lines, but comment the lines:
```python
#        if counter % z == 0:
#            send_payload(webhook_route, payload_failure_create, header_pelorus_event="failure", namespace="test-pelorus-operator")
#        time.sleep(TIME_IN_S_BETWEEN_EVENTS)
```

### Current behavior

4. In grafana select time window to be maximum 1h. You will see data is wrong, presenting some wired numbers. The CFR should not exceed 100%, see attached screenshot.

Screenshot from 2023-05-29 13-12-35

### Expected behavior

Proper CFR metric that shows around 100%. By modifying the script and re-sending the events we can check that CFR should be around 33% with modified line in the example send script (every 3 deployments one failure with resolution:

```python
send_events(1, 1, 3)

```

### Code of Conduct

mpryc:
Adding Timestamp to the Failure exporter is important to properly calculate Change Failure Rate Over Time (CFR)

The CFR metric is a simple number that represents the number of failed deployments to the number of all deployments. As such without the timestamp we were adding the Failures to the previously known ones and never calculating properly based on a chosen time window.

The max_over_time is required to be able to deduct creation time of the failure from the resolution time of that failure to properly calculate Mean Time to Restore and other metrics. Without that we would almost never had the time window that both metrics create/failure would be within it and as such the deduction would return nothing.

- [ ] I have read [Pelorus Contributing guidelines](https://github.com/dora-metrics/pelorus/blob/master/CONTRIBUTING.md)

## Linked Issues

resolves #1010

## Testing Instructions

Use script provided in the #1010 to test the scenario.

### DIFF
--- a/charts/pelorus/templates/prometheus-rules.yaml
+++ b/charts/pelorus/templates/prometheus-rules.yaml
@@ -20,7 +20,7 @@ spec:
         avg(sdp:lead_time:by_app)
     - record: sdp:time_to_restore:by_issue
       expr: >
-        failure_resolution_timestamp{app!="unknown"} - failure_creation_timestamp{app!="unknown"}
+        failure_resolution_timestamp{app!="unknown"} - max_over_time(failure_creation_timestamp{app!="unknown"}[5y])
     - record: sdp:time_to_restore:by_app
       expr: >
         avg by (app) (sdp:time_to_restore:by_issue)
--- a/docs/GettingStarted/configuration/ExporterFailure.md
+++ b/docs/GettingStarted/configuration/ExporterFailure.md
@@ -1,6 +1,10 @@
 # Failure Time Exporter
 
-Failure time exporter captures the timestamp at which a failure occurred, in a production environment, and when it was resolved. It does this by parsing the information of when the issue was created, and closed, in the Issue Tracker(s).
+Failure time exporter captures the timestamps at which a broken deployment occurred, and when it was fixed. It does this by parsing the information of when the issue related to that broken deployment was created, and closed, in the Issue Tracker(s).
+
+The Failure Time Exporter configuration should be adjusted based on the backend being used to restrict failures to only those related to failed deployments in a production.
+
+Failure exporter only collects failure events that are less than 30 minutes old. Older failures won't be included unless they have been already collected.
 
 Failure Time Exporter may be deployed with one of the [supported Issues Trackers](../Overview.md#issue-trackers). In one clusters' namespace there may be multiple instances of Failure Time Exporter, one for each provider (or each project). Each provider requires specific [configuration](#failureconfigmap).
 
--- a/exporters/failure/collector_base.py
+++ b/exporters/failure/collector_base.py
@@ -8,6 +8,9 @@ from prometheus_client.core import GaugeMetricFamily
 
 import pelorus
 
+# TODO 1: CI needs to create failures on the fly to enable this
+# from pelorus.timeutil import METRIC_TIMESTAMP_THRESHOLD_MINUTES, is_out_of_date
+
 
 class AbstractFailureCollector(pelorus.AbstractPelorusExporter):
     """
@@ -31,19 +34,41 @@ class AbstractFailureCollector(pelorus.AbstractPelorusExporter):
 
         if critical_issues:
             metrics = self.generate_metrics(critical_issues)
+            # TODO 1:
+            # number_of_dropped = 0
             for m in metrics:
+                # TBD_1:
+                # if not is_out_of_date(str(m.deploy_time_timestamp)):
                 if not m.is_resolution:
                     logging.info(
                         "Collected failure_creation_timestamp{ app=%s, issue_number=%s } %s"
                         % (m.labels[0], m.labels[1], m.time_stamp)
                     )
-                    creation_metric.add_metric(m.labels, m.get_value())
+                    creation_metric.add_metric(
+                        m.labels, m.get_value(), timestamp=m.get_value()
+                    )
                 else:
                     logging.info(
                         "Collected failure_resolution_timestamp{ app=%s, issue_number=%s } %s"
                         % (m.labels[0], m.labels[1], m.time_stamp)
                     )
-                    failure_metric.add_metric(m.labels, m.get_value())
+                    failure_metric.add_metric(
+                        m.labels, m.get_value(), timestamp=m.get_value()
+                    )
+            # TODO 1:
+            #     else:
+            #         number_of_dropped += 1
+            #         debug_msg = f"Failure too old to be collected: failure_{'resolution' " \
+            #                      "if m.is_resolution else 'creation'}_timestamp{{ app={m.labels[0]}, " \
+            #                      "issue_number={m.labels[1]} }} {m.time_stamp}"
+            #         logging.debug(debug_msg)
+            # if number_of_dropped:
+            #     logging.info(
+            #         "Number of Failure metrics that are older then %smin and won't be collected: %s",
+            #         METRIC_TIMESTAMP_THRESHOLD_MINUTES,
+            #         number_of_dropped,
+            #     )
+
             yield (creation_metric)
             yield (failure_metric)
 
--- a/charts/operators/Chart.yaml
+++ b/charts/operators/Chart.yaml
@@ -14,4 +14,4 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.0.10-rc.7
+version: 2.0.10-rc.8
--- a/charts/pelorus/Chart.lock
+++ b/charts/pelorus/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: exporters
   repository: file://./charts/exporters
-  version: 2.0.10-rc.7
-digest: sha256:2f4e5d396ce969b340d2768035737200c05ac074dc05aa109d737bfb336c07bb
-generated: "2023-05-31T17:19:59.256186361-03:00"
+  version: 2.0.10-rc.8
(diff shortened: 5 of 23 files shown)


docs/Development.md
```diff
@@ -242,7 +242,7 @@ If not defined specifically, exporters are using pre-built container images with
 
 Each Pelorus GitHub pull request that is [merged](https://github.com/dora-metrics/pelorus/pulls?q=is%3Apr+is%3Amerged) results in a new set of images that are tagged with the GitHub commit hash, for example `d6f6e6fa1c9d48ca1deeaf1c72585b94964cbf31` for the following [Pull Request](https://github.com/dora-metrics/pelorus/commit/d6f6e6fa1c9d48ca1deeaf1c72585b94964cbf31). The newest merged commit results in additional image tag `latest`.
```
mateusoliveira43, line 243: these changes should be done by script, but not now (just in the release PR)

charts/pelorus/templates/prometheus-rules.yaml
```diff
@@ -20,7 +20,7 @@ spec:
         avg(sdp:lead_time:by_app)
     - record: sdp:time_to_restore:by_issue
       expr: >
-        failure_resolution_timestamp{app!="unknown"} - failure_creation_timestamp{app!="unknown"}
+        failure_resolution_timestamp{app!="unknown"} - max_over_time(failure_creation_timestamp{app!="unknown"}[5y])
```
mateusoliveira43, line 23: Instead of 5 years I would like to use a dynamic value. We could use the value user sent from `prometheus_retention`, or even `$__range` from Grafana (but then we would not be able to store the rule here).

 But to be honest, I think I do not understand this metric. What should appear in selected ranges of Grafana: average of time to close issues that were created and resolved in that time range or it just need to be resolved inside that time range to appear?
 
I would like to have new tests in `_test/prometheus/test.yaml` to validate the changes made

exporters/failure/collector_base.py
```diff
@@ -8,6 +8,9 @@
 
 import pelorus
 
+# TBD_1: CI needs to create failures on the fly to enable this
```
mateusoliveira43, line 11: What is `TBD_1`?

mpryc: **T**o**B**e**D**one with number. Currently I can not enable this part, because we have pre-stored failures for various APIs (github, jira, ...) but I want to deliver this fix asap, so I commented out part which should be enabled. Currently it doesn't really matter as Prometheus will drop the old metrics anyway, however to be consistent with deploytime this part should be enabled in the future and we have also somewhere idea of creating failures during e2e tests.

charts/pelorus/templates/prometheus-rules.yaml
mpryc: This works slightly differently.

The time range from Grafana is applied for the entire `sdp:time_to_restore:by_issue` expression, without even specifying `$__range`. The `A-B` is for the time series within the `$__range`, so both **A** and **B** needs to be there. Now think of a situation that **B** was created before the `$__range` in this situation the `A-B` will not work even if the range is within **A** timestamp. To make sure we can properly calculate `A-B` where **A** is **failure_resolution_timestamp** and **B** is **failure_creation_timestamp**, we need to ensure the **B** is available at the time when **A** have happened. For this we use **max_over_time** which will make **B** available from the time it was created until **now** but not further then 5y ago. We do not have knowledge when the **B** occurred, so we can't have dynamic range here, but it's safe to say not older then 5y timeframe.

docs/GettingStarted/configuration/ExporterFailure.md
```diff
@@ -2,6 +2,8 @@
 
 Failure time exporter captures the timestamp at which a failure occurred, in a production environment, and when it was resolved. It does this by parsing the information of when the issue was created, and closed, in the Issue Tracker(s).
 
+Failure exporter only collects failure events that are less than 30 minutes old. Older failures won't be included unless they have been already collected.
```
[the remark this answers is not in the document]
mpryc: added.

exporters/failure/collector_base.py
mpryc: Changed.

docs/GettingStarted/configuration/ExporterFailure.md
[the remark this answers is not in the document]
mateusoliveira43: I think it still not clear

I would change the first phrase of this file to something like

"Failure time exporter captures the timestamp at which a broken deployment occurred, and when it was fixed. It does this by parsing the information of when the issue related to that broken deployment was created, and closed, in the Issue Tracker(s)."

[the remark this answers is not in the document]
mpryc: Changed as requested.

mateusoliveira43 (approved):

mpryc:
I am merging this change. It is approved and it has been previously tested by the CI and in the current revision only one doc line was changed.

PR #1013, opened 2023-06-02
fix: Restrict INFO log level verbosity

### LINKED ISSUE #1006: Too much logs using INFO level
### OpenShift version

Not related to OpenShift

### Problem description

When using INFO log level (the default), too much log is sent to user. It should be

### Steps to reproduce

Create a commitime, deploytime or failure exporter with default log level and check that number of logs grows fast as time passes.

### Current behavior

Too much log messages.

### Expected behavior

Only needed messages should be logged when using INFO log level.

### Code of Conduct

mateusoliveira43:

## Linked Issues

resolves #1006

## Description

Restricts the amount of logs when user sets it to INFO (the default).

## Testing Instructions

Check the logs of Committime, Deploytime, failure and webhook exporters with log level set to INFO.

### DIFF
--- a/exporters/tests/test_failure_exporter_app.py
+++ b/exporters/tests/test_failure_exporter_app.py
@@ -16,6 +16,7 @@
 import os
 
 import pytest
+from jira.exceptions import JIRAError
 
 from failure.app import set_up
 from pelorus.config.loading import MissingConfigDataError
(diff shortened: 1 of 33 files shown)


exporters/tests/test_failure_exporter_app.py
```diff
@@ -59,8 +62,9 @@ def test_app_pagerduty_with_required_options(caplog: pytest.LogCaptureFixture):
     )
 
     captured_logs = caplog.record_tuples
-    # 9 informational, 5 resolution, 58 creation
-    assert len(captured_logs) == 72
+    assert "Collected " not in caplog.text
+    # number of informational logs
+    assert len(captured_logs) == 8
     # number of error logs
     assert len([record for record in captured_logs if record[1] == 40]) == 0
```
mateusoliveira43, line 69: TODO use log variable instead of magic number

mateusoliveira43:
@mpryc did not touch webhook, but it has the same problem: each time the /metric endpoint is accessed, a new line is added to logs (but just one, not 20 or 100, like in the other exporters). But when a entry is added to it, also one log line appears. I though about just changing this add log to debug, what you think?

mpryc (approved):

mpryc (approved):

PR #900, opened 2023-03-18
fix: Azure devops URL parser

mateusoliveira43:
## Describe the behavior changes introduced in this PR

Fixes Azure devops URL parser.

## Linked Issues

resolves #898
resolves #902
related to #899 

## Testing Instructions

Use committime with data comming from `https://dev.azure.com/{organization}/{project}/_git/{repository}` (or custom enterprise URL, instead of  `dev.azure.com`) and using [`GIT_API`](https://pelorus.readthedocs.io/en/latest/GettingStarted/configuration/ExporterCommittime/#git_api) as `dev.azure.com` or `dev.azure.com/{organization}`.

- [x] provide exporter test image
- [x] test with `https://YOUR_USER@dev.azure.com/YOUR_ORG/YOUR_PROJECT/_git/YOUR_REPO`
- [x] test with `git@ssh.dev.azure.com:v3/YOUR_ORG/YOUR_PROJECT/YOUR_REPO`

### Example

To run e2e tests, run
```
./scripts/run-pelorus-e2e-tests -o mateusoliveira43 -b feat/azure-devops-test -e azure-devops-committime
```

To create test application, run
```
oc create namespace test-azure

oc create secret generic test-azure-secret --from-literal=password=YOUR_PERSONAL_ACCESS_TOKEN --type=kubernetes.io/basic-auth -n test-azure

oc secrets link builder test-azure-secret -n test-azure

oc new-app python:3.9~https://dev.azure.com/YOUR_ORG/YOUR_PROJECT/_git/YOUR_REPO \
--source-secret=test-azure-secret --strategy=docker \
-n test-azure -l app.kubernetes.io/name=test-azure
```
Create an Pelorus instance with
```yaml
kind: Pelorus
apiVersion: charts.pelorus.konveyor.io/v1alpha1
metadata:
  name: test-azure
  namespace: pelorus
spec:
  exporters:
    global: {}
    instances:
      - app_name: committime-exporter
        enabled: true
        exporter_type: committime
        image_name: quay.io/msouzaol/pelorus-exporter:latest
        extraEnv:
          - name: GIT_PROVIDER
            value: azure-devops
          - name: TOKEN
            value: YOUR_PERSONAL_ACCESS_TOKEN 
          - name: API_USER
            value: YOUR_USER
          - name: NAMESPACES
            value: test-azure
          - name: LOG_LEVEL
            value: DEBUG
      - app_name: deploy
        enabled: true
        exporter_type: deploytime
        extraEnv:
[...]

PR #1043, opened 2023-08-05
Introduce datastructures from committime cleanup

KevinMGranger:
I'm splitting up #720 into pieces so it's easier to review. This is an unobtrusive change, just copying over datastructures from that PR. (And unit tests for one of them).

Side note: `make unit-tests` gives me a bunch of `ModuleNotFoundError: No module named 'exporters'` errors. What gives?

### DIFF
--- a/exporters/committime/__init__.py
+++ b/exporters/committime/__init__.py
@@ -19,9 +19,9 @@ from __future__ import annotations
 
 import logging
 import re
-from typing import Optional
+from typing import NamedTuple, Optional
 
-import attr
+import attrs
 import giturlparse
 
 from pelorus.utils import collect_bad_attribute_path_error, get_nested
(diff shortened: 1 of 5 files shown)


exporters/committime/__init__.py
```diff
@@ -234,3 +234,81 @@ def commit_metric_from_build(app: str, build, errors: list) -> CommitMetric:
             setattr(metric, attr_name, value)
 
     return metric
+
+
+@attrs.frozen
+class GitRepo:
```
mateusoliveira43, line 240: This is a replace of [CommitMetric](https://github.com/dora-metrics/pelorus/blob/master/exporters/committime/__init__.py#L39) right?

If so, please checks the changes here https://github.com/dora-metrics/pelorus/pull/900 so they are also applied here

And if you want, you can look into this issue https://github.com/dora-metrics/pelorus/issues/903

KevinMGranger: It is not a replacement, but is part of the [replacement input](https://github.com/dora-metrics/pelorus/pull/720/files#diff-2793729d524eeb1428c47da0196219fc76db6a641df388fa295fadf1630bf582R91).

Mind if I ask a few more questions in #900?

PR #1047, opened 2023-08-10
Fix various dev tool install issues, make shellcheck a required lint

### LINKED ISSUE #518: install_dev_tools does not make it clear when an installation has failed
While error messages are printed in the console, the functions continue past the error, resulting in yet more errors.

In addition, the failure to install one program is not propagated to the exit status.

This could be solved in two ways:

1. Manually propagate the error condition until the end of the script
2. Design the script so that it's meant to be used on one program at a time, and exit early when there's an error. It could still be invoked for each program using the makefile. This could even be parallelized!

KevinMGranger:
Fixes #518

Fixes non-reported(?) issues with installing noobaa and shellcheck on ARM macs

It also looks like our shellcheck github action has been broken for a _while_. This PR:

- fixes the shellcheck makefile target
- simplifies that target
- makes shellcheck linting mandatory (I can separate this into another PR if you wish.)
- makes the github action only run on one version of python (it's irrelevant to python)

### Positive Test

1. Remove all non-python executables from your venv's `bin` dir (or just remove the whole venv)
2. `make cli_dev_tools`
3. Verify that all expected cli tools are present, including shellcheck and noobaa  
   (this only matters if you're on an ARM mac)

### Negative Test

1. Modify one of the tools' repo names / URLs in `get_tool_dl_url.py` so that the url check fails
2. Verify that the `install_dev_tools` script exits with a non-zero exit status

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10', '3.11']
+        python-version: ['3.11']
 
     steps:
       - name: Checkout
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -14,10 +14,6 @@ on:
 jobs:
   unit-test:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        python-version: ['3.11']
 
     steps:
       - name: Checkout
@@ -26,7 +22,7 @@ jobs:
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.11"
           cache: 'pip'
           cache-dependency-path: |
             **/requirements*.txt
--- a/scripts/get_tool_dl_url.py
+++ b/scripts/get_tool_dl_url.py
@@ -73,7 +73,7 @@ class Nooba:
     "Nooba's URL pattern."
 
     repo = "noobaa/noobaa-operator"
-    arch = "mac" if OS == "Darwin" else "linux"
+    arch = OS.lower()
     pattern = re.compile(f"https://(.*)-{arch}-(.*)[0-9]")
 
     @classmethod
--- a/scripts/get_tool_dl_url.py
+++ b/scripts/get_tool_dl_url.py
@@ -199,6 +199,7 @@ parser.add_argument(
     help=f"The executable to retrieve the URL for.\nSupported: {', '.join(CLI_NAMES)}",
     choices=CLI_NAMES,
 )
+parser.add_argument("-v", "--verbose", action="store_true", default=False)
 
 if __name__ == "__main__":
     args = parser.parse_args()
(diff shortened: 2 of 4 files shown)


.github/workflows/shellcheck.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
```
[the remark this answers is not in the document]
KevinMGranger: Good point!

KevinMGranger:
@mateusoliveira43 you okay with skipping the flaky openshift tests?

scripts/install_dev_tools.sh
mateusoliveira43, line 1: Here I had to make a change to `noobaa` install command to work for me (Linux fedora36redhat x86_64)
```diff
diff --git a/scripts/install_dev_tools.sh b/scripts/install_dev_tools.sh
index 87dbc97..1c568c3 100755
--- a/scripts/install_dev_tools.sh
+++ b/scripts/install_dev_tools.sh
@@ -267,8 +267,8 @@ if should_cli_be_installed "noobaa" "${cli_tools_arr[@]}" && \
       echo "$NOOBAA_CLIENT_URL"
       download_file_from_url "${NOOBAA_CLIENT_URL}" "${DWN_DIR}"
       NOOBAA_CLIENT_PATH="${DWN_DIR}"/$(basename "${NOOBAA_CLIENT_URL}")
-      mv "${NOOBAA_CLIENT_PATH}" "${VENV}/bin/noobaa"
-      chmod +x "${VENV}/bin/noobaa"
+      extract_file_to_dir "${NOOBAA_CLIENT_PATH}" "${VENV}/bin/"
+      mv "${VENV}/bin/noobaa-operator" "${VENV}/bin/noobaa"
 fi
 
 if should_cli_be_installed "operator-sdk" "${cli_tools_arr[@]}" && \
```

Does this also work on Mac?

scripts/get_tool_dl_url.py
```diff
@@ -73,7 +73,7 @@ class Nooba:
     "Nooba's URL pattern."
 
     repo = "noobaa/noobaa-operator"
-    arch = "mac" if OS == "Darwin" else "linux"
+    arch = OS.lower()
     pattern = re.compile(f"https://(.*)-{arch}-(.*)[0-9]")
```
mateusoliveira43, line 77: I think this can break for arm (noobaa url changed for versions higher than 5.11 https://github.com/noobaa/noobaa-operator/releases)

But if the change I suggested in `scripts/install_dev_tools.sh` works for you, maybe we can ignore this change, because I want to change this script in the future to be like a dependabot (looking for updates, like https://github.com/dora-metrics/pelorus/blob/master/.github/workflows/ci.yml looks for OpenShift updates)

scripts/install_dev_tools.sh
KevinMGranger: It works! Applied.

scripts/get_tool_dl_url.py
KevinMGranger: Can you elaborate a bit on how that script works?

mateusoliveira43: The github action looking for OpenShift version updates? It looks an url that returns a json and checks if we are using last 4 minor stable releases (for version 4) in the project (looking in some doc files for example) and also in PROW repo, to check that we are running CI also in last 4 minor stable releases

mateusoliveira43 (approved):

mateusoliveira43:
You can merge without waiting PROW succeed, @KevinMGranger

PR #1067, opened 2023-09-20
Script to sent fake data to the webhook exporter.

mpryc:
This is script that discovers the webhook exporter route and sends the webhook data in a given interval to the webhook endpoint.

It can be used for testing or demo purposes.

Usage of this script is explained within the script itself.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -30,3 +30,10 @@ Used to create Pelorus helm based operator from the helm charts.
 
 Ensures modification to the operator files can be cleanly applied and allows
 to recreate operator easily.
+
+## scripts/send_example_payload.py
+
+Used for testing and demo purposes to send fake data to the webhook exporter.
+Usually the script is copied over, modified (APP_NAME, send_events()) and
+used to fill out the webhook receiver to represent deployments, commits
+and failures in the Grafana.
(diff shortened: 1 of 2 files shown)


scripts/README.md
```diff
@@ -30,3 +30,10 @@ Used to create Pelorus helm based operator from the helm charts.
 
 Ensures modification to the operator files can be cleanly applied and allows
 to recreate operator easily.
+
+## scripts/send_example_payload.py
+
+Used for testing and demo purposes to send fake data to the webhook exporter.
+Usually the script is copied over, modified (APP_NAME, send_events()) and
```
mateusoliveira43, line 37: The change in send events would be in the function itself or just the intervals?

mpryc: Really not limited to anything, could be interval, etc... but don't want to make yet another application to do that so it's up to the person who could use that file as a base.

mateusoliveira43: I would let a comment about that it can customized to user's needs, but I would add CLI to change intervals and `APP_NAME`

Implementation example https://github.com/dora-metrics/pelorus/blob/master/scripts/update_projects_version.py#L370-L404

PR #1084, opened 2023-10-13
Update OCP check script and upstream support with the prow changes

mpryc:
The PROW CI jobs were updated to use only latest OCP release for their runs. The documentation and the check openshift version script needs to be changed as well.

PROW PR: https://github.com/openshift/release/pull/44348

### DIFF
--- a/docs/UpstreamSupport.md
+++ b/docs/UpstreamSupport.md
@@ -1,8 +1,8 @@
 # Pelorus Upstream Support Statement
 
-The Pelorus engineering team will provide **best-effort** level support for it on the currently latest and the latest 3 previous released stable minor versions of OpenShift version 4.
+The Pelorus engineering team will provide **best-effort** level support for it on the currently latest released stable minor version of the OpenShift 4.
 
-> **NOTE:** Pelorus is currently automated tested against versions **4.10**, **4.11**, **4.12** and **4.13** of OpenShift.
+> **NOTE:** Pelorus is currently automated tested against version **4.13** of OpenShift.
 
 * To file a bug, please create a [Bug issue](https://github.com/dora-metrics/pelorus/issues/new?assignees=&labels=kind%2Fbug%2Cneeds-triage&template=bug.yml)
 
--- a/scripts/check_openshift_version.py
+++ b/scripts/check_openshift_version.py
@@ -26,11 +26,12 @@ OPENSHIFT_REPO = "openshift/release"
 OPENSHIFT_BRANCH = "master"
 RAW_URL = f"https://raw.githubusercontent.com/{OPENSHIFT_REPO}/{OPENSHIFT_BRANCH}"
 
+OCP_VERSIONS_URL = "https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/graph"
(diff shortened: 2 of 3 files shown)


scripts/check_openshift_version.py
```diff
@@ -26,11 +26,12 @@
 OPENSHIFT_BRANCH = "master"
 RAW_URL = f"https://raw.githubusercontent.com/{OPENSHIFT_REPO}/{OPENSHIFT_BRANCH}"
 
+OCP_CI_VERSION_URL = "https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/graph"
```
mateusoliveira43, line 29: would change to `OCP_VERSIONS_URL`

docs/UpstreamSupport.md
```diff
@@ -1,8 +1,8 @@
 # Pelorus Upstream Support Statement
 
-The Pelorus engineering team will provide **best-effort** level support for it on the currently latest and the latest 3 previous released stable minor versions of OpenShift version 4.
+The Pelorus engineering team will provide **best-effort** level support for it on the currently latest released stable minor versions of the OpenShift version 4.
```
mateusoliveira43, line 3: ...stable minor **versions** of the... -> ...stable minor **version** of the...

mateusoliveira43 (changes requested):
Also need to remove old references in README.md and .github/ISSUE_TEMPLATE/bug.yml

I would also add a deprecated warning (just a comment is enough) to Makefile E2E test scenario 2 command https://github.com/dora-metrics/pelorus/blob/master/Makefile#L126

mateusoliveira43:
/hold
wait the other PR merge

mpryc:
> Also need to remove old references in README.md and .github/ISSUE_TEMPLATE/bug.yml
> 
> I would also add a deprecated warning (just a comment is enough) to Makefile E2E test scenario 2 command https://github.com/dora-metrics/pelorus/blob/master/Makefile#L126

I left the issue list of older OCP releases. It's a bug entry and we don't force user to use X or Y version of OCP + we should allow to log against already existing deployments.

I don't think we should deprecate scenario-2, we just don't run them in the periodics.

mateusoliveira43 (approved):

mateusoliveira43:
I think the regex for PROW is working, it did not trigger for your newest commit that does not touch the files delacred in it :smile: 

But I think we left this there https://github.com/openshift/release/blob/master/core-services/prow/02_config/dora-metrics/pelorus/_prowconfig.yaml#L19 we should remove it right? 

(if you want, you can bypass branch protection and merge this PR)

mpryc:
> I think the regex for PROW is working, it did not trigger for your newest commit that does not touch the files delacred in it 😄
> 
> But I think we left this there https://github.com/openshift/release/blob/master/core-services/prow/02_config/dora-metrics/pelorus/_prowconfig.yaml#L19 we should remove it right?
> 
> (if you want, you can bypass branch protection and merge this PR)

I think it's because my PR was after the above was merged, so it got old required jobs.

mpryc:
Merging as it previously passed checks.

PR #1122, opened 2024-01-30
Fixed issue with projects field in the failuretime exporter using Azure DevOps

jkmattatall:
Problem: If an Azure DevOps server has a project name with a space in it ("Test Project" for example), the _projects_ variable in _exporters/failure/collector_azure_devops.py_ separates the project into two, causing it not to pick up the project. In the documentation the environment variable for projects asks for a "comma separated list of strings" [here](https://pelorus.readthedocs.io/en/v2.0.11/GettingStarted/configuration/ExporterFailure/#projects), yet a space causes it to be separated as well. 

Example: The project "Test Project" would be separated into ["Test", "Project"], meaning "Test Project" would never be picked up by the failuretime exporter. Whereas we would want it to be separated to ["Test Project"].

Solution: Use the provided comma_separated function to properly separate the projects provided to the failuretime exporter.

### DIFF
--- a/exporters/failure/collector_azure_devops.py
+++ b/exporters/failure/collector_azure_devops.py
@@ -27,7 +27,7 @@ from msrest.authentication import BasicAuthentication
 
 from failure.collector_base import AbstractFailureCollector, TrackerIssue
 from pelorus.config import env_var_names, env_vars
-from pelorus.config.converters import comma_separated, comma_or_whitespace_separated, pass_through
+from pelorus.config.converters import comma_or_whitespace_separated, pass_through
 from pelorus.config.log import REDACT, log
 from pelorus.errors import FailureProviderAuthenticationError
 from pelorus.timeutil import parse_assuming_utc_with_fallback, second_precision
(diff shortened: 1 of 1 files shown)


jkmattatall:
@mateusoliveira43 Hey there! Looking for some assistance with this PR and I noticed you were the one who worked on most/all of the code I'm looking at. 

From my view, the change I made only makes it so that whitespaces are allowed within project names for the Azure DevOps failuretime exporter specifically. I've checked the Jira failuretime exporter files (where the OpenShift job seems to be failing), and it seems like there's no reference to the collector_azure_devops.py file, which means no reference to the change I made. 

It seems like the comma_or_whitespace_separated() and comma_separated() methods both have the same input/output types, the change shouldn't make any difference in building besides in testing. However, in the Azure DevOps failuretime exporter test files, it seems like all projects variables aren't whitespace separated anyways, therefore the output should be the same.

Either I am missing something or the build would be borked anyways without my change. Thanks!

PR #1132, opened 2024-05-09
Rebuilding docs requirements tree to get the builds working again.

etsauer:

## Description

This PR pulls in updated versions of all the packages involved in building our docs with `mkdocs`. This should fix the broken builds on readthedocs.org

## Testing Instructions

1. Check the readthedocs CI check and make sure that it passes
2. Check the built docs site and make sure everything looks right


pyproject.toml
```diff
@@ -42,12 +42,13 @@ yamllint = "^1.31.0"
 yq = "^3.2.2"
 
 [tool.poetry.group.doc.dependencies]
```
mateusoliveira43, line 44: I would change less things here

From what I searched, here is the problem https://github.com/readthedocs/readthedocs.org/issues/11247

And here the solution https://github.com/squidfunk/mkdocs-material/releases/tag/9.5.16

So, I would try to only update `mkdocs-material`

To do so, run `poetry update mkdocs-material --with=doc`

This command will update 3 packages listed in `pyproject.toml`, manually update then here
```yaml
mkdocs = "^1.6.0"
...
mkdocs-material = "^9.5.21"
...
pymdown-extensions = "^10.4"
```

Then, run `make update-requirements`

mateusoliveira43 (changes requested):
need to run `make update-requirements` and update `poetry.lock`

etsauer:
Ok, how's this look?

PR #1141, opened 2024-07-09
Add the ability to configure exporters system ca bundle

beelzetron:

## Linked Issues
N/A

## Description

Some people need to trust internal CAs emitting TLS certificates for internal only services.

Adding the internal CAs to the container system CA bundle avoids certificate validation errors in the exporters.

This PR adds the ability to override the exporters system CA bundle with a ConfigMap.
The ConfigMap must contain the `ca-bundle.crt` key populated with PEM formatted certificates to be trusted.

The ConfigMap will be mounted into the exporter pods updating the system CA bundle located in `/etc/pki/tls/certs/ca-bundle.crt`.

## Testing Instructions

1. Create the ConfigMap with the CA bundle:

    ```yaml
    apiVersion: v1
    kind: ConfigMap
    metadata:
      labels:
        config.openshift.io/inject-trusted-cabundle: "true"
      name: cluster-ca-bundle
    ```
2. Create a values file with the `custom_ca_configmap` configuration:
    ```yaml
    instances:
      - app_name: deploytime-exporter
        exporter_type: deploytime
        custom_ca_configmap: cluster-ca-bundle
      - app_name: committime-exporter
        exporter_type: committime
        custom_ca_configmap: cluster-ca-bundle
        extraEnv:
        - name: PROVIDER
          value: containerimage
    ```
3. Deploy the exporters helm chart using the values file just created.

### DIFF
--- a/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
+++ b/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
@@ -36,7 +36,9 @@ spec:
         - name: configmap-cluster-ca-bundle
           configMap:
             name: {{ .custom_ca_configmap }}
-            defaultMode: 420
+            items:
+            - key: ca-bundle.crt
+              path: tls-ca-bundle.pem
         {{- end }}
       containers:
         - name: {{ .app_name }}
@@ -50,7 +52,7 @@ spec:
             {{- if .custom_ca_configmap }}
             - name: configmap-cluster-ca-bundle
               readOnly: true
-              mountPath: /etc/pki/tls/certs
+              mountPath: /etc/pki/ca-trust/extracted/pem
             {{- end }}
 
           envFrom:
--- a/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
+++ b/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
@@ -32,6 +32,12 @@ spec:
           configMap:
             name: {{ $config.map_name }}
         {{- end }}
+        {{- if .custom_ca }}
+        - name: configmap-cluster-ca-bundle
+          configMap:
+            name: {{ .custom_ca.map_name }}
+            defaultMode: 420
+        {{- end }}
       containers:
         - name: {{ .app_name }}
           imagePullPolicy: Always
@@ -40,6 +46,11 @@ spec:
             {{- range $config := .custom_certs }}
             - name: custom-cert-volume-{{$config.map_name}}
               mountPath: /etc/pelorus/custom_certs/{{$config.map_name}}
+            {{- end }}
+            {{- if .custom_ca }}
+            - name: configmap-cluster-ca-bundle
+              readOnly: true
+              mountPath: /etc/pki/tls/certs
             {{- end }}
 
           envFrom:
--- a/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
+++ b/charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
@@ -32,10 +32,10 @@ spec:
           configMap:
             name: {{ $config.map_name }}
         {{- end }}
-        {{- if .custom_ca }}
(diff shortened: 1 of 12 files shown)


charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
```diff
@@ -32,6 +32,12 @@ spec:
           configMap:
             name: {{ $config.map_name }}
         {{- end }}
+        {{- if .custom_ca_configmap }}
```
mpryc, line 35: I think this will not work without updating CRD. It will require re-generating pelorus-operator and creating new release:

https://pelorus.readthedocs.io/en/latest/Development/#versioning-process

charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
```diff
@@ -32,6 +32,12 @@ spec:
...
+        - name: configmap-cluster-ca-bundle
+          configMap:
+            name: {{ .custom_ca_configmap }}
+            defaultMode: 420
```
mpryc, line 39: Based on the doc, isn't this part missing `items` ?

https://docs.openshift.com/container-platform/4.16/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki

charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
mpryc: Would be also great to add hashed example of how this works in the values:

https://github.com/dora-metrics/pelorus/blob/master/charts/pelorus/values.yaml

charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
beelzetron: This is just the pod `volumes:` definition pointing to a ConfigMap, no need to iterate here since we need just one. The ConfigMap can be created following the steps in the docs you linked.

mpryc: I refer to that part of the doc:

```
      - name: trusted-ca
        configMap:
          name: trusted-ca
          items:
            - key: ca-bundle.crt 
              path: tls-ca-bundle.pem 



ca-bundle.crt is required as the ConfigMap key.
tls-ca-bundle.pem is required as the ConfigMap path.

```

charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
beelzetron: Added a hashed example to the default values file. I also managed to bump the release to 2.0.13-rc.1.

charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
beelzetron: Ah now I got it. I refactored that part to be consistent with the documentation. It was working anyway because I chose another mount point where it was not required to rename the key.

charts/pelorus/charts/exporters/templates/_deploymentconfig.yaml
beelzetron: > I think this will not work without updating CRD. It will require re-generating pelorus-operator and creating new release:
> 
> https://pelorus.readthedocs.io/en/latest/Development/#versioning-process

I managed to bump the version.

beelzetron:
> Test images available! 🧪🚀 To test operator with them, run
> 
> ```
> operator-sdk run bundle \
> quay.io/pelorus/rc-pelorus-operator-bundle:vpr1141-23e90f7 \
> --namespace test-pelorus-operator
> ```
> 
> To clean up environment afterwards, run
> 
> ```
> operator-sdk cleanup pelorus-operator --namespace test-pelorus-operator
> ```

Tested successfully following this instructions.

